PR #63, opened 2016-06-13
fix reaction image in rtf report

JasonYCHuang:
after generating reaction reports, you should find correct and well-fit image in rtf reports.


Sergeykot:
:+1:

PR #68, opened 2016-06-22
generate multi-reactions report

JasonYCHuang:
I disable checkedAll checkBox at Element List under report-view.
This is due to current data structure when check checkedAll.

__Future work:__
1. improve report format
2. add sample analysis
3. re-open reactions list checkedAll box
4. a list let user to determine reactions order
5. user customised default setting
6. page break between each reactions in the RTF report.
7. comment function

### DIFF
--- a/app/api/chemotion/report_api.rb
+++ b/app/api/chemotion/report_api.rb
@@ -7,89 +7,11 @@ module Chemotion
         requires :id
       end
       get :rtf do
-        reaction = Reaction.find(params[:id])
-
-        svg_paths = {}
-        svg_paths[:starting_materials] = reaction.starting_materials.map do |material|
-          material.get_svg_path
-        end
-        svg_paths[:reactants] = reaction.reactants.map do |material|
-          material.get_svg_path
-        end
-        svg_paths[:products] = reaction.products.map do |material|
-          material.get_svg_path
-        end
-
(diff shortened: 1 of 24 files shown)


app/api/chemotion/report_api.rb
```diff
@@ -198,5 +198,146 @@ class ReportAPI < Grape::API
...
+      params do
+        requires :ids
+        requires :settings
+      end
+
+      get :rtf do
+        ids = params[:ids].split("_")
+        settings = params[:settings].split("_")
+
+        report = Report::RTFReport.new do |r|
```
Sergeykot: Looks not good. Please make a module for that. Example

https://github.com/ComPlat/chemotion_ELN/blob/master/lib/svg/reaction_composer.rb

You can do this without a class, just a module with function would also fine. However if you want you can build a class like done in the example.

In general, Rails and API actions have to laconic and don't contains much logic.

PiTrem: same could be said for `get :rtf` from line 9.
This is just following the practice.
If this is to be cleaned so should be the upper part.
I don't mind it for now as long as,  the whole report generation part is moved to, say, somewhere  in lib/report/ with a new issue (next iteration?)

Sergeykot:
Perfect :+1:

PR #69, opened 2016-06-23
reaction-product is a sample copy

JasonYCHuang:
Under reaction edit mode, we use a sample copy, not split, for __product__.


Sergeykot:
:+1:

PR #71, opened 2016-06-24
sample editor warning for descendents resolve #301

JasonYCHuang:
Sample editor will show warning message if the sample has parents or children.


Sergeykot:
:+1:

PR #86, opened 2016-07-19
sample inchikey/smile copy to clipboard resolve #303

JasonYCHuang:
You need to ```npm install``` before using it.


Sergeykot:
:+1:

PR #87, opened 2016-07-19
fix warning setState & empty-list when back from collection-managements

JasonYCHuang:
solve:
1. ```setState``` warning
2. When coming back from collection-managements, __List__ will no longer be empty.


Sergeykot:
:+1:

PR #94, opened 2016-07-22
Fix the bug while searching the polymer

CamAnNguyen:
- Fix the warning when 'input' controller has null value
- Change the fingerprint generation to use original molfile from samples table
- Since OpenBabel doen't use the last 3 bits, we will enable those bits for polymer search. We have 3 bits, so we can have total 8 "special groups"
- Resolved for #353


Sergeykot:
:+1:

PR #95, opened 2016-07-22
Update polymer search

CamAnNguyen:
- Calculate Tanimoto as float, more accurate sort
- Re-calcuate the num_set_bits after changing polymer molecule


Sergeykot:
:+1:

PR #111, opened 2016-08-05
fix broken validation  for name_abbreviation

### LINKED ISSUE #400: Upload option in generic element/segment
It would be great to have the option to upload a file in the generic segment/element to be stored with the element. The upload function should be given with another input field that can be used for a label (e.g. CIF for MOF or similar). We could use that later on to retrieve information from this data.

PiTrem:
add spec
resolve #400

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -16,10 +16,26 @@ class User < ActiveRecord::Base
 
 
   validates_presence_of :first_name, :last_name, allow_blank: false
+  validates :name_abbreviation, uniqueness:  {message: " has already been taken." },
+    format: {with: /\A[a-zA-Z][a-zA-Z0-9\-_]{0,3}[a-zA-Z]\Z/,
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -16,11 +16,27 @@ class User < ActiveRecord::Base
 
 
   validates_presence_of :first_name, :last_name, allow_blank: false
+  validates :name_abbreviation, uniqueness:  {message: " has already been taken." },
+    format: {with: /\A[a-zA-Z][a-zA-Z0-9\-_]{0,3}[a-zA-Z]\Z/,
+    message: "can be alphanumeric, middle '_' and '-' are allowed,"+
+    " but leading or trailing digit, '-' and '_' are not."}
```
JasonYCHuang: maybe?

~~can be~~ Alphanumeric, middle '_' and '-' are allowed ...

JasonYCHuang:
👍

PR #115, opened 2016-08-09
Fix assign sample, upd shared coll permissions

### LINKED ISSUE #402: select box generic segment/element
it would be great if one could see additional information in the select box which is then not used for the data entry. e.g. 
molecule name - string
only string is used for the data entry
This could help to identify the right component in the selectbox

PiTrem:
resolve #402 #405 #407


JasonYCHuang:
👍

PR #117, opened 2016-08-12
add 8d grace time to really_destroy_task

### LINKED ISSUE #430: There is no consistency in the naming system of new elements
The generated elements in the ELN are named differently based on the element type in an inconsistent way. If the name abbreviation for the user is "user", then:
The new samples are called: user-1, user-2 (no prefix, user name abbreviation is included)
The new reactions are called: user-R1, user-R2 (prefix and user name abbreviation is included)
The new wellplates are called: New Wellplate, New Wellplate (no numbers or prefix, different elements have the same name by default, user name abbreviation is not included)
The new screens are called: New Screen, New Screen  (no numbers or prefix, different elements have the same name by default, user name abbreviation is not included)
The new Research Plan is called:  New Research Plan, New Research Plan (no numbers or prefix, different elements have the same name by default, user name abbreviation is not included)

The naming of reactions seems to be the best.

PiTrem:
resolve #430

### DIFF
--- a/lib/really_destroy_task.rb
+++ b/lib/really_destroy_task.rb
@@ -1,7 +1,7 @@
 module ReallyDestroyTask
   def self.execute!
     models.each do |model|
-      model.constantize.only_deleted.map(&:really_destroy!)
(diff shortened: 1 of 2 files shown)


JasonYCHuang:
👍

PR #121, opened 2016-08-17
React unit-test with mocha, enzyme, and sinon

JasonYCHuang:
I also improve Report FLux during the task, but it is not directly related to React unit-test.
Therefore, I separate into 2 commits.


JasonYCHuang:
# Not complete. Don't merge!!!

Sergeykot:
:+1:

PR #125, opened 2016-08-22
Fingerprint RSpec

CamAnNguyen:
Update fingerprint RSpec

### DIFF
--- a/spec/models/fingerprint_spec.rb
+++ b/spec/models/fingerprint_spec.rb
@@ -3,9 +3,34 @@ require 'rails_helper'
 RSpec.describe Fingerprint, type: :model do
   describe 'creation' do
     let(:fingerprint) { build :fingerprint }
+    let(:new_fingerprint) { build :fingerprint }
 
     it 'possible to create valid fingerprint' do
       expect(fingerprint.valid?).to be(true)
+    end
+
+    it 'validates the presence of fp fields' do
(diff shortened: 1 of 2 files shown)


spec/models/fingerprint_spec.rb
```diff
@@ -62,4 +62,56 @@
...
+
+      expect(standard.include? 'R#').to be(false)
+      expect(standard.include? 'PolymersList').to be(false)
+    end
+
+    it 'should create a fingerprint' do
+      fp_id = Fingerprint.find_or_create_by_molfile(polymer_molfile)
+      fp = Fingerprint.find(fp_id)
+
+      expect(fp.fp0).to eq("0000000000000000000000000000000000000000000000000000000000000000")
```
Sergeykot: when the string contains only from "0" I would user "0" * n
n is an Integer

Sergeykot:
plz squash in one commit

Sergeykot:
git rebase -i HEAD~2

spec/models/fingerprint_spec.rb
CamAnNguyen: It's more simple for me when I manually generated a 64-bits fingerprint. It's also easy to clearly see the fingerprint I want to test, that the generated fingerprint should be exactly like this. Can I just keep it like this?

Sergeykot: yes sure, now I agree with you

Sergeykot:
:+1:

PR #129, opened 2016-08-23
Automatically set Dataset name based on attachment

### LINKED ISSUE #454: Create a new hidden list such as for reactants and solvents for dilutions of one sample
The sample list easily becomes overcrowded. For chemists who generate a set of dilutions from one sample, it could be useful to have a table of concentrations in one sample (could be the original sample with the highest concentration, or an abstract one with no concentration). This table can look similar to the table in the reaction scheme, where it has links to hidden samples. Those hidden samples can be generated simply by just adding the concentration, and they adopt all the other properties from the original sample.

image

CamAnNguyen:
- Analysis dataset name will be automatically set based on the last attachment name
- Change CSS bug for Dataset Modal to fit with StructurEditorModal
- Resolve #454, #451

### DIFF
--- a/app/assets/javascripts/components/Dataset.js
+++ b/app/assets/javascripts/components/Dataset.js
@@ -60,8 +60,12 @@ export default class Dataset extends Component {
   handleFileDrop(files) {
     const {dataset} = this.state;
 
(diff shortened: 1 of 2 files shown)


Sergeykot:
:+1:

PR #131, opened 2016-08-25
Moving element using keyboard

CamAnNguyen:
- From this commit we can move element (samples, reactions, wellplates, screens) using the arrow key. Press "Enter" or "right arrow" to open element.
- "Keydown" will be manipulate through added KeyboardStore


Sergeykot:
:+1:

PiTrem:
seems to not work with ff
breaks with chrome

CamAnNguyen:
Sorry for my carelessness. Code updated

PR #133, opened 2016-08-29
Add button to hide CollectionTree

CamAnNguyen:
- Add a button a hide colelctionTree, so that we have more space for ElementsDetails table
- Fix a warning to unknown "inverse" props in CollectionManagement


Sergeykot:
:+1:

PR #134, opened 2016-08-29
Update ReactionDetails UI

CamAnNguyen:
- Make old "GenerateReport" button smaller and move to title, next to close button
- Reaction title now will take 1 line space above reaction svg
- Stretch reaction svg to 100% width
- Svg Generating: Add a small space between starting materials
- Use JS to auto update StructureEditorModal. 
- Fix CSS bugs when Sample or Element content is too big, lead to the breaking of ElementsTable
- Fix FF bug in Ketcher and ElementsDetail
- Redmine: #448 #450 #462 #465

### DIFF
--- a/app/assets/javascripts/components/ReactionDetails.js
+++ b/app/assets/javascripts/components/ReactionDetails.js
@@ -176,15 +176,13 @@ export default class ReactionDetails extends Component {
       )
   }
 
-  reactionSVG(reaction, svgContainerStyle) {
+  reactionSVG(reaction) {
     if(!reaction.svgPath) {
       return false;
     } else {
       return (
-        <Col md={9}>
-          <div style={svgContainerStyle}>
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/components/ReactionDetails.js
```diff
@@ -181,10 +181,8 @@ export default class ReactionDetails extends Component {
       return false;
     } else {
       return (
-        <Col md={9}>
-          <div style={svgContainerStyle}>
-            <SVG key={reaction.svgPath} src={reaction.svgPath} className='reaction-details'/>
-          </div>
+        <Col md={12}>
```
Sergeykot: variable svgContainerStyle is still present at function 'reactionSVG' definition. But we don't use it anymore.
So plz remove it from function definition if it is not needed

Sergeykot:
:+1:

PR #139, opened 2016-08-31
Re-enable keyboard navigation

CamAnNguyen:
From now, we only process arrow key from BODY. All `onKeyDown` event from other elements will be ignored


Sergeykot:
:+1:

PR #143, opened 2016-09-02
fix Dom element error when not on pages/welcome

### LINKED ISSUE #472: Adding pH to the scheme tab of reactions
Some users requested the addition of pH-value as an important parameter to the scheme tab for reactions

PiTrem:
resolve #472


Sergeykot:
can be merged after tests pass :+1:

PR #147, opened 2016-09-06
resolve dual export in Material.js

### LINKED ISSUE #474: Postgres Database names inconsistent during installation process
The database names in the database.yml.example and in the installation scripts are not the same: chemotion_dev vs. chemotion_eln_dev.
https://github.com/ComPlat/chemotion_ELN/blob/development-5/config/database.yml.example
vs.
https://github.com/ComPlat/chemotion_ELN/blob/development-5/scripts/install_development.sh
https://github.com/ComPlat/chemotion_ELN/blob/development-5/scripts/install_production.sh

Causes error in the installation process.

PiTrem:
resolve #474

### DIFF
--- a/app/assets/javascripts/components/Material.js
+++ b/app/assets/javascripts/components/Material.js
@@ -16,7 +16,7 @@ const collect = (connect, monitor) => ({
   isDragging: monitor.isDragging()
 });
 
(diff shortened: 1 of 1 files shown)


Sergeykot:
:+1:

PR #152, opened 2016-09-07
display multiple solvents resolve #502

### LINKED ISSUE #502: Checkbox is dancing around on hover
See attached Gif.

Peek 2021-08-05 14-45

As we defined a border only upon hover, the checkbox dances around.

May I suggest to either:

1. Get rid of the hover effect. It's not 2013 anymore, no one does that.
2. Define an equally wide border with transparent color for the non hovered state to prevent dancing?
3. Choose the method that is designed for that: we could use `outline` instead of `border` as that one doesn't change the layout model.

The classes this should be applied to are:
`.all-checkbox` and `.span-checkbox` as defined in `app/assets/stylesheets/common.scss`.

JasonYCHuang:
display multiple solvents @ reactions detail

### DIFF
--- a/app/assets/javascripts/components/Material.js
+++ b/app/assets/javascripts/components/Material.js
@@ -388,6 +388,7 @@ class Material extends Component {
         <td>
           <Button
             bsStyle="danger"
+            bsSize="small"
             onClick={() => deleteMaterial(material)} >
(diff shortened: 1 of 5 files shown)


Sergeykot:
:+1:

PR #153, opened 2016-09-07
Update export features

### LINKED ISSUE #496: skip tests
Add ability to skip single or all tests at once.
A structure for that is in testacceptance workflow yml file, needs to be tested and adapted to the other workflow files.

CamAnNguyen:
- Export all the Sample table
- Change reaction and wellplates export.
- Create blank xls report file if no sample or element created
- Exclude some fields from samples
- Include sum_formula and cano_smiles from molecule
- "generate_file" function include using: {Associated_model}.{field}
- Resolve #496

### DIFF
--- a/lib/report/excel_export.rb
+++ b/lib/report/excel_export.rb
@@ -7,36 +7,70 @@ class Report::ExcelExport
     @@sample_list << sample
   end
 
-  def generate_file
-    p = Axlsx::Package.new
+  def generate_file excluded_field, included_field
+    return -1 if @@sample_list.empty? || @@sample_list.first == nil
 
+    p = Axlsx::Package.new
     img = File.expand_path('../image1.jpeg', __FILE__)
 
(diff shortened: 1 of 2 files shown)


lib/report/excel_export.rb
```diff
@@ -42,6 +42,60 @@ def generate_file
...
+
+    img = File.expand_path('../image1.jpeg', __FILE__)
+
+    p.workbook.styles.fonts.first.name = 'Calibri'
+    p.workbook.add_worksheet(:name => "ChemOffice") do |sheet|
+      if @@sample_list.empty? || @@sample_list.first == nil
+        break
+      end
+
+      header = [""] + @@sample_list.first.attributes.keys
```
PiTrem: uses `Sample.attribute_name` instead.
Check requirements. not all attributes should be exported

CamAnNguyen: Updated

PR #157, opened 2016-09-09
Update import feature

### LINKED ISSUE #497: Continiuos Deployment :rocket:
very time consuming, no blue print available in GitLab CI, maybe via push to Docker from GH actions & pull on the Server side?

CamAnNguyen:
- Import file must have mofile or smiles or cano_smiles
- Importing is based on molfile or smiles
- Additional field other than Sample column will be ignored
- Fields that do not exist in the import file will take default value
- Exclude: short_label, external_label, ancestry
- Resolve #497


.nvmrc.example
```diff
@@ -1 +1 @@
-0.10.40
\ No newline at end of file
+4.4.7
```
Sergeykot: its 4.4.3 on  server. we can update it late though

Sergeykot:
build failed

PR #161, opened 2016-09-12
Add Element icon to ElementDetail header

### LINKED ISSUE #463: Add "undefined structure" as default value for decoupled samples
I am not sure if this was mentioned in another issue but we had this point in one of our meetings: could you add the term "undefined structure" as a default value to the field "sum formula" in the properties tab for samples (save sample would give undefined structure if no change added by the user)? That would help to avoid misunderstandings and will facilitate the process.

CamAnNguyen:
- Add icon to header of SampleDetail, WellplateDetail, and ScreenDetail
- Resolve #463

### DIFF
--- a/app/assets/javascripts/components/SampleDetails.js
+++ b/app/assets/javascripts/components/SampleDetails.js
@@ -210,7 +210,9 @@ export default class SampleDetails extends React.Component {
 
   sampleHeader(sample) {
     return (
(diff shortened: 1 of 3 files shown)


Sergeykot:
:+1:

PiTrem:
#463 has been closed already; this is a part only of #492

PR #163, opened 2016-09-12
Display Polymer in Reaction Product

### LINKED ISSUE #441: Temp. Chart - Changing time step difficult - 25-200 wrong order of figures in scheme
- [x] I am a system/ELN admin

# Expected Behavior

- Easily changing numbers in the table, NOT only if i mark one single figure and change it, but also if I put the cursor somewhere and then delete from there or just start typing
- If my first step in the table is 25 °C and the second 200°C this should be depicted as 25-200°C in the scheme, NOT 200-25°C (For 25-400°C it is working fine)

# Current Behavior

It is very difficult to change the time (see GIF attached)
Bug - Temp Chart - Changing time step to 2hrs difficult - 25-200 wrong order of figures in scheme

Display of 200-25°C in the scheme

# Failure Information 

It should be 25-200°C and times should be changeable if I put the cursor somewhere and then delete from there or just start typing.

## Steps to Reproduce

see GIF attached

## Context

[x] Chrome Browser 
  Chemotion ELN
Version: 0.8.0
Base revision: 9b86adc3dbf10c16faea6599f327d4ef6d4fbe3d
Current revision: d2ee9f4389486b1dc75ea723e2336486f54e79b1

## Failure Logs / Screenshots
see attached GIF

CamAnNguyen:
- Fix polymer display bug while dragging polymer into reaction Product
- Resolve #441

### DIFF
--- a/app/assets/javascripts/components/models/Sample.js
+++ b/app/assets/javascripts/components/models/Sample.js
@@ -173,6 +173,7 @@ export default class Sample extends Element {
       residues: [],
       imported_readout: ''
     });
+    sample.sample_svg_file = molecule.sample_svg_file
 
     // allow zero loading for reaction product
     sample.reaction_product = (materialGroup == 'products');


Sergeykot:
:+1:

PR #165, opened 2016-09-13
Report can show product-only formula resolve #524

### LINKED ISSUE #524: add hover over information to generic elements' symbols
Referring to a users' comment: "would be good to have additional information besides the symbols as it might be hard to select the right element based on the symbols without further information"
hover over request

JasonYCHuang:
Users can determine to show full formula or product-only formula in report >> configs


Sergeykot:
é

Sergeykot:
:+1:

PR #173, opened 2016-09-19
Move/Assign/Remove associated elements

CamAnNguyen:
- When move, assign, or remove reaction/wellplate/screen, associated will also be processed
- Moving  a screen will also move its wellplates, and samples
-  Removing a screen only remove the screen, not the wellplates


Sergeykot:
:+1:

PR #175, opened 2016-09-20
Remove Reactant/Solvents counter

### LINKED ISSUE #506: Element details tab layout: segment tab to show if data present
As a ELNer 
Who has set my sample-tabs layout to hide most of the tabs globally 
When I open   sample with  segment data
I want  the tab of the segment to show up
so that I don t have to change the tabs layout

CamAnNguyen:
- Reactant and Solvents sample will not increase the counter any more.
- Reactant and Solvents short_label will now be 'reactants' or 'solvents'
- In ReactionDetails, only reactant/solvents IUPAC name with the link to its details.
- Resolve #506


Sergeykot:
:+1:

PR #176, opened 2016-09-20
Add Reaction counter

### LINKED ISSUE #507: Rename /public/welcome-message-sample.md to /public/welcome-message.md to make automatic and local testing easier.
- [ ] I am an ELNer
- [x] I am a testing developer
- [ ] I am a system/ELN admin

# Expected Behavior
Acceptance tests should run with cloning from current GitHub repository w/o additionally changes to precompiled code during tests. Maybe this is also useful for (continuous) deployment.

# Current Behavior
When running acceptance tests remotely with GitHub actions and locally it is necessary to manually rename welcome-message-sample.md into to welcome-message.md in public directory.

# Failure Information 
error failure with old test run on GH actions: https://github.com/ComPlat/chemotion_ELN/runs/3263273589?check_suite_focus=true

## Steps to Reproduce
It is fixed now with addition of extra cp command but would be great if that won't be necessary in the future.

## Solution proposal:

- remove /public/welcome-message.md from .gitignore
- or rename welcome-message-sample.md into to welcome-message.md
- could also be adapted to config files

## Code environment:
ELN version, or branch, or commit:
development-5

CamAnNguyen:
- Add short_label and created_by to Reactions table
- Automatically increase reaction counter on create a new one
- Reactions will now have owner/creator
- Change ReactionDetails title to use short_label as in SampleDetails
- Resolve #507

### DIFF
--- a/app/models/reaction.rb
+++ b/app/models/reaction.rb
@@ -80,9 +80,15 @@ class Reaction < ActiveRecord::Base
 
   has_many :sync_collections_users, through: :collections
 
+  belongs_to :creator, foreign_key: :created_by, class_name: 'User'
+  validates :creator, presence: true
+
   before_save :update_svg_file!
   before_save :cleanup_array_fields
   before_save :auto_format_temperature!
+  before_save :auto_set_short_label, on: :create
+
+  after_create :update_counter
 
   def self.get_associated_samples(reaction_ids)
     ( ReactionsProductSample.get_samples(reaction_ids) +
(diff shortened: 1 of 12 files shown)


app/models/reaction.rb
```diff
@@ -80,9 +80,15 @@ class Reaction < ActiveRecord::Base
...
 
+  belongs_to :creator, foreign_key: :created_by, class_name: 'User'
+  validates :creator, presence: true
+
   before_save :update_svg_file!
   before_save :cleanup_array_fields
   before_save :auto_format_temperature!
+  before_save :auto_set_short_label, on: :create
+
+  after_create :update_counter
```
Sergeykot: does that work fine when copying and cloning reactions?

Sergeykot:
@CamAnNguyen Please check that counter works fine in all cases. For example copy a reaction.
There was some reason why I placed the counter increment not to a AR callback

Sergeykot:
but I can't remember for sure now..

app/models/reaction.rb
CamAnNguyen: Yes, it works whenever a new record is created

Sergeykot:
wait, I did a check for a second time. Yes the counter should work fine now!

PR #181, opened 2016-09-21
Navigation button Regroup

### LINKED ISSUE #493: Export-Import of samples
- I am an ELNer

# Expected Behavior
the file generated by the export of samples can be used as input format to import samples

# Current Behavior
File generated from export cannot be imported directly (selected all info to be exported except "molfile" and "image")

# Failure Information
green message "0 samples were imported successfully" (actually the export/import file contained several samples)

Info: deleted first columns in the exported scheme and then the import worked. Seems to be a bug with columns such as stereo abs/rel, polymer support, and so on...Would be great if someone could check the compatibility of export and import files. It would be great to have information for the user that import files with information on smiles and molfile cannot be imported (the user needs to delete either smiles or molfile).

CamAnNguyen:
- Change 'Generate Report' button icon
- Hide button if it is disabled
- Group Move/Assign, Remove/Delete, Share/Split into Dropdown Button
- Resolves #493


Sergeykot:
does it need to be reviewed/merged?

PR #182, opened 2016-09-22
Fix DeleteAll

### LINKED ISSUE #552: Color code for keys is not consistent
- [ x] I am an ELNer
- [ ] I am a developer
- [ ] I am a system/ELN admin

# Expected Behavior

Function keys like "save", "cancel", "close" on the modules and the ketcher editor should be uniform with respect to their color code

# Current Behavior

On the ketcher structure editor "cancel" is yellow, "create" is blue. However, when creating a new sample "save" is yellow and "cancel" blue.

# Failure Information 

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

## Steps to Reproduce

1. step 1
2. step 2
3. you get it...

## Context

 - [ x] Chrome Browser 
 - [ ] Firefox Browser
 - ELN version, or branch, or commit

## Failure Logs / Screenshots

image
image

CamAnNguyen:
- Fix a bug when deleting all elements in current collection but accidentaly remove every element in other collection
- Resolves #552

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -541,6 +541,3 @@ DEPENDENCIES
   webmock
   whenever
   yaml_db
-
-BUNDLED WITH
-   1.12.5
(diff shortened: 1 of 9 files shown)


Sergeykot:
I think we can merge it as hotfix

PR #184, opened 2016-09-26
add concentration of samples in reactions (merge Mass/Vol) resolve #534 #535 #536

### LINKED ISSUE #534: Private note rework

JasonYCHuang:
merge Mass/Vol to get more place for concentration.

### DIFF
--- a/app/assets/javascripts/components/Material.js
+++ b/app/assets/javascripts/components/Material.js
@@ -3,6 +3,7 @@ import {Radio,FormControl, Button, InputGroup, OverlayTrigger, Tooltip} from 're
 import {DragSource} from 'react-dnd';
 import DragDropItemTypes from './DragDropItemTypes';
 import NumeralInputWithUnitsCompo from './NumeralInputWithUnitsCompo';
+import NumInputWithUnitsTypesCompo from './NumInputWithUnitsTypesCompo';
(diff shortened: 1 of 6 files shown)


Sergeykot:
nan1

concentration - NaN when adding a reaction starting material

Sergeykot:
I am testing it myself, and notice that now reactions yield is calculated when user unfocuses input field
before it was on keyboard event

I have nothing against if those changes were requested :)

app/assets/javascripts/components/Material.js
```diff
@@ -248,9 +261,12 @@ class Material extends Component {
 
   generalMaterial(props, style, handleStyle, inputsStyle) {
     const {material, deleteMaterial, isDragging, connectDragSource,
-           showLoadingColumn } = props;
+           showLoadingColumn, totalVolume } = props;
     const isTarget = material.amountType === 'target'
 
+    const mol = material.amount_mol;
+    const concentration = mol / totalVolume;
```
Sergeykot: totalVolume should not be 0 otherwise we get `NaN`

JasonYCHuang: Done. plz check it again.

Sergeykot:
:+1:

PR #186, opened 2016-09-26
add concentration of samples in reactions (switch Concn/Yield) resolve #534 #535 #184 

### LINKED ISSUE #534: Private note rework

JasonYCHuang:
switch Equiv/Concn


Sergeykot:
rebased from master, otherwise tests will fail

Sergeykot:
:+1:

PR #188, opened 2016-09-27
ElementDetails header refactor

### LINKED ISSUE #554: Exporting literature is not possible currently
The users are trying to export the DOI of the references added to the literature section of a sample into the spreadsheet. It would be great to have just one column there for that. Currently there is no way to link the literature to samples neither by exporting nor by creating a report.

CamAnNguyen:
- Refactor ElementDetails header (CSS and JS)
- Fix the UI bug in create Wellplate/Screen
- Add a tooltip on every button in the header (save, close, generateReport)
- Resolve #554

### DIFF
--- a/app/assets/javascripts/components/ElementCollectionLabels.js
+++ b/app/assets/javascripts/components/ElementCollectionLabels.js
@@ -113,7 +113,7 @@ export default class ElementCollectionLabels extends React.Component {
         </div>
       )
     } else {
-      return (<div></div>)
+      return (<span></span>)
     }
   }
 }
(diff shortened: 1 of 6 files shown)


Sergeykot:
I suggest to add 'tooltip' attribute on each button like this. just add 'tooltip'="save <name_of_item>'
so if user hover mouse there he/she sees the title

Sergeykot:
update: it should be better like 'tooltip'="Save sample" or "Save wellplate"

CamAnNguyen:
Good idea, will add a tooltip for this kind of button

CamAnNguyen:
Code updated, I also update the Pull Request

Sergeykot:
:+1:

PR #191, opened 2016-09-27
Split Sample

### LINKED ISSUE #544: reaction sample coefficient

CamAnNguyen:
- Split sample (using splitButton and split during creating Reaction) will inherit some properties from parent
  - Name: {parent_name}-split (if parent_Name not empty)
  - External_label: {parent_external_label}-split (if parent_external_label not empty)
  - Short_label is updated (exclude reactants and solvents)

- Resolves #544


Sergeykot:
:+1:

PR #192, opened 2016-09-27
add concentration of samples in reactions (5 Colum) resolve #534 #535 #536

### LINKED ISSUE #534: Private note rework

JasonYCHuang:


Sergeykot:
:+1:

PR #193, opened 2016-09-27
Navigation Button revised

### LINKED ISSUE #563: supervisor comment function
Feature requested in the today's ELN FAQ session: 
Can we embed a function to allow comments by supervisors or colleagues in shared or synchronized collections? 
A simple input field is not sufficient as the user should see directly that a comment was placed (e.g. indication in list entries or other message options).

CamAnNguyen:
- Disabled navigation button instead of hiding them
- Resolves #563


Sergeykot:
:+1:

PR #204, opened 2016-10-04
Remake ElementCollectionLabel

### LINKED ISSUE #488: Molecule preview gets clipped
-  I am a system/ELN admin

# Expected Behavior

Molecule should be visible.

# Current Behavior

It's covered by the following row if the browser's width is below 974px (when the "responsive realignment" kicks in ...).

# Failure Information 

image

CamAnNguyen:
- Add new column `is_synchronized` to Collection table, which turn on to synchronized collection. This flag will be automatically update while synchronized collection is created.
- Remake ElementCollectionLabel with a counter for own collections and shared (include synchronized) collections.
- Fix a bug when non-owned collections are displayed inside the label
- Resolves #488 #489

### DIFF
--- a/app/api/api.rb
+++ b/app/api/api.rb
@@ -25,26 +25,18 @@ class API < Grape::API
 
     def group_by_molecule(samples,own_collection = false)
       groups = Hash.new
-      if own_collection
-        samples.each do |s|
-          moleculeName = get_molecule_name(s)
-          serialized_sample = SampleSerializer::Level10.new(s, 10).serializable_hash
-          if !groups[moleculeName]
-            groups[moleculeName] = [].push(serialized_sample)
-          else
-            groups[moleculeName] = groups[moleculeName].push(serialized_sample)
-          end
+      sample_serializer_selector =
+        if own_collection
+          lambda { |s| SampleSerializer::Level10.new(s, 10).serializable_hash }
+        else
+          lambda { |s| ElementPermissionProxy.new(current_user, s, user_ids).serialized }
         end
-      else
-        samples.each do |s|
-          moleculeName = get_molecule_name(s)
-          serialized_sample = ElementPermissionProxy.new(current_user, s, user_ids).serialized
-          if !groups[moleculeName]
-            groups[moleculeName] = [].push(serialized_sample)
-          else
-            groups[moleculeName] = groups[moleculeName].push(serialized_sample)
-          end
-        end
+
+      samples.each do |sample|
+        next if sample == nil
+        moleculeName = get_molecule_name(sample)
+        serialized_sample = sample_serializer_selector.call(sample)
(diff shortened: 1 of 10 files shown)


Sergeykot:
there were some comments from @PiTrem  ?
I am not sure if I can review this, because a lot of logic chages

CamAnNguyen:
Yes, I have updated the code for synchronized collection. Do you remember what did you comment @PiTrem  ?

app/api/api.rb
```diff
@@ -25,26 +25,18 @@ def is_public_request?
...
-          end
+      sample_serializer_selector =
+        if own_collection
+          lambda { |s| SampleSerializer::Level10.new(s, 10).serializable_hash }
+        else
+          lambda { |s| ElementPermissionProxy.new(current_user, s, user_ids).serialized }
         end
+
+      samples.each do |sample|
+        next if sample == nil
```
Sergeykot: better not to skip if array item is nil, but use `Array#compact` method that removes empty items.
Of course if you don't need nil items in array

app/models/sync_collections_user.rb
```diff
@@ -7,4 +7,12 @@ class SyncCollectionsUser < ActiveRecord::Base
   has_many :reactions, through: :collection
   has_many :wellplates, through: :collection
   has_many :screens, through: :collection
+
+  before_create :auto_set_synchronized_flag
+
+  def auto_set_synchronized_flag
+    c = self.collection
```
Sergeykot: `self.collection.update_attribute!(is_sync...., true) if self.collection.present?`

db/migrate/20161004121244_add_is_synchronized_to_collections.rb
```diff
@@ -0,0 +1,12 @@
+class AddIsSynchronizedToCollections < ActiveRecord::Migration
+  def change
+    add_column :collections, :is_synchronized, :boolean, null: false, default: false
+    Collection.reset_column_information
+
+    SyncCollectionsUser.find_each do |sc|
```
Sergeykot: self.collection.update_attribute!(is_sync...., true) if self.collection.present?

Sergeykot: the same but not literally

Sergeykot:
I mean from the code style looks almost good, but I am lost in this logic :)

Sergeykot:
can be merged after little code style changes

app/api/api.rb
CamAnNguyen: This is not an array, its type is `ActiveRecord_AssociationRelation` so that I cannot use `compact` here. Do you have any other idea beside of `compact` ?

Sergeykot: to_a.compact :)

Sergeykot: but lets leave it like this now..

Sergeykot:
:+1:

PR #208, opened 2016-10-05
relocate add analysis button

### LINKED ISSUE #593: Adding elements to "All" collection automatically after accepting ownership of a synchronized collection
As a user I would like to see all elements in my All collection after I have accepted th ownership of a collection that was synchronized with me.
Currently, if someone passes the ownership of her/his collection to me, I can appect it and then it appears in my collection list. However, the data of this collection is not automatically copied to the All collection. This should be done automatically when accpeting the ownership of a collection.

PiTrem:
fully resolve #593


Sergeykot:
:+1:

PR #216, opened 2016-10-05
Edit/Close Sample with WellplateDetail

### LINKED ISSUE #603: Color code for keys is not consistent
- [x] I am an ELNer
- [ ] I am a developer
- [x] I am a system/ELN admin

# Description
Currently, keys with the same functionality have different names (close, cancel) and different colors.

To deal with this issue, the following convention is suggested (As most keys in the ELN follow it):

- Generating data: Dark blue (keys like "Add")
- Modifying data: yellow (keys like "save", "update") 
- Losing data: red (keys like "delete", "cancel")

This file has a summary of most [buttons in the ELN](https://docs.google.com/spreadsheets/d/1_SltG3_423LVUbUb9TEipLP7BFrNJVJwbgwKVTjVJms/edit?usp=sharing)

Buttons that need to change:

- toolbar: remove elements (yellow) -> red
- Sample, reaction... modal: Close (blue) -> red
- Ketcher: Save (dark blue) -> yellow
- Ketcher: Cancel (yellow) -> Close (red)

Optional: To comply with the above description, One can restrict the three colors for the described purposes. (e.g. buttons like synchronize collections are not in blue any more)
Also one can restrict the purposes to the three colors (e.g., buttons like add data set become in blue instead of green).

Possible solutions are provided here:
image
image
image

Additionally, the relative location of cancel and save buttons needs discussion

CamAnNguyen:
- We can edit sample, close it then back to wellplate
- Resolves #603

### DIFF
--- a/app/assets/javascripts/components/SampleDetails.js
+++ b/app/assets/javascripts/components/SampleDetails.js
@@ -140,41 +140,45 @@ export default class SampleDetails extends React.Component {
   }
 
   submitFunction() {
-  let {sample} = this.state;
-  let { currentReaction } = ElementStore.getState();
+    let {sample} = this.state
+    let { currentReaction, currentWellplate } = ElementStore.getState()
 
-  if(currentReaction) {
+    if(currentReaction) {
       if(sample.isNew) {
-        ElementActions.createSampleForReaction(sample);
+        ElementActions.createSampleForReaction(sample)
       } else {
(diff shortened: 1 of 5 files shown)


Sergeykot:
:+1: 
tested it locally, works well!

PiTrem:
:+1:

PR #219, opened 2016-10-06
Analysis update while editing

### LINKED ISSUE #607: Generic elements/system defined: fix Enzyme activity unit
- [x] I am an ELNer
- [ ] I am a developer
- [x] I am a system/ELN admin

Currently the provided unit for enzyme activity is U/L, while it should be "U" or "IU" (the same). Also conversions to  "katal" (kat) which is the preferred unit now would be great to have, along with its submultiples (millikatal mkat, microkatal µkat, nanokatal nkat).

1 U = 16.67 nkat

CamAnNguyen:
- Update analysis while editing
- Resolves #607

### DIFF
--- a/app/assets/javascripts/components/AnalysisDatasets.js
+++ b/app/assets/javascripts/components/AnalysisDatasets.js
@@ -16,6 +16,12 @@ export default class AnalysisDatasets extends Component {
     };
   }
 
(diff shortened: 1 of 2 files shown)


Sergeykot:
:+1:

PR #225, opened 2016-10-06
Revise Search

### LINKED ISSUE #596: Generic elements: Enhance integer fields by adding a range
- [ ] I am an ELNer
- [ ] I am a developer
- [ ] I am a system/ELN admin

# Description
Currently there is integer field in the generic elements where the user can enter any number. It would be great to make it possible for the admins to decide a range for the numbers to limit the errors. For example, a field for the pH should not accept 25 as an input.

CamAnNguyen:
- Reduce number search results by increase the threshold
- Add external_label for searching
- Fix substring search
- Resolves #596 #597

### DIFF
--- a/app/api/chemotion/search_api.rb
+++ b/app/api/chemotion/search_api.rb
@@ -13,7 +13,7 @@ module Chemotion
       end
 
       def structure_search
-        params[:selection].molfile == nil ? false : true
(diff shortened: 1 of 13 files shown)


PiTrem:
:+1:

Sergeykot:
:+1:

PR #228, opened 2016-10-07
Well popover stays in a correct position after scrolling. resolve #616

JasonYCHuang:
1. make the well popover always points to the corresponding well after scrolling.
2. css update for panel-detail.


Sergeykot:
:+1:

PR #230, opened 2016-10-10
CollectionTree enhanced

### LINKED ISSUE #631: Include pdfs
- [ x] I am an ELNer
- [ ] I am a developer
- [ ] I am a system/ELN admin

# Description

1. It would be great if one could include pdfs into research plans, in the same way as images. I do not include an image as my wish would be that the space where I can drop my image should also accept pdfs.

2. Often protocols are shared as pdfs and not as images. Therefore, one has to copy the entire text into the research protocols.

CamAnNguyen:
- Save expanded collection
- Fix a bug while colelction and synchronized collection with the same id are clicked
- Resolves #631, #632


Sergeykot:
:+1:

PR #235, opened 2016-10-10
reaction SVG solvents are 15 characters long. resolve #635

JasonYCHuang:


lib/svg/reaction_composer.rb
```diff
@@ -26,11 +26,25 @@ def initialize(materials_svg_paths, options = {})
...
+    end
+
+    def num_solvents_per_line
+      reactants_size = @reactants.size > 0 ? @reactants.size : 1
+      reactants_size > 3 ? 3 : reactants_size
+    end
+
+    def solvents_string(group)
+      group.map do |j|
+        if j && j.size > 15
```
Sergeykot: magic number

JasonYCHuang:
wait a minute. Rspec update needed.

lib/svg/reaction_composer.rb
JasonYCHuang: When the string is longer than 15 characters, it will be limited to 15 characters with ```..```.
This is set by Nicole.

JasonYCHuang: done

PR #237, opened 2016-10-10
Display reactant name

### LINKED ISSUE #650: default settings for structure search options
Hi all, it would be great if someone could change the default settings in the search window for structure search. Currently, we have similarity search as default, but substructure search (RDKit or Openbabel) is most feasible. This change will avoid some problems with the search function. 
Also: it would be great to have the option "exact search"

CamAnNguyen:
- Reactant display order: iupac-name > name > sum_formular
- Resolves #650


Sergeykot:
:+1:

PR #249, opened 2016-10-11
Disabled polymer density and melting-point

### LINKED ISSUE #638: ChemSpectra conversion time
Please check how we could accelerate the processing time with chemspectra. Upload and processing of analytical data takes too long in my opinion

CamAnNguyen:
- Disabled polymer density and melting-point
- Resolves #638


Sergeykot:
:+1:

PR #252, opened 2016-10-11
Reaction Drag and Drop

CamAnNguyen:
- Re-factor Sample and Reaction js model
  + Keep sample logic (copy, split, new) in Sample model
  + Keep reaction logic (drag sample to starting, reactant/solvent, product) in Reaction model
- Rebuild logic while moving between MaterialGroup
- If create new Product through Reaction create button, product name must be manually input
- Resolves #649

### DIFF
--- a/app/assets/javascripts/components/ReactionDetailsScheme.js
+++ b/app/assets/javascripts/components/ReactionDetailsScheme.js
@@ -23,26 +23,12 @@ export default class ReactionDetailsScheme extends Component {
 
   dropSample(sample, materialGroup, external_label) {
     let {reaction} = this.state;
-    let splitSample ;
+    let splitSample;
 
     if (sample instanceof Molecule || materialGroup == 'products'){
       // Create new Sample with counter
-      splitSample =
-        Sample.buildReactionSample(reaction.collection_id,
-          reaction.temporary_sample_counter, materialGroup, sample)
-
-      if (materialGroup == 'products') {
-        let productsCount = reaction.products.length
-        splitSample.name = reaction.short_label + "-" +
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/components/models/Sample.js
```diff
@@ -60,48 +60,71 @@ export default class Sample extends Element {
...
-    splitSample.target_amount_value = 0;
-    splitSample.real_amount_value = null;
-    splitSample.is_split = true;
-    splitSample.is_new = true;
     return splitSample;
   }
 
   buildChildWithoutCounter() {
-    let splitSample = this.clone();
+    let splitSample = this;
```
Sergeykot: will cause an error

Sergeykot: change back to clone() if you dont have a reason for that

Sergeykot:
others looks good, however I can not test it now.

app/assets/javascripts/components/ReactionDetailsScheme.js
```diff
@@ -23,26 +23,14 @@ export default class ReactionDetailsScheme extends Component {
...
-        splitSample.name = reaction.short_label + "-" +
-          String.fromCharCode('A'.charCodeAt(0) + productsCount)
-      }
-
-      if (sample.external_label)
-        splitSample.external_label = sample.external_label
-      if (sample.elemental_compositions)
-        splitSample.elemental_compositions = sample.elemental_compositions
-
-    } else if (sample instanceof Sample){
```
JasonYCHuang: Where are  ```splitSample.external_label```, ```splitSample.elemental_compositions```? line(34-43)

Or do you think they should not be assigned to ```splitSample```?

PR #254, opened 2016-10-12
set values to nd for polym samp (density , bl pt, ml pt) reresolve #638

PiTrem:
set values to n.d instead 0.000 for disabled input boiling pt, melting pt, density of SP supported samples


Sergeykot:
:+1:

Sergeykot:
I see `not defined` for Polymer

PR #257, opened 2016-10-13
adjust reaction scheme table

JasonYCHuang:
1. adjust style for reaction scheme table
2. input padding smaller
3. align T/R of starting materials, reactants, products
4. no space between number and ```%``` in equiv/yield


Sergeykot:
tested

PR #263, opened 2016-10-17
Fix padding on sample view resolve #660

### LINKED ISSUE #660: Customisable Report templates
- [] I am a system/ELN admin

# Description
Please provide a description of:
Some of our users have stated that the current report templates are not entirely suitable for their requirements and have to manually change / delete entries for each report when generating a sup info. We tried editing the report templates ourselves (e.g. supporting_information.docx) but were unsuccessful (for us at least, it was too complicated). For our users it is extremely important that they can create their own templates. Could you please add a feature where users can create their own templates? 

If this is not doable in the short term, would it be possible at least for our users to receive 1-2 templates which better suit their needs?

Best regards,

John

Sergeykot:


Sergeykot:
adding

PR #270, opened 2016-10-18
Make Reaction temperature a time series

### LINKED ISSUE #550: Include data from custom generic  segments into the PDF/Word file export
We have started using the generic segments functionality which has been really useful. However, we found that the data is not actually exported in the Word/PDF files, at least not by default in v0.9.1 of Chemotion. Has this already been implemented by any chance in a newer version? If not, would it be possible to implement this to have this data also exported?

CamAnNguyen:
- Change Reaction temperature field to a Hash
- Add a temperature chart based on temperature field
- User still can input a static temperature field
- Add a temperature table to edit new temperature field
- Add a button to hide/display temperature chart
- Add a button to change temperature unit.
- User are able to convert between Celcius, Fahrenheit, and Kelvin degree
- Limit temperature table length
- Resolves #550, #551

### DIFF
--- a/app/assets/stylesheets/elements_details.css
+++ b/app/assets/stylesheets/elements_details.css
@@ -33,3 +33,89 @@
   top: 50%;
   left: 50%;
 }
+
+#line-chart {
+  width: 100%;
+}
+
+#line-chart .plot {
+  fill: rgba(250, 250, 255, 0.6);
+}
+
+#line-chart .grid .tick {
+  stroke: lightgrey;
+  opacity: 0.7;
+}
+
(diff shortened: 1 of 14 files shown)


app/assets/stylesheets/elements_details.css
```diff
@@ -33,3 +33,89 @@
   top: 50%;
   left: 50%;
 }
+
```
Sergeykot: if I comment out changes in this file reaction scheme is OK, please check

Sergeykot:
works, tested

PR #274, opened 2016-10-20
Apply temperature convert to user text

CamAnNguyen:
- If user enter text, keep original. If user enter a number, treat it as temperature:
  + Display unit in SVG
  + Apply unit conversion on entered text
- Increase chart range: 10% of total variability. For example: if the temperature change from 1 to 100. We have: (100 - 1) * 0.1 = 9.9,  the chart range is from 8.9 to 109.9. With this way, the temperature line can fit within the chart
- Resolves #667 #668 #669

### DIFF
--- a/app/assets/javascripts/components/ReactionDetails.js
+++ b/app/assets/javascripts/components/ReactionDetails.js
@@ -92,7 +92,11 @@ export default class ReactionDetails extends Component {
     }).filter(s => s)
 
     const solventsArray = solvents.length !== 0 ? solvents : [reaction.solvent]
-    const temperature = reaction.temperature_display
+    let temperature = reaction.temperature_display
+    if (/^\-?\d*\.{0,1}\d{0,2}$/.test(temperature)) {
+      temperature = temperature + " " + reaction.temperature.valueUnit
+    }
+
     ElementActions.fetchReactionSvgByMaterialsSvgPaths(materialsSvgPaths, temperature, solventsArray);
   }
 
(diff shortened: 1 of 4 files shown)


Sergeykot:
:+1:

app/assets/javascripts/components/ReactionDetails.js
```diff
@@ -92,7 +92,11 @@ export default class ReactionDetails extends Component {
     }).filter(s => s)
 
     const solventsArray = solvents.length !== 0 ? solvents : [reaction.solvent]
-    const temperature = reaction.temperature_display
+    let temperature = reaction.temperature_display
+    if (/^\-?\d*\.{0,1}\d{0,2}$/.test(temperature)) {
```
PiTrem: temp like "0.12345" should work
also some people may input "0,12345   "
`/^\-?\d+(\.|,){0,1}\d*\s*$/`

app/assets/javascripts/components/models/Reaction.js
```diff
@@ -135,6 +135,14 @@ export default class Reaction extends Element {
         break
     }
 
+    // If userText is number only, treat as normal temperature value
+    if (/^\-?\d*\.{0,1}\d{0,2}$/.test(temperature.userText)) {
```
PiTrem: will have to be the same regexp as in ReactionDetails. (needs to substitute potential  ',' to '.')

app/assets/javascripts/components/models/Reaction.js
```diff
@@ -135,6 +135,14 @@ export default class Reaction extends Element {
...
+      temperature.userText =
+        convertFunc(newUnit, temperature.userText).toFixed(2)
+
+      return temperature
+    }
+
     temperature.data.forEach(function(data, index, theArray) {
       theArray[index].value = convertFunc(newUnit, data.value).toFixed(2)
```
PiTrem: if you Fixed(2) every time you change unit, you loose precision. displayed value should be Fixed(2) but stored value should have more

PR #275, opened 2016-10-20
Hotfix temperature conversion

CamAnNguyen:


Sergeykot:
TODO: regexp `/^[\-|\d]\d*\.{0,1}\d{0,2}$/`
should be in one place (maybe Functions.js)
but

it is hotfix so we accept it..

Sergeykot:
:+1:

CamAnNguyen:
Agree! I will put js regex in one place. Do you have any suggestion about ruby regex? How can we keep both ruby and js in one place ?

Sergeykot:
we can make some settings with and get them JSON request.
but at least regexp should be in 1 place on backend

Sergeykot:
and on frontend

Sergeykot:
UPD: ideally such settings that have to be handled both on server and front-end are handled with something like `Settings` gem and served in .js with help of .erb

CamAnNguyen:
Thank for your suggestion. Currently, we have many of these settings in Sample and Reaction model in both front-end and back-end.

PR #279, opened 2016-10-24
Ruby 2.3.1

Sergeykot:
Upgrade Ruby as recommended

ruby-versio


PiTrem:
travis.yml to be updated as well

Sergeykot:
ok

PR #280, opened 2016-10-24
Report order dnd

JasonYCHuang:
1. refactor report local state to store
2. add Report-Order tab which let you drag-n-drop


Sergeykot:
:+1:

app/assets/javascripts/components/report/OrdersDnD.js
```diff
@@ -0,0 +1,74 @@
+import React, {Component} from 'react'
+import SVG from 'react-inlinesvg';
+import classnames from 'classnames';
+import { DragSource, DropTarget } from 'react-dnd';
+import { compose } from 'redux';
```
PiTrem: do we need  redux just for this one function?

PR #281, opened 2016-10-25
Add editor to Reaction description

CamAnNguyen:
- Add Quill editor to Reaction description
- Reaction is updated when user finish typing (after 300ms)
- Report use raw text, not the formatted description
- Resolves #680


Sergeykot:
> Reaction is updated when user finish typing (after 300ms)

not work for me unless I press 'Save'

CamAnNguyen:
> Reaction is updated when user finish typing (after 300ms)

With this line, I mean that reaction is updated (considered as changed, display floppy, broadcast description changes to other components) only when the user finishes typing. It doesn't save to the DB when the user hasn't pressed "Save"

Sorry for this vague explanation

PiTrem:
:+1:

PiTrem:
:+1:

PR #290, opened 2016-11-09
Fix reaction samples naming

Sergeykot:
The problem before was that we did not refresh counter so had to use variable `temporary_samples_counter` in reaction. Now when we create a sample we refresh current user so don't need to use this counter.

**Question:** do I have to change old sample `short_label`-s ?


PiTrem:
:+1:

PR #293, opened 2016-11-14
Fix element refresh on change collection

Sergeykot:


Sergeykot:
Alternative solution is to check if the current element if the same that we get from server and not update it in order not to loose the data

Sergeykot:
code updated

Sergeykot:
ok this solution works. however when I investigated a bit found that I can fix the method `componentWillReceiveProps` to reaction and it fixes the behaviour of it. However for sample same method does not work. I would suggest to merge this solution

Sergeykot:
ok this to be refactored by me now, please do not merge

PR #295, opened 2016-11-15
Fix for update reaction material purity: recalculate amount_mmol

Sergeykot:
* recalculate amount when amount type is target
* recalculate equivalent when amount type is real
* fix bug for target/real switch for reaction product


PiTrem:
:+1:

PR #297, opened 2016-11-28
Reaction label report fix

Sergeykot:


Sergeykot:
deployed and tested on server

PR #305, opened 2016-12-12
Selected collection is not "selected"

CamAnNguyen:
- After clear search condition, selected collection CSS is not considered as "selected"


Sergeykot:
tested it :+1:

PiTrem:
:+1:

PR #306, opened 2016-12-12
Samples action in Molecule Sort

CamAnNguyen:
- While in molecule sorting: Sample managements (create, delete, move, split ...) will no longer back to samples sorting


Sergeykot:
:+1: tested

PR #311, opened 2016-12-20
Fix not defined error on collection management

Sergeykot:


Sergeykot:
Fixes following when trying to share/move element to collection

sharing_bug

PiTrem:
:+1:

PR #312, opened 2016-12-20
report-docx can access CDraw cdx files

JasonYCHuang:
Under __windows__ environment with __ChemDraw software installed__, users can access cdx files by clicking reaction equations in reports.


Sergeykot:
PNG image spec failing

Sergeykot:
:+1:

PR #313, opened 2016-12-20
User can now change tab layout

CamAnNguyen:
- Add a button next to final tab to open tab layout
- Drag and drop to change tab order 
- Drag tab to "hidden area" to hide tab
- Change will be applied when exiting layout editor
- Tab layout is saved in user table

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161109141353) do
+ActiveRecord::Schema.define(version: 20161215133014) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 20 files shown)


db/schema.rb
```diff
@@ -358,6 +371,28 @@
   add_index "samples", ["molecule_id"], name: "index_samples_on_sample_id", using: :btree
   add_index "samples", ["user_id"], name: "index_samples_on_user_id", using: :btree
 
+  create_table "scifinding_credentials", force: :cascade do |t|
```
Sergeykot: these tables are odd
`rake db:drop && rake db:create && rake db:migrate` and then updating schema.rb will help

Sergeykot: @CamAnNguyen @camannguyenthanh

CamAnNguyen: Thank you, just updated the `schema.rb`

Sergeykot:
tested :+1:

PR #331, opened 2017-01-13
Fix Research Plan sharing & spec

Sergeykot:


Sergeykot:
merged as hotfix for important functionality

PR #353, opened 2018-06-19
Reflect Apple's rebranding

katrinleinweber:
See diff ;-) I'm not sure about the [path on L10](https://github.com/ComPlat/chemotion_ELN/blame/c42191fbaf56c684b263597b7f4c2e3b6348f50b/INSTALL.md#L10), but the OS mentions in the text could be updated, if none of the commands requires El Capitan or older. Cheers!

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,16 +19,16 @@ if there are errors with nokogiri compilation with new xcode7:
 * openbabel will be automatically installed through `bundle install`
 
 ## rmagick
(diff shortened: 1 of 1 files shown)


katrinleinweber:
[This build error](https://travis-ci.org/ComPlat/chemotion_ELN/builds/394066834#L2704-L2711) seems to be [more general](https://travis-ci.org/ComPlat/chemotion_ELN/pull_requests) than from my PR only.

PR #388, opened 2021-04-01
Create main.yml

Ira-Maier:
convert travis to GitHub Actions


Ira-Maier:
Rails 5:
[x] npm test
[x] ruby test
[ ] acc test

ptrxyz:
Hm, this messes with our git history quite a lot. I suggest merging the "non-working" commits into a new one. To keep things clean...

Ira-Maier:
> Hm, this messes with our git history quite a lot (248 commits). I suggest merging the commits into a few new ones to keep things clean.

Yes, but actually just 5 files have meaningful changes, so I could create a new clean branch when every test is passing. But at the moment there are still issues with acceptance tests, so work is in progress.

PR #469, opened 2021-07-08
separated versions to fix #468

### LINKED ISSUE #468: Moving version information to a separate file to be able to track them more easily
To improve tracking the current/used versions of npm, nvm, node, ruby, etc between installations/updates and when it comes to build the docker containers, I suggest to move the version information to a separate env-file.

ptrxyz:
Suggested fix for #468, using versions from 0.9.1


PiTrem:
obsolet since packing is using docker

PR #489, opened 2021-07-30
Fix #488

### LINKED ISSUE #488: Molecule preview gets clipped

ptrxyz:
Should fix #488. It's not clear to me if this component is used somewhere else, but I assume there is no need at all for an element style. Couldnt the style property on `<row>` be removed all together?

### DIFF
--- a/app/packs/src/components/SampleDetails.js
+++ b/app/packs/src/components/SampleDetails.js
@@ -517,7 +517,7 @@ export default class SampleDetails extends React.Component {
   }
 
   sampleInfo(sample) {
-    const style = { height: '200px' };
+    const style = { min-height: '200px' };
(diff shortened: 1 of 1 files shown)


app/packs/src/components/SampleDetails.js
```diff
@@ -517,7 +517,7 @@ export default class SampleDetails extends React.Component {
   }
 
   sampleInfo(sample) {
-    const style = { height: '200px' };
+    const style = { min-height: '200px' };
```
PiTrem, line 520: did you mean minHeight ?

PiTrem (changes requested):
min-height vs minHeight

PR #549, opened 2021-10-05
Inline edit

jochenklar:
* Extend sample, reaction, wellplate, and screen API to provide full models as list for inline editing
* Update front end for inline edit functionality
*


PiTrem:
image

PR #561, opened 2021-10-19
Fix #560

### LINKED ISSUE #560: fixing js imports in postinstall script fails with alternative NODE_PATH.
When installing node modules to a different path than ./node_modules, the `package_postinstall.sh` script will fail sind the paths are hard coded. Node/Yarn/npm however allow different module paths (i.e. through `NODE_PATH`, `yarn global` or `yarn install --modules-folder`, etc).

I suggest to resolve the module files using [require.resolve](https://nodejs.org/api/modules.html#requireresolverequest-options).

ptrxyz:
Simple fix for #560.

### DIFF
--- a/package_postinstall.sh
+++ b/package_postinstall.sh
@@ -4,11 +4,12 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


PiTrem:
:fast_forward:

PR #585, opened 2021-11-23
Orcid login

jochenklar:
This PR adds [OmniAuth](https://github.com/omniauth/omniauth) support to Chemotion ELN to login with external providers, namely [ORCID](https://orcid.org/).

* Add OmniAuth configuration in `config/initializers/devise.rb` and a `omniauth.yml` to hold the OAuth2 parameters
* Add GitHub login
* Add custom ORCID login based on omniauth-orcid, which should support both the public and the member API
* Add login links to the home screen and the corresponding react/alt machinery.

The list of providers is customizable, in principle (with a few edits) all Oauth2 providers should be suitable.


jochenklar:
I added 3 gems:

```
gem 'omniauth', '~> 1.9.1'
gem 'omniauth-github', '~> 1.4.0'
gem 'omniauth-orcid', git: 'https://github.com/datacite/omniauth-orcid'
```

and 2 fields to `User`: `omniauth_provider`, `omniauth_uid`.

Issue #574, opened 2021-11-16
Flaky acceptance tests


JanCBrammer:
Some acceptance tests seem to be flaky (sometimes passing, sometimes failing, in the abscence of code changes).

Tests suspected of flakiness 🕵️‍♀️❄️:

rspec ./spec/features/sync_collection_spec.rb:59 # Sync Collection Test Sync collection with write permission
rspec ./spec/features/sync_collection_spec.rb:73 # Sync Collection Test Sync collection with read permission
rspec ./spec/features/copy_reaction_spec.rb:71 # Copy reactions to diff collection

JanCBrammer:
Useful links (for future reference):
* http://johnpwood.net/2015/04/09/fix-flaky-feature-tests-by-using-capybaras-apis-properly/
* http://joshfrankel.me/blog/lessons-learned-from-using-capybara-for-feature-testing/
* https://thoughtbot.com/blog/write-reliable-asynchronous-integration-tests-with-capybara

JanCBrammer:
Currently fixing this on https://github.com/ComPlat/chemotion_ELN/tree/end_to_end_tests

JanCBrammer:
Closing with af0f658025e6819db451c31037eef6c917a5b0aa.

PR #606, opened 2021-12-15
Attempt to fix flaky acceptance tests.

JanCBrammer:
For context see #574.

### DIFF
--- a/spec/features/copy_reaction_spec.rb
+++ b/spec/features/copy_reaction_spec.rb
@@ -2,115 +2,107 @@
 
 require 'rails_helper'
 
-describe 'Copy reactions' do
-  let!(:user) { create(:user, first_name: 'Hallo', last_name: 'Complat', account_active: true, confirmed_at: Time.now) }
-  let!(:user2) { create(:user, first_name: 'User2', last_name: 'Complat', account_active: true, confirmed_at: Time.now) }
+describe 'Copy reaction' do
+  let!(:user1) { create(:user, first_name: 'User1', last_name: 'Complat', name_abbreviation: 'US1', account_active: true, confirmed_at: Time.now) }
+  let!(:user2) { create(:user, first_name: 'User2', last_name: 'Complat', name_abbreviation: 'US2', account_active: true, confirmed_at: Time.now) }
+
   let!(:m1) { create(:molecule, molecular_weight: 171.03448) }
   let!(:m2) { create(:molecule, molecular_weight: 133.15058) }
-  let(:material) { create(:sample, name: 'Material', target_amount_value: 7.15, molecule: m1) }
-  let(:product) { create(:sample, name: 'Product', real_amount_value: 4.671, molecule: m2) }
-  let(:reaction) { create(:reaction, status: 'Successful', short_label: 'Reaction1') }
+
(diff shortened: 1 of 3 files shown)


spec/features/copy_reaction_spec.rb
```diff
@@ -2,115 +2,107 @@
...
-
-    CollectionsSample.find_or_create_by!(sample: material3, collection: cshare2)
-    CollectionsSample.find_or_create_by!(sample: product3, collection: cshare2)
-    CollectionsReaction.find_or_create_by!(reaction: reaction3, collection: cshare2)
-    ReactionsStartingMaterialSample.create!(reaction: reaction3, sample: material3, reference: true, equivalent: 1)
-    ReactionsProductSample.create!(reaction: reaction3, sample: product3, equivalent: 1)
-
-  end
-
-  it ' new reaction', js: true do
```
PiTrem, line 54: is this one gone?

spec/features/sync_collection_spec.rb
```diff
@@ -3,83 +3,58 @@
...
-    create_reaction
-    sync_collection('Write')
-    first('span.glyphicon-log-out').click
-    find_by_id('user_login').set('US2')
-    find_by_id('user_password').set('iamuser2')
-    find('input[type="submit"]').click
-    find_by_id('synchron-home-link').click
-    find_all('span.glyphicon-plus')[0].click
-    find_by_id('tree-id-sync_test').click
-    update_reaction
```
PiTrem, line 69: updating reaction is gone.  Do yo think it should be moved in another test?

PiTrem (commented):
Nice. just 2 small questions..

spec/features/copy_reaction_spec.rb
JanCBrammer: Yes, I removed it because as far as I can tell it only asserts that the reaction detail dialog isn't open. This assertion is a bit redundant because it's also (implicitely) tested in the other `it` blocks.

spec/features/sync_collection_spec.rb
JanCBrammer: I removed the `update_reaction` function because it's used to assert if a user can manipulate fields in the reaction dialog if they have write permission. This is now replaced by asserting that the dialog fields are disabled with read permission and enabled with write permission. I find the latter way easier and more uniform (across test cases).

spec/features/copy_reaction_spec.rb
PiTrem: I think the reactionDetail panel shoulb be  there, but because the reaction is new and not existing yet in the DB, there is no option to copy it

PiTrem: right I see, something has been missing all along.
I think the `first('i.icon-reaction').click`  line should have been duplicated and run twice.
- 1st click is to switch the list pannel to display the reaction list => that will  change the add-element button on the page-top action bar to an add-reaction button  => that Button will then contain the first(i.icon-reaction ) of the page
- 2nd click is to build a new reaction => new built dont have the copy button until after created (the purpose of the this test I think)
I though the current line first('i.icon-reaction').clik was targetted at the reaction icon to add a reaction. But this one is not visible unless you are on the reaction list. so as of now this test is unlikely to ever fail  .

PiTrem (approved):

PR #736, opened 2022-05-25
Fix #735: harmonized mail settings

### LINKED ISSUE #735: Mail settings in container environments.
I suggest to have an easy toggle for mail settings in production + development configurations in containerized environments. I envision something like this:

- read the SMTP settings using environment variables in development config just like in production environments, however set sane defaults in case the environment is not set.
- Currently SMTP_HOST ist used to build public URLs. That's more often incorrect than it is correct, so I suggest to set it to some dedicated value like i.e. URL_HOST.
- If `SMTP_ADDRESS` is not set or `DISABLE_MAIL_DELIVERY` is set, mail delivery should be disabled completely.

ptrxyz:
Harmonized mail settings as described in #735:
- action_mail now builds it's URL based on the environment variable HOST_URL.
- HOST_URL defaults to "http://localhost" in dev environments and to "https://localhost:443" in production environments.
- If no SMTP_ADDRESS is set or if DISABLE_MAIL_DELIVERY is set, mail delivery is disabled completely.

Note: HOST_URL is intended to be used in container environments to point to the public web entrypoint of the Chemotion instance.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -1,3 +1,5 @@
+require "uri"
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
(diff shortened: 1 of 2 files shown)


ptrxyz:
Closed: I guess an initializer would be better here.

PR #737, opened 2022-05-25
Fixing #735 using initializer

ptrxyz:
- action_mail now builds it's URL based on the environment variable
  HOST_URL or "http://localhost" if not set.
- If no SMTP_ADDRESS is set or if DISABLE_MAIL_DELIVERY is set, mail delivery is disabled completely.

Note: HOST_URL is intended to be used in container environments to point to the public web entrypoint of the Chemotion instance.

### DIFF
--- a/config/initializers/mail.rb
+++ b/config/initializers/mail.rb
@@ -1,0 +1,27 @@
+require "uri"
+
+Rails.application.configure do
+  uri = URI.parse("#{ENV['HOST_URL'] || 'http://localhost'}")
+  scheme = uri.scheme   || 'http'
+  host   = uri.host     || 'localhost'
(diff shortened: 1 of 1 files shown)


PiTrem (changes requested):
* corresponding config lines should be removed from config/environments/{production|development}.rb

* the whole smtp should  be skipped for Rails.env.test?

config/initializers/mail.rb
```diff
@@ -0,0 +1,27 @@
+require "uri"
+
+Rails.application.configure do
+  uri = URI.parse("#{ENV['HOST_URL'] || 'http://localhost'}")
```
PiTrem, line 4: HOST or HOST_URL ?

PR #739, opened 2022-05-27
Elemental composition values now have 2 decimal places

### LINKED ISSUE #721: In CHN analysis reports, all values should be given with two relevant digits
- [x] I am an ELNer
- [ ] I am a developer
- [x] I am a system/ELN admin

# Expected Behavior

When reporting CHN analysis data, all values should be given with two relevant digits, even if the digits to the right of the decimal separator are zero, since these are precisely known while digit not shown means "unknown"

# Current Behavior

Values are incorrectly given as 76.9 instead of correctly as 76.90 (and possibly also 76 instead of 76.00)

# Failure Information 

Insufficient number of digits to right of decimal separator given although they are precisely known

## Context

 - OS: Windows
 - [ ] Chrome Browser 
 - [ x] Firefox Browser
 - ELN version, or branch, or commit: 1.13

## Failure Logs / Screenshots

Include any relevant log snippets or files here.
Screenshot 2022-05-20 at 17-24-48 Chemotion

dp19982:
CHANGELOG: Closes #721

Significant zeroes are added to elemental compositional values, to ensure that all values are 2 decimal places.

The initial example (2,2-bipyridine) is now:
bipyridine

Example 2 (C10H18MnO6):
Original
old1
New
new1

### DIFF
--- a/app/packs/src/components/ElementalComposition.js
+++ b/app/packs/src/components/ElementalComposition.js
@@ -26,7 +26,7 @@ export default class ElementalComposition extends React.Component {
           </strong>
           &nbsp;
           &nbsp;
-          {value}
+          {Number(value).toFixed(2)}
           &nbsp;
           &nbsp;
         </span>
--- a/app/packs/src/components/ElementalComposition.js
+++ b/app/packs/src/components/ElementalComposition.js
@@ -19,14 +19,6 @@ export default class ElementalComposition extends React.Component {
 
     keys.map(function(key, index) {
       let value = elemental_composition.data[key];
-      // adding significant figures to value if needed - no decimal places, add .00; 1 decimal place, add 0;
-      const regexp = /^\d+$/; // regex for an integer (no decimal places)
-      const regexp1 = /\.\d{0,1}$/; // regex to search for a value with ONLY ONE decimal place
-      if (value.match(regexp)) {
-        value += '.00'
-      } else if (value.match(regexp1)){
-        value += '0'
-      }
       elements.push(
         <span className="data-item" key={key}>
           <strong>
--- a/app/packs/src/components/ElementalComposition.js
+++ b/app/packs/src/components/ElementalComposition.js
@@ -19,6 +19,14 @@ export default class ElementalComposition extends React.Component {
 
     keys.map(function(key, index) {
       let value = elemental_composition.data[key];
+      // adding significant figures to value if needed - no decimal places, add .00; 1 decimal place, add 0;
+      const regexp = /^\d+$/; // regex for an integer (no decimal places)
+      const regexp1 = /\.\d{0,1}$/; // regex to search for a value with ONLY ONE decimal place
+      if (value.match(regexp)) {
+        value += '.00'
+      } else if (value.match(regexp1)){
+        value += '0'
+      }
       elements.push(
         <span className="data-item" key={key}>
           <strong>
@@ -32,7 +40,7 @@ export default class ElementalComposition extends React.Component {
         </span>
       );
     });
-
+    
     return elements;
   }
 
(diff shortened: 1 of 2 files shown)


JanCBrammer (commented):
@dp19982, could you check if the numeric precision can be specified where it originates? Might be a more sustainable / permanent solution than cosmetically changing it in the UI. Maybe in https://github.com/ComPlat/chemotion_ELN/blob/b27f9c6434364a12fe7b302911894ae07f23f480/app/models/elemental_composition.rb#L18?

dp19982:
> @dp19982, could you check if the numeric precision can be specified where it originates? Might be a more sustainable / permanent solution than cosmetically changing it in the UI. Maybe in
> 
> https://github.com/ComPlat/chemotion_ELN/blob/b27f9c6434364a12fe7b302911894ae07f23f480/app/models/elemental_composition.rb#L18
> 
> ?

Hi @JanCBrammer, the conversion to percentage values is shown on 
https://github.com/ComPlat/chemotion_ELN/blob/b27f9c6434364a12fe7b302911894ae07f23f480/lib/chemotion/calculations.rb#L152
This leads to: https://github.com/ComPlat/chemotion_ELN/blob/b27f9c6434364a12fe7b302911894ae07f23f480/lib/chemotion/calculations.rb#L196

Using the example of 2,2-bipyridine - 
Removing rounding:
Screenshot 2022-06-14 095433

rounding to 2 decimal places:
Screenshot 2022-06-14 095732

The numerical precision is specified within the calculations, but the issue of not having the second decimal place still persists. This can be changed in the back end calculations by altering https://github.com/ComPlat/chemotion_ELN/blob/development-5/lib/chemotion/calculations.rb#L196 to: `elements[key] = sprintf('%.2f',(value.to_d * 100.0))`, which formats the value as a string, and returns it as a float, providing the correct functionality in a single line of code. 😄

2,2-bipyridine - Screenshot 2022-06-14 100520
C10H18MnO6 - Screenshot 2022-06-14 100837

I hope this is sufficient?

JanCBrammer:
Looks good to me. 

> providing the correct functionality in a single line of code. 😄
Very satisfying 😄

dp19982:
> Hi
> 
> could be
> 
> ```
> 
> var x=76.901002356708;
> var formatedX=x.toFixed(2);
> ```
> 
> an alternative?

Changing https://github.com/ComPlat/chemotion_ELN/blob/b27f9c6434364a12fe7b302911894ae07f23f480/app/packs/src/components/ElementalComposition.js#L29 to `{Number(value).toFixed(2)}` would also work, yes. I had not considered that. 😅 

I personally prefer the back end solution, as that alters how the values are formatted and the front end just renders them.

JanCBrammer:
Looks like the formatting change broke some tests?

dp19982:
> Looks like the formatting change broke some tests?

I wonder if the front end implementation would also break some tests? I'll try and have a look before I push anything new.

dp19982:
@JanCBrammer I pushed the front end implementation before your comment, and all the tests have passed. To avoid unneccessary back and forth commits, I'm happy with the front end implementation. 😄

JanCBrammer (approved):
@PiTrem, can you merge?

JanCBrammer:
Turns out, _I_ can merge 😎 @PiTrem, for future reference (and the review guidelines), what do we do?

1. merge commit
2. squash & merge
3. rebase & merge

JanCBrammer:
@dp19982, @FabianMauz, for future reference: In this case we had three commits that together make up on small change. However, in the main branch we don't want those commits to show up. Instead we want a single commit that summarizes what's been done. This is why we merged this PR  using a "squash" commit, meaning we combine the three commits into one and giving it a new commit message. This seems easier than @dp19982 having to locally edit his git history via rebasing and then force pushing.

PR #741, opened 2022-06-01
add new fields for sample color and state at RT

### LINKED ISSUE #722: In sample/molecule window, add field to specify room-temperature aggregation state and color
- [ x] I am an ELNer
- [ ] I am a developer
- [ x] I am a system/ELN admin

# Description

In the sample/molecule window, currently only properties such as boiling/melting point, density, and purity can be provided.

It would be nice to also have additional fields here for (room-temperature) aggregration state (e.g. solid, liquid, oil, ...) and color.

In the 20.05.2022 Chemotion "Stammtisch" Kevin Jablonka from EPFL also showed a nice system how to assign colors more systematically based on picture and match to RGB color table

# Image

dp19982:
CHANGELOG: Closes #722

2 new free text fields have been implemented in the sample properties tab - allowing the user to specify both color and state at room temperature:

color_state_fields

### DIFF
--- a/app/packs/src/components/models/Sample.js
+++ b/app/packs/src/components/models/Sample.js
@@ -191,7 +191,7 @@ export default class Sample extends Element {
       sum_formula: '',
       xref: {},
       color: '',
-      state: ''
+      aggregateState: ''
     });
 
     sample.short_label = Sample.buildNewShortLabel();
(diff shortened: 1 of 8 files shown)


app/packs/src/components/models/Sample.js
```diff
@@ -319,11 +321,29 @@ export default class Sample extends Element {
       molecular_mass: this.molecular_mass,
       sum_formula: this.sum_formula,
       segments: this.segments.map(s => s.serialize()),
+      color: this.color,
+      state: this.state,
```
FabianMauz, line 325: May be confused with the component state of react. Better find another name (e.g. aggregateState).

Issue #760, opened 2022-06-22
Equivalent information in reagent table can get lost with one misplaced click. 


RohrbachS:
- I am an ELNer

- I am a system/ELN admin

# Expected Behavior

Information added to the reagent table should never be lost/overwritten.

# Current Behavior

It is possible to inadvertently override equivalent entries as detailed below. 

## Steps to Reproduce

_Please provide detailed steps for reproducing the issue._

1) Add two starting materials, both without density.
2) Set an amount in mmol to the limiting reagent.
3) Set the equivalents of the second reagent to 1.
4) Click in the volume field of the second reagent.
5) Click in the mass field of the second reagent. 
Result: The equivalent value for the second reagent is reset to 0. 

## Context

_Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions._

 - OS: Windows 10
 - Chrome Browser 
 
 - ELN version 1.2.0

PiTrem:
resolved with https://github.com/ComPlat/chemotion_ELN/pull/788

PR #788, opened 2022-07-08
Issue on input of equivalent value 

mehreenmansur:
Equivalent information in reagent table can get lost with one misplaced click.
[760](https://github.com/ComPlat/chemotion_ELN/issues/760)


app/packs/src/components/Material.js
```diff
@@ -323,7 +323,8 @@ class Material extends Component {
     }
   }
 
-  handleAmountUnitChange(e) {
+  handleAmountUnitChange(e, value) {
+    if (e.value == value) return;
```
baolanlequang, line 327: Why `==` instead of `===`?

baolanlequang (commented):
It needs to change the variables name for clearer meaning

baolanlequang (approved):

PiTrem (approved):

PR #803, opened 2022-07-15
inventory feature

adambasha0:

- [x] added code is linted

- [x] in case the changes are visible to the end-user,  video or screenshots should be added to the PR => helps  with user testing


app/packs/src/components/InventoryTab.js
```diff
@@ -0,0 +1,926 @@
...
+        if (newinventory) {
+          this.setState({ inventory });
+        }
+      }).catch((errorMessage) => {
+        console.log(errorMessage);
+      });
+    }
+  }
+
+  querySds = () => {
```
jhmegorei, line 86: Question: Why do you use the anonymous function syntax here instead of normal method definition?

app/packs/src/components/models/Inventory.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  serialize() {
+    return super.serialize({
+      inventory_parameters: this.inventory_parameters
+    })
+  }
+
+
+  invenParameters(parameter, value) {
```
jhmegorei, line 28: Question: What does this method do? I have seen some invocations but I do not understand its use.

From what I can read it is a convenience writer for the first entry in inventory_parameters. 
I have not read the whole InventoryTab component through so the answer may be hidden there, but I will ask anyway: What is the point of defining inventory_parameters as an array of objects if you only ever use the first object?

app/packs/src/components/InventoryTab.js
```diff
@@ -0,0 +1,926 @@
...
+    };
+    if (inventory.isNew) {
+      InventoryFetcher.create(params).then((newinventory) => {
+        if (newinventory) {
+          this.setState({ inventory });
+        }
+      }).catch((errorMessage) => {
+        console.log(errorMessage);
+      });
+      inventory.isNew = false;
```
jhmegorei, line 74: Question: Why is this line necessary? From what I can see, inventory is a local variable fetched from the component's state. In line 67 you overwrite the state with the return value of the fetcher, so the old inventory object referenced by this variable is obsolete and no longer used in the context of either the state or this method. So therefore I think this line is redundant.

app/packs/src/components/InventoryTab.js
```diff
@@ -0,0 +1,926 @@
...
+    const inventoryParameters = inventory._inventory_parameters;
+    const params = { 
+      inventory_parameters: inventoryParameters,
+      inventoriable_id: element.id, 
+      inventoriable_type: element.type 
+    };
+    if (inventory.isNew) {
+      InventoryFetcher.create(params).then((newinventory) => {
+        if (newinventory) {
+          this.setState({ inventory });
```
jhmegorei, line 69: Warning: InventoryFetcher.create does **NOT** return an instance of the Javascript Inventory model, but only the json from the Backend. See my question in the Fetcher.

app/packs/src/components/fetchers/InventoryFetcher.js
```diff
@@ -0,0 +1,110 @@
...
+    return fetch('/api/v1/inventories/create', {
+      credentials: 'same-origin',
+      method: 'post',
+      headers: {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify(params)
+    }).then(response => response.json())
+      .then(json => json)
```
jhmegorei, line 43: Question: Why does this method not return an instance of the Javascript Inventory Model? (like fetchByInventoriableId does)

app/packs/src/models/Element.js
```diff
@@ -44,6 +44,10 @@ export default class Element {
     return this.is_new == true
   }
 
+  set isNew(boolean) {
```
jhmegorei, line 47: See my question on InventoryTab line 74. I think the assignment there is redundant and as such the definition of the setter here is redundant as well

app/packs/src/components/InventoryTab.js
```diff
@@ -0,0 +1,926 @@
...
+    this.state = {
+      inventory: undefined,
+      safetySheets: [],
+      checkSaveIconThermofischer: true,
+      checkSaveIconMerck: true,
+      vendorValue: 'All',
+      vendorSafetyPhrasesValue: '',
+      vendorChemPropertiesValue: '',
+      queryOption: 'Common Name',
+      languageOfSdd: 'en',
```
jhmegorei, line 22: Question: what is SDD? Do you mean SDS, aka SafetyDataSheet?

lib/chemotion/inventory_service.rb
```diff
@@ -0,0 +1,212 @@
+# frozen_string_literal: true
+
+module Chemotion::InventoryService
```
jhmegorei, line 3: Please refactor this into a proper class, so you can reduce redundant code (like for example always the same http options) into private methods

PR #857, opened 2022-08-03
Introduce an optimized structure for javascript files

matt-hh:
This PR optimizes the directory structure of js files by:

- **Change all relative import paths to absolute (based on src)**
  This allows updating file imports by global search and replace, instead of manually finding all occurrences

- **Move apps aka entrypoints into separate folders**
  This cleans up the file structure under app/packs/src/components, so we can gradually put files that are only required in one app into this app's subfolder, essentially turning the src/components folder into a collection of shared components.

- **New place for models, fetchers, utilities, stores**
  Create a file structure for non-component code, leaving only React Components under app/packs/src/components

- **Cleanup unused imports**
   Because they are not in use 🤡


PiTrem (approved):
Click Marathon

PR #864, opened 2022-08-09
remove unnecessary json parse

Allenskywalker92:
fix failed test case, using id from json result

skip tests that relate with level details

remove present by entity

- [ ] added code is linted

### DIFF
--- a/app/api/chemotion/molecule_api.rb
+++ b/app/api/chemotion/molecule_api.rb
@@ -188,8 +188,9 @@ module Chemotion
             end
           end
         end
-        
         molecule&.attributes&.merge(temp_svg: svg_process[:svg_file_name], ob_log: ob)
+
+        Entities::MoleculeEntity.represent(molecule, temp_svg: svg_process[:svg_file_name], ob_log: ob)
       end
 
       desc 'return CAS of the molecule'
--- a/app/api/chemotion/molecule_api.rb
+++ b/app/api/chemotion/molecule_api.rb
@@ -188,9 +188,8 @@ module Chemotion
             end
           end
         end
+        
         molecule&.attributes&.merge(temp_svg: svg_process[:svg_file_name], ob_log: ob)
-
-        present molecule, with: Entities::MoleculeEntity
       end
 
       desc 'return CAS of the molecule'
--- a/spec/features/copy_reaction_spec.rb
+++ b/spec/features/copy_reaction_spec.rb
@@ -21,7 +21,7 @@ describe 'Copy reaction' do
   let(:product3) { create(:sample, name: 'Product3', real_amount_value: 4.671, molecule: m2) }
   let(:reaction3) { create(:reaction, status: 'Successful', short_label: 'Reaction3') }
 
-  let!(:col) { create(:collection, user_id: user1.id, label: 'Col3') }
+  let!(:col) { create(:collection, user_id: user1.id, label: 'Col3', permission_level: 10, reaction_detail_level: 10) }
 
   let!(:root_share) { create(:collection, user: user1, shared_by_id: user2.id, is_shared: true, is_locked: true) }
   let!(:cshare) { create(:collection, user: user1, label: 'share-col', permission_level: 10, sample_detail_level: 10, reaction_detail_level: 10, shared_by_id: user2.id, is_shared: true, ancestry: root_share.id.to_s) }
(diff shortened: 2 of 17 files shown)


app/api/chemotion/molecule_api.rb
```diff
@@ -188,9 +188,8 @@ class MoleculeAPI < Grape::API
             end
           end
         end
+        
         molecule&.attributes&.merge(temp_svg: svg_process[:svg_file_name], ob_log: ob)
-
-        present molecule, with: Entities::MoleculeEntity
```
jhmegorei, line 193: Why did you remove this line? The change looks rather random... The API should present their objects using Grape Entities. Otherwise you get pretty much undefined results, missing relations defined in MoleculeIdentity etc.

spec/features/copy_reaction_spec.rb
```diff
@@ -81,7 +80,7 @@ def copy_reaction(source_collection, target_collection)
     expect(page).not_to have_button('copy-element-btn', wait: 5)
   end
 
-  it 'to same collection', js: true do
+  xit 'to same collection', js: true do
```
phuang26, line 83: why?

app/packs/src/components/fetchers/SamplesFetcher.js
```diff
@@ -82,8 +82,8 @@ export default class SamplesFetcher {
       },
       body: JSON.stringify(sample.serialize())
     }).then(response => response.json())
-      .then(json => GenericElsFetcher.uploadGenericFiles(sample, json.sample.id, 'Sample')
-        .then(() => this.fetchById(json.sample.id))).catch((errorMessage) => {
+      .then(json => GenericElsFetcher.uploadGenericFiles(json, json.id, 'Sample')
```
phuang26, line 85: why?

app/api/chemotion/molecule_api.rb
jhmegorei: I see. But then the correct move would be to add these additional attributes to the entity (using the context hash allows this, see ApplicationEntity#displayed_in_list? for how to implement it), not removing the entity entirely.

Returning the attributes hash directly is undefined behaviour as new attributes are added or old ones deleted over time. Using the entity, the returned data structure can be inferred directly without looking up the current data model of the molecule table.

app/packs/src/components/fetchers/SamplesFetcher.js
Allenskywalker92: @phuang26 I will revert it.

app/api/chemotion/molecule_api.rb
Allenskywalker92: I updated it like you said. Please help me to take a loook.

jhmegorei: This should work fine :) The options hash access evaluates to nil so any other place using the MoleculeEntity will always get nil. While it is possible to only render the temp_svg and ob_log (what is an ob_log anyway? could we refactor this variable here and in all other related places to have a meaningful name?) only when not nil, I don't consider that strictly necessary.

Allenskywalker92: While it is possible to only render the temp_svg and ob_log - Could you show me how to achieve it?
What is an ob_log anyway? could we refactor this variable here and in all other related places to have a meaningful name? - Actually I don't know either, I fix here for it to run correctly at first, we can improve it later.

jhmegorei: The MoleculeEntity actually contains an example of how it works :) see the separate exposure of  :molecule_names ? the :unless parameter receives a Proc with the options as parameter, so you could write it like this:

`expose :temp_svg, unless: ->(instance, options) { options[:temp_svg].present? }`

spec/features/copy_reaction_spec.rb
Allenskywalker92: updated

app/api/chemotion/molecule_api.rb
Allenskywalker92: updated

jhmegorei: 👍

PiTrem (changes requested):
check report feature is working

Allenskywalker92:
> check report feature is working

@PiTrem The report only works after my changes.

PiTrem (approved):

PR #877, opened 2022-08-12
metadata-converter-dataset

phuang26:


cllde8 (approved):

app/packs/src/admin/generic/Utils.js
```diff
@@ -36,7 +36,7 @@ const inputEventVal = (event, type) => {
 };
 
 const absOlsTermId = val => (val || '').split('|')[0].trim();
-const absOlsTermLabel = val => val.replace(absOlsTermId(val), '').replace('|', '').trim();
+const absOlsTermLabel = val => val?.replace(absOlsTermId(val), '')?.replace('|', '').trim();
```
jhmegorei, line 39: Instead of the ?. operator you could supply a default value just like in the line above :)

lib/analyses/converter.rb
```diff
@@ -5,6 +5,10 @@
 
 module Analyses
   class Converter
+    def self.logger
```
jhmegorei, line 8: Can you add some kind of spec for this file? It only consists of public class methods and those should be tested.

If time permits, I recommend refactoring this file into instance based code, so only the intended interface is public and all other methods are private. This way a reader can instantly see which methods are meant to be called from the outside.

Another good thing would be reworking the variable and method naming. Especially in the larger codeblocks you added, you use a lot of very short abbreviations that don't make sense to anyone but you. This harms readability a lot.

PiTrem:
covered by 193d90d065546258e29e0e706f40cef1e46398ed 
https://github.com/ComPlat/chemotion_ELN/pull/1149

PR #878, opened 2022-08-15
Ci image update

PiTrem:
update CI image:

- node 14.20

- ruby 2.6.8 (default) 2.7.5 (use .tool-versions)

- built  from 98160166c 

also removed checkboxes from issue templates

### DIFF
--- a/app/packs/src/models/Attachment.js
+++ b/app/packs/src/models/Attachment.js
@@ -1,9 +1,9 @@
 import Element from 'src/models/Element';
 
 export default class Attachment extends Element {
-  filePreview(file) {
+  static filePreview(file) {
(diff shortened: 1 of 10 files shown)


app/packs/src/models/Attachment.js
```diff
@@ -1,9 +1,9 @@
 import Element from 'src/models/Element';
 
 export default class Attachment extends Element {
-  filePreview(file) {
+  static filePreview(file) {
```
jhmegorei, line 4: Why is this still displayed as a change? development-5 should already have this change

PiTrem: I think github uses ...

`git diff development-5...ci-image-update`

PR #898, opened 2022-08-29
Refactored `ReactionHelpers#update_materials_for_reaction`.

JanCBrammer:
Refactoring for readability and in preparation of further developments.


app/api/chemotion/reaction_api.rb
```diff
@@ -45,6 +45,121 @@ def rangebound(lower, upper)
     end
   end
 
+  def create_sub_sample(sample, current_user, reaction, fixed_label)
```
jhmegorei, line 48: Can you remove this code from the API and move this to a usecase object.
The api should be as lean as possible (the Grape Format gets even more unreadable with large chunks of code).

JanCBrammer: @jhmegorei, to clarify, moving this to a usecase object means moving `update_materials_for_reaction` to `usecases/reactions/update_materials`?

```Ruby
module Usecases
  module Reactions
    class UpdateMaterials

    end
  end
end
```

jhmegorei: Hmm... I would recommend naming the class Usecases::Reactions::Update if you plan on naming the method `update_materials_for_reaction`, as otherwise the term `material` occurs both in the class name and the method.

While I would love to see the whole update process extracted into the usecase, it does not appear feasible at the moment, as there are more helpers involved (like ContainerHelper's `update_datamodel`), this is at least a step into the right direction.

JanCBrammer: > While I would love to see the whole update process extracted into the usecase, it does not appear feasible at the moment, as there are more helpers involved (like ContainerHelper's `update_datamodel`), this is at least a step into the right direction.

@jhmegorei, agreed. I included the ContainerHelpers in the usecase for now (see latest commits).

app/usecases/reactions/update_materials.rb
```diff
@@ -0,0 +1,213 @@
...
+
+  def to_boolean(string)
+    !!"#{string}".match(/^(true|t|yes|y|1)$/i)
+  end
+end
+
+module Usecases
+  module Reactions
+    class UpdateMaterials
+      include ContainerHelpers
```
JanCBrammer, line 39: @jhmegorei, using `include` makes it so that `ContainerHelplers` isn't treated as a Grape helper anymore like previously : https://github.com/ComPlat/chemotion_ELN/blob/e412ff7f8c61f2c347458cfc9b0ca6d2f2ffb99b/app/api/chemotion/reaction_api.rb#L220

Therefore `update_datamodel` looses access to the Grape namespace / scope, which causes `./spec/features/sync_collection_spec.rb:56` to break:

```
     NameError:
       undefined local variable or method `current_user' for #<Usecases::Reactions::UpdateMaterials:0x000055acd0812068>
     # ./app/api/helpers/container_helpers.rb:117:in `can_update_container'
     # ./app/api/helpers/container_helpers.rb:27:in `create_or_update_containers'
     # ./app/api/helpers/container_helpers.rb:18:in `update_datamodel'
     # ./app/usecases/reactions/update_materials.rb:160:in `update_existing_sample'
     # ./app/usecases/reactions/update_materials.rb:69:in `block (3 levels) in execute!'
     # ./app/usecases/reactions/update_materials.rb:59:in `each'
     # ./app/usecases/reactions/update_materials.rb:59:in `each_with_index'
     # ./app/usecases/reactions/update_materials.rb:59:in `block (2 levels) in execute!'
     # ./app/usecases/reactions/update_materials.rb:56:in `each'
     # ./app/usecases/reactions/update_materials.rb:56:in `block in execute!'
     # ./app/usecases/reactions/update_materials.rb:54:in `execute!'
     # ./app/api/chemotion/reaction_api.rb:173:in `block (3 levels) in <class:ReactionAPI>'
```

Ideas on how to fix this?

jhmegorei: ContainerHelpers is just a ruby module, so writing `include ContainerHelpers` should work (constant resolution issues aside). Thing is that ContainerHelpers extends Grape::Api::Helpers, which in turn does some stuff that might interfere with your object if not careful.

The error message was caused by `current_user` not being  available in your usecase class. This can easily be rectified by adding the current_user as an attr_reader to your usecase class and passing it in the initializer

```
class Usecases::Reactions::UpdateMaterials
  attr_reader :current_user

  def initialize(someStuff, current_user:)
    @current_user = current_user
  end
end
```

JanCBrammer: > This can easily be rectified by adding the current_user as an attr_reader to your usecase class

Thanks for the pointer! All tests (Ruby unit & acceptance) pass now locally.

spec/usecases/reactions/update_materials_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+        'equivalent' => 1,
+        'reference' => true,
+        'is_new' => true,
+        'molfile' => molfile,
+        'container' => root_container
+      ]
+    }.merge(starting_materials, reactants, products)
+  end
+
+  it 'updates existing sample' do
```
jhmegorei, line 78: Please insert a middle layer between the `describe Usecases::Reactions::UpdateMaterials` and the `it` blocks.

Usually the recommended layout of an Rspec file for a class looks like this:
```
describe Usecases::Reactions::UpdateMaterials do
   describe '#execute!' do
      context 'when SOME_CONDITION' do
         it 'does something' do
             ...
         end
     end
  end
end
```

With this structure you can navigate your file easier and some Editor/IDE tooling may even allow you to jump from method definition to the corresponding test and vice versa.

spec/usecases/reactions/update_materials_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+
+  it 'creates sample for new materials that are products' do
+    described_class.new(reaction, products, user).execute!
+    expect(Sample.all.size).to eq(1) # ReactionHelpers create new sample
+    expect(Sample.find_by(name: 'product').target_amount_value).to eq(99.08304)
+  end
+
+  it 'associates reaction with materials' do
+    described_class.new(reaction, materials, user).execute!
+    expect(ReactionsSample.all.size).to eq(4)
```
jhmegorei, line 98: You can use `ReactionsSample.count` to just to a count query without instantiating all fetched objects

FabianMauz (approved):

PR #904, opened 2022-09-07
Prevent sample-model from unnecessarily updating reaction-SVG

JanCBrammer:
Calling `save` on a sample triggers an `after_save` callback in the sample-model that updates the reaction-SVGs that are associated with that sample.

https://github.com/ComPlat/chemotion_ELN/blob/dd3e2c4852ed953b4a89a36dcc1593094e64397a/app/models/sample.rb#L167
https://github.com/ComPlat/chemotion_ELN/blob/dd3e2c4852ed953b4a89a36dcc1593094e64397a/app/models/sample.rb#L511

When a sample is saved in the context of the reaction-API it is not necessary for the sample-model to update the reaction-SVG, since the reaction-model itself updates the reaction-svg.

https://github.com/ComPlat/chemotion_ELN/blob/dd3e2c4852ed953b4a89a36dcc1593094e64397a/app/usecases/reactions/update_materials.rb#L82
https://github.com/ComPlat/chemotion_ELN/blob/dd3e2c4852ed953b4a89a36dcc1593094e64397a/app/models/reaction.rb#L144

The performance of the reaction-API suffers from this, since calling the reaction-API's `post` and `put` endpoints causes the sample-model to carry out unnecessary updates of the reaction-SVG. To improve performance, I prevent the sample-model from updating the reaction-SVG when `sample.save` is called from the reaction-API.

I submit this optimization in preparation of a feature that requires creating multiple reactions in quick succession.
All tests pass locally.

### DIFF
--- a/spec/usecases/reactions/update_materials_spec.rb
+++ b/spec/usecases/reactions/update_materials_spec.rb
@@ -13,7 +13,6 @@ describe Usecases::Reactions::UpdateMaterials do
   let(:starting_materials) do
     {
       'starting_materials' => [
-        # hits existing_sample branch
         'id' => sample1.id,
         'name' => 'starting_material',
         'target_amount_unit' => 'mg',
@@ -29,7 +28,6 @@ describe Usecases::Reactions::UpdateMaterials do
   let(:reactants) do
     {
       'reactants' => [
-        # hits subsample branch
         'target_amount_unit' => 'mg',
         'target_amount_value' => 86.09596,
         'equivalent' => 2,
@@ -44,7 +42,6 @@ describe Usecases::Reactions::UpdateMaterials do
   let(:products) do
     {
       'products' => [
-        # hits new_sample branch
         'name' => 'product',
         'target_amount_unit' => 'mg',
         'target_amount_value' => 99.08304,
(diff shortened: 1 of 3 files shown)


spec/usecases/reactions/update_materials_spec.rb
```diff
@@ -73,23 +73,36 @@
...
+        ReactionsStartingMaterialSample.create!(
+          reaction: reaction, sample: sample, reference: true, equivalent: 1
+        )
+      end
+
+      it 'does not update reaction-SVG from sample model' do
+        # Capturing SVG::ReactionComposer:new is the most straightforward way I could think of to test updates to the reaction-SVG.
+        # Since the updates to the reaction-SVG are a side-effect they are difficult to test.
+        expect(SVG::ReactionComposer).to receive(:new).once # only one final update to reaction-SVG once reaction is saved
+        described_class.new(reaction, starting_materials, user).execute!
```
matt-hh, line 86: All other it-blocks are built with a "given..when..then" structure (aka. "setup..call the method..expect something") except this one ("given..than..when"). It-Blocks are easier to refactor if they follow the same pattern.

```suggestion
        allow(SVG::ReactionComposer).to receive(:new) # this enables the have_received-matcher
        described_class.new(reaction, starting_materials, user).execute!
        expect(SVG::ReactionComposer).to have_received(:new)
```

Let's refactor this spec a little bit in a later step...

matt-hh (changes requested):

matt-hh (approved):
LGTM!

JanCBrammer:
@PiTrem, can you merge? @FabianMauz and @matt-hh don't have write access (and hence cannot approve merging) and I cannot approve my own PR it appears :grimacing:

PR #906, opened 2022-09-08
Improve Attachment API and Attachment Specs

jhmegorei:
This PR improves the Attachment API and adds a lot of specs for the attachment model and its included concerns.
Not everything is specced though, due to time constraints. I urge other developers to add to the specs, so that especially the AttachmentJcampAasm Concern gets completely tested (it uses the also untested Jcamp Interface)

### WARNING
This PR removes the following methods from the Attachment model:
- for_report?
- for_template?
- report_id
- report
- update_report!


app/api/chemotion/attachable_api.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module Chemotion
+  class AttachableAPI < Grape::API
+    resource :attachable do
```
FabianMauz, line 5: Where is this resource used?

app/usecases/attachments/upload_chunk_complete.rb
```diff
@@ -0,0 +1,58 @@
...
+          filename: file_name,
+          key: params[:key],
+          file_path: file_path,
+          created_by: user.id,
+          created_for: user.id,
+          content_type: MIME::Types.type_for(file_name)[0].to_s
+        )
+        attachment.save!
+
+        true
```
FabianMauz, line 54: Why always true, even if an attachment could not be saved?

JanCBrammer (approved):

spec/models/attachment_spec.rb
```diff
@@ -3,102 +3,833 @@
...
+        expect(subject).to eq(true)
+      end
+    end
+  end
+
+  describe '#for_container?' do
+    subject { attachment.for_container? }
+
+    context 'when not attached to container' do
+      let(:attachment) { create(:attachment, :attached_to_research_plan) }
```
phuang26, line 110: typo? attached_to_research_plan?

jhmegorei: no this is exactly the point. The context describes an attachment NOT being attached to a container, so I created on attached to a research_plan (which is not a container)

app/usecases/attachments/unlink.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+
+      def self.execute!(attachment)
+        new(attachment).execute!
+      end
+
+      def execute!
+        attachment.tap do |a|
+          a.attachable_id = nil
+          a.attachable_type = 'Container'
```
phuang26, line 19: why?

jhmegorei: why what? can you be more specific please?

phuang26: why the attachable_type set to 'Container' when unlink?

jhmegorei: That is what we wondered as well... please see https://github.com/ComPlat/chemotion_ELN/commit/e04c83b047af65f528e965023341ba04b3a070cd

phuang26 (commented):
the attachable_spec.rb were removed and they were introduced for "Report LSDF - Spec".

phuang26 (approved):

PR #912, opened 2022-09-13
Ketcher service

ptrxyz:
Stuff to render molfiles on a [ketcher backend service](https://github.com/ptrxyz/chemotion-ketchersvc) as here.

### DIFF
--- a/app/models/molecule.rb
+++ b/app/models/molecule.rb
@@ -95,7 +95,7 @@ class Molecule < ApplicationRecord
       molecule.molfile = is_partial && partial_molfile || molfile
       molecule.assign_molecule_data(babel_info, pubchem_info)
     end
- 
+
     molecule.ob_log = babel_info[:ob_log]
     molecule
   end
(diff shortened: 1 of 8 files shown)


app/models/molecule.rb
```diff
@@ -246,6 +245,20 @@ def unique_molecule_name(new_name)
...
+    return svg unless Rails.configuration.try(:ketcher_service).try(:url).present?
+    return svg if svg.present? && !svg&.include?('Open Babel')
+
+    svg = KetcherService::RenderSvg.svg(molfile)
+
+    if svg&.present?
+      svg = Ketcherails::SVGProcessor.new(svg)
+      svg.centered_and_scaled_svg
+    else
+      Chemotion::OpenBabelService.svg_from_molfile(molfile)
```
PiTrem, line 258: line unreachable if any error is raised by KetcherService 
eg: when the service is configured but down, then it is not possible to create sample with smiles
(see comment below)

lib/ketcher_service/render_svg.rb
```diff
@@ -0,0 +1,42 @@
...
+      raise
+    end
+
+    def self.svg(molfile)
+      url = URI(Rails.configuration.ketcher_service.url)
+      request = Net::HTTP::Post.new(url.path, {'Content-Type' => 'application/json'})
+      request.body = { molfile: molfile.force_encoding('utf-8') }.to_json
+      svg = RenderSvg.call_render_service(url, request)
+      svg.force_encoding('utf-8')
+    end
```
PiTrem, line 40: we could `rescue` here to return nil in any cases, so that Molecule.svg_reprocess can always fall back to openbabel?
(see comment above)

PR #924, opened 2022-09-16
Applying gem shrine

PiTrem:


app/models/concerns/attachment_jcamp_aasm.rb
```diff
@@ -127,19 +127,25 @@ def generate_att(meta_tmp, addon, to_edit = false, ext = nil)
...
+      )
+    end
+    att.attachment_attacher.attach(File.open(meta_tmp.path, binmode: true))
+    att.attachment_attacher.create_derivatives
     att.save!
     att.set_edited if ext != 'png' && to_edit
     att.set_image if ext == 'png'
     att.set_json if ext == 'json'
+
+    att.update!(attachable_id: attachable_id, attachable_type: 'Container')
```
jhmegorei, line 148: Why do you change the attachable type to Container here? An attachment can also be related to other classes like ResearchPlan. Doing this would destroy existing relations

app/packs/src/apps/mydb/elements/details/researchPlans/researchPlanTab/ImageAttachmentFilter.js
```diff
@@ -0,0 +1,11 @@
+export default class ImageAttachmentFilter {
```
jhmegorei, line 1: This is a utility class and should be placed in /packs/src/utilities I think.
The /packs/src/app/ and /packs/src/components/ folders should only contain react components if possible

PiTrem:
re-opened there https://github.com/ComPlat/chemotion_ELN/pull/997

app/models/concerns/attachment_jcamp_aasm.rb
PiTrem: probably something :fish:  but as of now the spectra feature is expected only for  attachments  a dataset.

Issue #663, opened 2022-04-13
CI: remove duplication in GH Actions workflows with reusable workflows


JanCBrammer:
https://github.blog/2021-11-29-github-actions-reusable-workflows-is-generally-available/

PR #952, opened 2022-10-05
Refactored CI

JanCBrammer:
This PR introduces the following changes to our continuous integration (CI):

* Removed duplication from CI in order to save ourselves from [shotgun surgery](https://en.wikipedia.org/wiki/Shotgun_surgery), and the [rain forest from dying](https://en.wikipedia.org/wiki/Green_computing). By duplication, I mean that the workflow definitions `testacceptance.yml`, `testruby.yml`, and `testjs.yml` are identical in large parts. Those three workflow definitions are now combined in `CI.yml`.
* Don't run CI on pushes to feature branches; run CI only on pushes to main and pull requests (because rain forest).
* Added Docker image for running CI on GitHub CI; rather than working around problems that stemmed from trying to make Git**Hub** CI work with the Docker image that's intended for Git**Lab** CI (https://github.com/ComPlat/chemotion_ELN/blob/development-5/Dockerfile.focal.gitlab-ci).

Part of our effort to improve CI: #663.


JanCBrammer:
Discussed the following changes with @PiTrem and @matt-hh:

in the following, "Docker image" refers to the image that's used to run the CI ([Docke](https://github.com/ComPlat/chemotion_ELN/blob/development-5/Dockerfile.focal.github-ci)

- [ ] 1. instead of caching the gems and npm packages, build gems and npm packages in Docker image and update them on the GitHub runner (i.e., revert back to how we're handling the gems and npm packages currently). We favor this solution because it reduces the build time for the dependencies more than caching, since the dependencies potentially need to be re-cached frequently.

- [ ] 2. to facilitate 1., build the Docker image with a Ruby base image (in order to avoid having to install the Ruby environment in the image manually). Ideally, the base image includes node and yarn as well.

- [ ] 3. build the Docker image automatically with GitHub Actions to ensure up-to-date dependencies. To clarify: Which events trigger building the image (e.g., pushed including yarn and/or gem lockfiles), and from which branch?

PiTrem (approved):

PR #976, opened 2022-10-18
Puma to production

ptrxyz:
Moved puma gem to production group since we use it in containers.


jhmegorei:
Can you please check the base branch or the target branch? This seems to be branched off Release 1.4.0, which is miles away from development-5 and needs to be merged first and then manually integrated into development-5 due to the massive amount of changes.

Additionally please set this PR to draft, so no one merges it by accident.

PiTrem:
solved with 3af61a8e41

PR #977, opened 2022-10-18
renamed app.root_url in delayed jobs

ptrxyz:
This patch fixes a bug, where the delayed_job was referring to the wrong property.
This caused the delayed_job to fail, when the user requested to download all analysis files
or to recreate `.edit.jdx` files.


jhmegorei:
Can you please check the base branch or the target branch? This seems to be branched off Release 1.4.0, which is miles away from development-5 and needs to be merged first and then manually integrated into development-5 due to the massive amount of changes.

Additionally please set this PR to draft, so no one merges it by accident.

PiTrem:
replaced with https://github.com/ComPlat/chemotion_ELN/pull/1042

PR #990, opened 2022-10-21
fix cannot open chemspectra in research plan and reaction

baolanlequang:

- [x] added code is linted

- [x] in case the changes are visible to the end-user,  video or screenshots should be added to the PR => helps  with user testing

- [x] CHANGELOG :  add a bullet point on top (optional: reference to github issue/PR )

- [x] parallele PR for documentation  on docusaurus  if the feature/fix is tagged for a release

### DIFF
--- a/app/packs/src/models/Element.js
+++ b/app/packs/src/models/Element.js
@@ -87,6 +87,38 @@ export default class Element {
     return cleanParams;
   }
 
+  // get analyses container if any
+  analysesContainers() {
+    if (this.container.children.length === 0) {
+      const analyses = Container.buildEmpty();
+      analyses.container_type = 'analyses';
+      this.container.children.push(analyses);
+    }
+    return this.container
+      .children
+      .filter((el) => ~el.container_type.indexOf('analyses'));
+  }
+
+  analysisContainers() {
+    let target = [];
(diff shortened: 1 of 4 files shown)


jhmegorei:
Can you please update the ViewSpectra component by renaming the "sample" prop to something that makes it more clear that also ResearchPlans or Reactions can be passed?

Currently the property is named sample but the ResearchPlanDetailsContainers.js and ReactionDetailsContainers.js pass something that is not a sample, which is confusing.

jhmegorei (changes requested):
Can you please update the ViewSpectra component by renaming the "sample" prop to something that makes it more clear that also ResearchPlans or Reactions can be passed?

Currently the property is named sample but the ResearchPlanDetailsContainers.js and ReactionDetailsContainers.js pass something that is not a sample, which is confusing.

baolanlequang:
For now, we should keep it first because this pull request needs to be merged for the next release. It is a critical bug.

For the changing name, I think you can open another issue for that.

PiTrem (changes requested):
DRY

identical routines , that also appears in ../Sample.js if I am not mistaken

PiTrem (approved):

PR #998, opened 2022-10-25
Full unit test coverage of `ReactionAPI`

JanCBrammer:
Increased unit test coverage of reaction API in preparation of feature development.


matt-hh (changes requested):
Hi @JanCBrammer 

I suggest that you remove the outer context `context 'authorized user logged in' do` and use `include_context 'api request authorization context'` instead. See https://github.com/ComPlat/chemotion_ELN/blob/development-5/spec/api/chemotion/attachment_api_spec.rb#L6 It's much clearer and cleaner, the reaction_api is never accessible for non-logged in users and "RSpec/NestedGroups: Maximum example group nesting exceeded [4/3]." should disappear.

Also please move the file to the right location from spec/api/reaction_api_spec.rb to spec/api/**chemotion**/reaction_api_spec.rb

spec/api/chemotion/reaction_api_spec.rb
```diff
@@ -0,0 +1,551 @@
...
+  include_context 'api request authorization context'
+
+    let(:new_root_container) { create(:root_container) }
+
+  describe 'GET /api/v1/reactions' do
+    let!(:c1) do
+      create(:collection, label: 'C1', user: user, is_shared: false, reaction_detail_level: 10)
+    end
+    let!(:r1) { create(:reaction, name: 'r1', collections: [c1]) }
+    let!(:r2) { create(:reaction, name: 'r2', collections: [c1]) }
```
FabianMauz, line 15: Would it make sense to create a reaction in another collection so that it is checked that this is not delivered with it? Or are these already set with a seeding?

spec/api/chemotion/reaction_api_spec.rb
```diff
@@ -0,0 +1,551 @@
...
+
+      describe 'reading reaction r1' do
+        before { get "/api/v1/reactions/#{r1.id}" }
+
+        it 'is allowed' do
+          expect(response.status).to eq 200
+        end
+      end
+
+      describe 'reading reaction r2' do
```
FabianMauz, line 109: Why testing the same behavior as in line 101?

FabianMauz (approved):

FabianMauz (commented):
Please fix the rubocop comments

spec/api/chemotion/reaction_api_spec.rb
JanCBrammer: Right, seems redundant. See e71b7e2.

spec/api/chemotion/reaction_api_spec.rb
JanCBrammer: See 515aebd.

JanCBrammer:
@matt-hh, @FabianMauz, thanks for your comments! They're all addressed now. 
I fixed the linter complaints that were auto-fixable.
TBD: what to do with the non-autofixable complaints?

PiTrem (approved):

PR #1036, opened 2022-11-15
Development container for Visual Studio Code

JanCBrammer:
This PR adds a development container for Visual Studio Code (in the following referred to as "devcontainer" for brevity):
https://code.visualstudio.com/docs/devcontainers/containers.

The PR is meant to unify our existing attempts at providing a devcontainer:
* https://github.com/JanCBrammer/Chemotion_ELN_devcontainer
* https://github.com/ptrxyz/chemotion/tree/v1.0.3D0.1/client-vscode-remote-container

Those attempts are fragmented and not as well maintained as they could be. By providing the devcontainer as part of the ELN repo, I'm hoping that it'll be more accessible and salient, and therefore easier to use and maintain.

The devcontainer uses the Dockerfile and compose configuration that are already in the repository:
* https://github.com/ComPlat/chemotion_ELN/blob/1709c2edcb9ecf7dd4471b2e44558de6ef07eb3a/Dockerfile.chemotion-dev
* https://github.com/ComPlat/chemotion_ELN/blob/1709c2edcb9ecf7dd4471b2e44558de6ef07eb3a/docker-compose.dev.yml

I.e., the present changes are merely a configuration layer on top of the existing containerization.

If this proposal is accepted, I'll update our documentation about using devcontainers:
https://www.chemotion.net/chemotionsaurus/docs/eln/development_docker_installation#eln-in-docker

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,58 @@
+// https://code.visualstudio.com/docs/editor/debugging#_launchjson-attributes
+// https://code.visualstudio.com/docs/editor/variables-reference
+// https://github.com/microsoft/vscode-recipes/tree/master/debugging-Ruby-on-Rails
+// https://github.com/rubyide/vscode-ruby/blob/main/docs/debugger.md
+// Debug configurations requires..
+// the solargraph, debase, and ruby-debug-ide gems to be (bundle) installed, and
+// the castwide.solargraph, and rebornix.ruby extensions to be installed.
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Debug Rails server",
+      "type": "Ruby",
+      "request": "launch",
+      "cwd": "${workspaceRoot}",
+      "useBundler": true,
+      "pathToBundler": "${env:BUNDLER_PATH}",
+      "pathToRDebugIDE": "${env:RUBY_DEBUG_IDE_PATH}",
+      "program": "${workspaceRoot}/bin/rails",
+      "args": [
+        "server",
+        "-b",
+        "0.0.0.0",
+        "-p",
+        "3000"
+      ],
+      "env": {
+        "RAILS_ENV": "development"
+      }
+    },
+    {
+      "name": "Debug RSpec - selected line, active file",
+      "type": "Ruby",
+      "request": "launch",
+      "cwd": "${workspaceRoot}",
+      "useBundler": true,
+      "pathToBundler": "${env:BUNDLER_PATH}",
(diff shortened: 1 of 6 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,58 @@
...
+    },
+    {
+      "name": "Debug RSpec - selected line, active file",
+      "type": "Ruby",
+      "request": "launch",
+      "cwd": "${workspaceRoot}",
+      "useBundler": true,
+      "pathToBundler": "${env:BUNDLER_PATH}",
+      "pathToRDebugIDE": "${env:RUBY_DEBUG_IDE_PATH}",
+      "program": "/home/chemotion-dev/.asdf/installs/ruby/2.7.6/bin/rspec", // "program" doesn't support env variable substitution
```
jhmegorei, line 39: Shouldn't a ```bundle exec rspec``` suffice? you are already in workspace root and as long as the devcontainer does not use some strange PATH shenanigans, the regular way of invoking rspec should be fine

.vscode/tasks.json
```diff
@@ -0,0 +1,38 @@
...
+    },
+    {
+      "label": "RSpec - all",
+      "type": "shell",
+      "command": "RAILS_ENV=test $(which bundle) exec rspec"
+    },
+    {
+      "label": "RSpec - unit",
+      "type": "shell",
+      "command": "RAILS_ENV=test $(which bundle) exec rspec --exclude-pattern spec/{features}/**/*_spec.rb"
```
jhmegorei, line 20: please wrap the pattern in "", otherwise shell globs will cause rspec to miss some files that should be tested

.vscode/launch.json
JanCBrammer: Do you mean changing the value for "program" from
`"/home/chemotion-dev/.asdf/installs/ruby/2.7.6/bin/rspec"`
to
`"bundle exec rspec"`?

matt-hh: What about the useBundler option in line 36?

JanCBrammer: The `useBundler` option is documented here: https://github.com/rubyide/vscode-ruby/wiki/2.-Launching-from-VS-Code.

The `launch.json` debugging configurations are somewhat intransparent. What happens is that VSCode infers from `type: "Ruby"` that it's supposed to debug a "program" using the `ruby-debug-ide`. In case the latter gem has been bundle-installed, the configuration requires setting the `useBundler` option to `true`.

ptrxyz (approved):
Looking good to me.

PR #1056, opened 2022-11-29
dependabot html2pdf version issue

mehreenmansur:
html2pdf.js@0.9.3 requires jspdf@1.4.1

dependabot


JanCBrammer:
The upgrade of `html2pdf.js` required an upgrade of `process` which in turn required the changes in `config/webpack/custom.js` (https://github.com/react-dnd/react-dnd/issues/3425#issuecomment-1214554950)

PR #1131, opened 2023-02-15
Spec tests for FindByShortLabel endpoint and class

### LINKED ISSUE #1050: Tests for Reaction API findByShortLabel endpoint
Adding spec tests for:
- findByShortLabel endpoint in Reaction API
https://github.com/ComPlat/chemotion_ELN/pull/999/files/72204c01f0d2fd0d3ec9eaa5bf0c907cd5f4eb6f#r1005684667
- FindByShortLabel class https://github.com/ComPlat/chemotion_ELN/pull/999/files/72204c01f0d2fd0d3ec9eaa5bf0c907cd5f4eb6f#r1005685264

dpatelev:
Adds tests for FindByShortLabel as per #1050

### DIFF
--- a/app/models/reaction.rb
+++ b/app/models/reaction.rb
@@ -264,7 +264,16 @@ class Reaction < ApplicationRecord
   def auto_set_short_label
     prefix = creator.reaction_name_prefix
     counter = creator.counters['reactions'].succ
-    self.short_label = "#{creator.initials}-#{prefix}#{counter}"
+    sh_label = self['short_label']
+    return if short_label && !short_label_changed?
+
+    if sh_label && Reaction.find_by(short_label: sh_label)
+      if creator && counter
+        self.short_label = "#{creator.initials}-#{prefix}#{counter}"
+      end
+    elsif !sh_label && creator && counter
+      self.short_label = "#{creator.initials}-#{prefix}#{counter}"
+    end
   end
 
   def update_counter
--- a/app/models/reaction.rb
+++ b/app/models/reaction.rb
@@ -266,14 +266,10 @@ class Reaction < ApplicationRecord
     counter = creator.counters['reactions'].succ
     sh_label = self['short_label']
     return if short_label && !short_label_changed?
+    return if !(creator && counter)
 
-    if sh_label && Reaction.find_by(short_label: sh_label)
-      if creator && counter
-        self.short_label = "#{creator.initials}-#{prefix}#{counter}"
-      end
-    elsif !sh_label && creator && counter
-      self.short_label = "#{creator.initials}-#{prefix}#{counter}"
-    end
+    label_already_exists = Reaction.find_by(short_label: sh_label)
+    self.short_label = "#{creator.initials}-#{prefix}#{counter}" if !sh_label || label_already_exists
   end
 
   def update_counter
--- a/spec/lib/import/import_collections_spec.rb
+++ b/spec/lib/import/import_collections_spec.rb
@@ -90,7 +90,7 @@ RSpec.describe 'ImportCollection' do
         expect(reaction.temperature.to_s).to eq('{"data"=>[], "userText"=>"30", "valueUnit"=>"°C"}')
         expect(reaction.status).to eq('Done')
         expect(reaction.solvent).to eq('')
-        expect(reaction.short_label).to eq('UU-R1')
+        expect(reaction.short_label).to eq('FM-R2')
         expect(reaction.role).to eq('gp')
         expect(reaction.duration).to eq('1 Day(s)')
         expect(reaction.conditions).to eq('')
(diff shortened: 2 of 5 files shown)


endevor-music:
New commit causes more testing failures.

Screenshot 2023-02-24 100349

endevor-music:
> New commit causes more testing failures. Running tests locally.
> 
> Screenshot 2023-02-24 100349

I believe the above screenshot is a failed test that has come about by the incorrect hardcoding of a short label. @PiTrem @FabianMauz could you please look this PR over and confirm my suspicions, before I edit `import_collections_spec.rb`?

Too long; didn't read: incorrect expected value for a reaction's short label `UU-R1` in `import_collections_spec.rb`, should be `FM-R2`.

If my (brief) understanding is correct, `import_collections_spec.rb` works as follows:
1. creates a user with the abbreviation `UU`
2. imports collections from `spec/fixtures/import/{import_id}.zip`.
3. executes tests comparing values from the imported collection to what is hard-coded in the test file (which should be correct!)

In the screenshot above we can see that the reaction import test in `import_collections_spec.rb` fails, as it expects a short label of `UU-R1`, but receives `FM-R2` instead.

Before I modified `auto_set_short_label` in the reaction model, this error was not occurring, as `auto_set_short_label` statically assigns a short label, and does not check for prior existences of short labels. In the case of a user with an abbreviation of UU (as per `import_collections_spec.rb`), the first reaction's short label would be `UU-R1`. Therefore `import_collections_spec.rb` ran with no issues, as the expected short label (UU-R1), matched the *newly assigned* short label! https://github.com/ComPlat/chemotion_ELN/blob/main/spec/lib/import/import_collections_spec.rb#L93

My modifications replicated the functionality of `auto_set_short_label` in the Sample model, where it checks for prior existences of short labels, and assigns short labels only if they are not present. This would naturally break the reaction import test, as the short label returned by the test would now be a short label that exists already. Looking at `export.json` file in `spec/fixtures/import/collection_reaction.zip`, the file used in the "broken" test, the short label of the reaction is `FM-R2`, which should be the expected short label in `import_collections_spec.rb`.

spec/lib/import/import_collections_spec.rb
```diff
@@ -90,7 +90,7 @@
         expect(reaction.temperature.to_s).to eq('{"data"=>[], "userText"=>"30", "valueUnit"=>"°C"}')
         expect(reaction.status).to eq('Done')
         expect(reaction.solvent).to eq('')
-        expect(reaction.short_label).to eq('UU-R1')
+        expect(reaction.short_label).to eq('FM-R2')
```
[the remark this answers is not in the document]
endevor-music: postponed until after v1.5 release

app/models/reaction.rb
```diff
@@ -264,7 +264,12 @@ def cleanup_array_fields
   def auto_set_short_label
     prefix = creator.reaction_name_prefix
     counter = creator.counters['reactions'].succ
-    self.short_label = "#{creator.initials}-#{prefix}#{counter}"
+    sh_label = self['short_label']
+    return if short_label && !short_label_changed?
+    return if !(creator && counter)
+
+    label_already_exists = Reaction.find_by(short_label: sh_label)
```
PiTrem, line 271:
``` rb
# build and set short label
# if already exists increase counter & and run again
# if label not used , nothing to do

short_label = "#{creator.initials}-#{prefix}#{counter}"
if Reaction.find_by(short_label: short_label)
 update_counter
 autoset_short_label
end

```

app/models/reaction.rb
```diff
@@ -264,7 +264,12 @@ def cleanup_array_fields
   def auto_set_short_label
     prefix = creator.reaction_name_prefix
     counter = creator.counters['reactions'].succ
-    self.short_label = "#{creator.initials}-#{prefix}#{counter}"
+    sh_label = self['short_label']
+    return if short_label && !short_label_changed?
```
PiTrem, line 268: not needed imo

PiTrem (changes requested):

app/models/reaction.rb
endevor-music: @PiTrem - this logic causes more problems, unfortunately. The following reaction API tests fail:

https://github.com/ComPlat/chemotion_ELN/blob/3899b7303da99c3a046f80031007e27fc7e70a0a/spec/api/chemotion/reaction_api_spec.rb#L421

https://github.com/ComPlat/chemotion_ELN/blob/3899b7303da99c3a046f80031007e27fc7e70a0a/spec/api/chemotion/reaction_api_spec.rb#L549

Screenshot 2023-06-23 113834

I am proposing that my requested logic be kept as it is.

app/models/reaction.rb
endevor-music: agreed, I'll remove once the logic discussion is resolved.

PR #1134, opened 2023-02-20
Add logs to detect problematic device while data collection

TasnimMehzabin:


FabianMauz (approved):
TO DO:
 - make small changes we discussed
 - write an rspec test to see that an exception is not breaking the loop

PR #1145, opened 2023-02-23
fix the issue that chemspectra does not work at sometimes

baolanlequang:

- [x] added code is linted

### DIFF
--- a/app/packs/src/stores/alt/stores/SpectraStore.js
+++ b/app/packs/src/stores/alt/stores/SpectraStore.js
@@ -80,9 +80,12 @@ class SpectraStore {
 
   handleLoadSpectra({ fetchedFiles, spcInfos }) {
     const spcMetas = this.decodeSpectra(fetchedFiles);
-    const newArrSpcIdx = spcMetas.map(spci => (
+    let newArrSpcIdx = spcMetas.map(spci => (
       spci.idx
     )).filter(r => r !== null);
+    if (newArrSpcIdx.length <= 1) {
+      newArrSpcIdx = [];
+    }
     this.setState({
       spcInfos,
       spcMetas,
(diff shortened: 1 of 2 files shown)


app/packs/src/stores/alt/stores/SpectraStore.js
```diff
@@ -80,9 +80,12 @@ class SpectraStore {
 
   handleLoadSpectra({ fetchedFiles, spcInfos }) {
     const spcMetas = this.decodeSpectra(fetchedFiles);
-    const newArrSpcIdx = spcMetas.map(spci => (
+    let newArrSpcIdx = spcMetas.map(spci => (
       spci.idx
     )).filter(r => r !== null);
+    if (newArrSpcIdx.length <= 1) {
```
ptrxyz, line 86: Just double checking: so we need at least two elements in our filtered array, otherwise we set it `[]`, right?
only one element is not enough?

yarn.lock
```diff
@@ -1171,6 +1171,13 @@
   dependencies:
     regenerator-runtime "^0.13.11"
 
+"@babel/runtime@^7.17.2":
```
ptrxyz, line 1174: Hm, we should avoid installing `babel/runtime` twice. Are we sure we need a downgraded version of what is already present?

app/packs/src/stores/alt/stores/SpectraStore.js
baolanlequang: Yes, because chemspectra only supports the display of multiple spectra at the same time when the spectra files is a Cyclic Voltammetry spectra for now. So in the case of only on spectra file, we treat it like the others.

yarn.lock
baolanlequang: It's a dependency of our libraries and is managed by `yarn`

PiTrem (approved):

PR #1169, opened 2023-03-06
report: alias element.analyses

PiTrem:
Report function expecting  sample.analyses 
but this data is now only under sample.containers

### DIFF
--- a/app/packs/src/apps/mydb/elements/details/reports/SectionReaction.js
+++ b/app/packs/src/apps/mydb/elements/details/reports/SectionReaction.js
@@ -11,11 +11,7 @@ const SectionReaction = ({reaction, settings, configs}) => {
          tlc_solvents, rf_value, status } = reaction;
   const dangerousProducts = dangerous_products;
 
-  const has_analyses = products.map( sample => {
-    if(sample.analyses.length != 0) {
-      return true;
-    }
-  }).filter(r => r!=null).length != 0;
+  const has_analyses = !!(products.find((sample) => sample.analysesPresent()));
 
   const showPuri = settings.purification
     && purification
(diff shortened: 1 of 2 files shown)


app/packs/src/apps/mydb/elements/details/reports/SectionReaction.js
```diff
@@ -11,11 +11,7 @@ const SectionReaction = ({reaction, settings, configs}) => {
          tlc_solvents, rf_value, status } = reaction;
   const dangerousProducts = dangerous_products;
 
-  const has_analyses = products.map( sample => {
-    if(sample.analyses.length != 0) {
-      return true;
-    }
-  }).filter(r => r!=null).length != 0;
+  const has_analyses = !!(products.find((sample) => sample.analysesPresent()));
```
FabianMauz, line 14: Why again the double exclamation marks ? 

sample.analysesPresent( ) already returns a boolean

app/packs/src/models/Element.js
```diff
@@ -119,6 +119,27 @@ export default class Element {
...
+  // Return true if the element has at least one analysis
+  analysesPresent() {
+    if (!this.container) { return false; }
+
+    const analysesContainer = this.container.children?.find((container) => container?.container_type === 'analyses');
+    if (!analysesContainer) { return false; }
+
+    const analysis = analysesContainer.children?.find((container) => container?.container_type === 'analysis');
+    return !!analysis;
+  }
```
FabianMauz, line 131: could that function be shorten to
```suggestion
analysesPresent() {
   return analyses().length > 0
} 
```

FabianMauz (approved):
Some minor questions

app/packs/src/models/Element.js
PiTrem: Indeed.
But I think it proposed change could be faster when the element has heaps of analyses :) 
Also  the functions  defined above that one could also have re-used, but I think they   they need a rewrite, so did not do that here.

PR #1181, opened 2023-03-08
Issue 902 export reasearch plan rev

PiTrem:
remove listener from component that was triggering the  processing of  data unrelated to the component

( + some linting)


PiTrem:
oops wrong branch

PR #1194, opened 2023-03-13
Extend search

beque:
Add search modal for all searches
The search result is shown in modal as preview


adambasha0:
Hi, 
I found the following bug. copy to clipboard on search result does not work as expected.

https://github.com/ComPlat/chemotion_ELN/assets/55552142/4f11876f-5741-49a3-916c-80b130240d3e

adambasha0:
Another Bug: on advanced search searching exact sample name should work as expected! but it does not. can you check? thank you
https://github.com/ComPlat/chemotion_ELN/assets/55552142/d75a4d7d-8e4a-4d7f-ae6c-072de647ec5e

beque:
@adambasha0 Hi,
I have fixed the copy to clipboard.

There could be many reasons why the server was slow, when you are testing the search.
It is just a staging server. I don't know if it's really comparable with a production server.

In your second video you are searching for a molecule name.
But in the advanced search for samples you can only search for the name of a sample.
Or other fields of a sample but not for a molecule name.
So it is correct that you do not get a result.

In the detail search I will add a molecule name field.
Than it will be possible to search for a molecule name

PiTrem (approved):

PR #1221, opened 2023-03-21
reset melting & boiling point for products on reaction copy

adambasha0:

- [x] added code is linted

### DIFF
--- a/app/packs/src/apps/mydb/elements/details/samples/SampleDetails.js
+++ b/app/packs/src/apps/mydb/elements/details/samples/SampleDetails.js
@@ -349,7 +349,8 @@ export default class SampleDetails extends React.Component {
   checkMolfileChange() {
     const { trackMolfile } = this.state;
     const { sample } = this.props;
-    if (trackMolfile !== sample.molfile) {
+    // !sample.isNew to allow setting mp & bp for new samples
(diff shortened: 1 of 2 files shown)


app/packs/src/apps/mydb/elements/details/samples/SampleDetails.js
```diff
@@ -349,7 +349,8 @@ export default class SampleDetails extends React.Component {
   checkMolfileChange() {
     const { trackMolfile } = this.state;
     const { sample } = this.props;
-    if (trackMolfile !== sample.molfile) {
+    // !sample.isNew to allow setting mp & bp for new samples
+    if (trackMolfile !== sample.molfile && !sample.isNew) {
```
baolanlequang, line 353: What happens when `trackMofile` or `sample` is `undefined` or `null`?

adambasha0: The trackMolfile has initial state defined. Undefined or null should not happen. Let me know if you can produce an error. If it happens, I guess if statement would not fire and nothing will happen in this case

baolanlequang: Ok, then it is not the case.

baolanlequang (approved):

PR #1223, opened 2023-03-22
Minor image annotation fixes

FabianMauz:
Add opacity as styling element to svg allowlist. 

The original reason why i created the branch was the issue found by Silvia that free hand drawings without a filling got a filling after the save. But i can not reproduce it. 

@reviewer, please check if you can reproduce the problem.


JanCBrammer (approved):
Could reproduce the issue (i.e., opacity wasn't preserved) on the main branch on the staging server. I confirm that this PR fixed the issue, i.e., the opacity is preserved.

PR #1300, opened 2023-04-26
Remove redundant command from devcontainer configuration

JanCBrammer:
The removed command is a duplicate of https://github.com/ComPlat/chemotion_ELN/blob/cd7cb6c597b155be0e6506e4e6137e429ad9f3be/package_postinstall.sh#L36


FabianMauz (approved):
Correct, link is now created via the original installation script.

Issue #1292, opened 2023-04-21
db migration fails durring update from 1.3.1 to 1.5.1


boontifex:
- I am a system/ELN admin

DB Migration `bundle exec rails db:migrate` throws the following exception:

```
2023-04-21 04:27:46 +0000, [INFO]: Migrating to SampleSvgFileCorrection (20220816070825) 
== 20220816070825 SampleSvgFileCorrection: migrating ==========================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

no implicit conversion of nil into String
/chemotion/app/db/migrate/20220816070825_sample_svg_file_correction.rb:19:in `file?'
/chemotion/app/db/migrate/20220816070825_sample_svg_file_correction.rb:19:in `block in change'
```

I tried to figure out which records are affected using rails console: `bundle exec rails console`

```ruby
date = Time.new(2022, 6, 28)

molecules = Molecule.where('created_at > :date or updated_at > :date', date: date)
messy_m = []

molecules.find_each do |m|
  svg_file = m.send(:full_svg_path)
  unless svg_file && File.file?(svg_file)
    samples = Sample.where('molecule_id = :m_id', m_id: m.id)
    samples.find_each do |s|
      puts "User_Id: '#{s.user_id}', Molecule_Id: '#{s.molecule_id}', sample_svg_file: '#{s.sample_svg_file}'"
    end    
  end
end
```

So we figured out, that:

* one molecule has no svg_file
* 26 molecules svg files are missing in filesystem
* the affected molecules are related to samples without user_id (user_id is null)

So, whats the best way to deal with this situation?

PiTrem:
until the next release is out , and if you are willing to tinker the db-migration files you can replace the following lines in 
db/migrate/20220816070825_sample_svg_file_correction.rb (line 19 and 31 ): 



```rb
 next unless File.file?(svg_file)
 ```
 with 
```rb
  next unless svg_file && File.file?(svg_file)
```


Also the following script would help rebuild SVG for those molecules (and samples) missing it.
(note that the Ketcher Service will return an SVG with no content if the molfile cannot be processed)


```rb
Molecule.find_each do |molecule|
  svg_path = molecule.send(:full_svg_path)
  next if svg_path && File.file?(svg_path)

  puts "Missing SVG for #{molecule.id}\n"

  if molecule.molfile.blank?
    puts "No molfile for #{molecule.id}\n"
    next
  end
[...]

PR #1307, opened 2023-05-05
Migration fixes

PiTrem:
resolve https://github.com/ComPlat/chemotion_ELN/issues/1292


db/migrate/20230503090936_change_xref_cas_datatype.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# rubocop:disable Rails::SkipsModelValidations
+# frozen_string_literal: true
+
```
endevor-music, line 5: these are duplicates of lines 1 and 6, and are not needed.

lib/tasks/svg.rake
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+
+namespace :svg do
+  desc "rerender SVG for molecules  \n
+     args[:all] default to false, if true, all molecules will be rerendered, \n
+     if false, only molecules with missing SVG will be rerendered"
+  task :molecule, %i[all from to] => [:environment] do |_t, args|
+    #   if true, render all molecules, otherwise only render those that have missing svg
+    # args.with_ketcher is a boolean: default to true, always check ketcher rendering service is available
```
endevor-music, line 9: Some commented examples in the file would be ideal to properly show the list of arguments and how they can be used.

Some clarification on the existence of the `with_ketcher` boolean argument is needed. It is only in the commented line 9, and isn't used in the code at all to check if the ketcher service is available before continuing.

lib/tasks/svg.rake
```diff
@@ -0,0 +1,104 @@
...
+    molecule_scope = molecule_scope.where(id: ..to) if to.positive?
+    message_header = 'Reprocessing '
+    message_header += all ? 'all' : 'missing'
+    message_header += ' molecule SVGs: '
+    puts "#{message_header}Ketcher rendering service seems to #{ketcher_running ? '' : 'not'} be available"
+
+    # if ketcher_only && !ketcher_running
+    #   puts "#{message_header}Ketcher rendering service not available, exiting"
+    #   break
+    # end
```
endevor-music, line 24: The `ketcher_only` variable is also nonexistent in the rest of the code. Clarification on its usage & existence needed.

Either:
1) this conditional should be implemented, to avoid the task running to completion with the ketcher service unavailable.
2) the user should be made aware in line 19 (& 67), that they should exit the task if the ketcher service is unavailable.

lib/tasks/svg.rake
```diff
@@ -0,0 +1,104 @@
...
+    from = args[:from].presence&.to_i || 0
+    to = args[:to].presence&.to_i || 0
+    sample_scope = Sample.all
+    sample_scope = sample_scope.where(id: from..) if from.positive?
+    sample_scope = sample_scope.where(id: ..to) if to.positive?
+
+    message_header = 'Reprocessing '
+    message_header += all ? 'all' : 'missing'
+    message_header += ' sample SVGs: '
+    puts "#{message_header}Ketcher rendering service seems to #{ketcher_running ? '' : 'not'} be available"
```
endevor-music, line 67: (see point 2 from comment on line 19)

endevor-music (changes requested):

lib/tasks/svg.rake
PiTrem: fx and moved to https://github.com/ComPlat/chemotion_ELN/pull/1308

lib/tasks/svg.rake
PiTrem: fx and moved to https://github.com/ComPlat/chemotion_ELN/pull/1308

PR #1339, opened 2023-06-05
fixed cannot load the correct url of nmrium wrapper

baolanlequang:

- [x] added code is linted

### DIFF
--- a/app/api/chemotion/chem_spectra_api.rb
+++ b/app/api/chemotion/chem_spectra_api.rb
@@ -200,12 +200,8 @@ module Chemotion
         desc 'Return url of nmrium wrapper'
         route_param :host_name do
           get do
-            if Rails.configuration.spectra.nmriumwrapper.blank?
-              { nmrium_url: '' }
-            else
-              nmrium_url = Rails.configuration.spectra.nmriumwrapper.url
-              { nmrium_url: nmrium_url }
-            end
+            nmrium_url = Rails.configuration.spectra.nmriumwrapper.url
+            { nmrium_url: nmrium_url }
           end
         end
       end
--- a/app/api/chemotion/chem_spectra_api.rb
+++ b/app/api/chemotion/chem_spectra_api.rb
@@ -201,14 +201,10 @@ module Chemotion
         route_param :host_name do
           get do
             if Rails.configuration.spectra.nmriumwrapper.blank?
-              { protocol: '', host: '', port: '' }
+              { nmrium_url: '' }
             else
               nmrium_url = Rails.configuration.spectra.nmriumwrapper.url
-              nmrium_uri = URI(nmrium_url)
-              protocol = nmrium_uri.scheme
-              host = nmrium_uri.host
-              port = nmrium_uri.port == 443 ? '' : nmrium_uri.port
-              { protocol: protocol, host: host, port: port }
+              { nmrium_url: nmrium_url }
             end
           end
         end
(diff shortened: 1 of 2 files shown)


app/api/chemotion/chem_spectra_api.rb
```diff
@@ -208,7 +208,8 @@ def raw_file(att)
               protocol = nmrium_uri.scheme
               host = nmrium_uri.host
               port = nmrium_uri.port == 443 ? '' : nmrium_uri.port
-              { protocol: protocol, host: host, port: port }
+              nmrium_origin = port == '' ? "#{protocol}://#{host}" : "#{protocol}://#{host}:#{port}"
```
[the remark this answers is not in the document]
baolanlequang: I want to do it in the backend, the frontend only needs to get the values and compare. The `nmrium_origin` is used in the frontend to fill the message from nmrium wrapper.
The comparison of port value is to make sure we don't use port 443 in the past.
But as you suggested, we're using `nmriumwrapper` in the `spectra.yml` instead of separated values, I leave it as to cover the older config file.
I am open to changing this, but is it necessary or not?

app/packs/src/components/nmriumWrapper/NMRiumDisplayer.js
```diff
@@ -95,15 +92,27 @@ export default class NMRiumDisplayer extends React.Component {
...
+          let nmriumData = null;
+          if (eventData.data) {
+            const { state } = eventData.data;
+            if (state) {
+              nmriumData = state;
+            }
+            else {
+              nmriumData = eventData.data;
+            }
+          }
```
ptrxyz, line 105: Hm, the whole `if` tree looks a complex to me, making it hard to follow:
image

can we simplify the code to:

```
nmriumData = eventData.data?.state || eventData.data;
```

image

This leads to the same results except that in the first two cases, we end up with undefined instead of null, which could probably further been worked around with `nmriumData = ( ....the line above... ) || null` in case it makes a difference later in the code.

baolanlequang: Yes, we can keep it simple. But I doubt is it clear to follow up, especially when we will update our site based on nmrium-wrapper returned data.

app/packs/src/components/nmriumWrapper/NMRiumDisplayer.js
```diff
@@ -95,15 +92,27 @@ export default class NMRiumDisplayer extends React.Component {
...
+          
+          if (!nmriumData) {
```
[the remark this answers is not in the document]
ptrxyz: This one is still pending. Can you please add some explaining comments or maybe even proper JSDoc annotations to make the types somehow understandable to someone from outside?

app/api/chemotion/chem_spectra_api.rb
```diff
@@ -201,14 +201,10 @@ def raw_file(att)
...
+              { nmrium_url: '' }
             else
               nmrium_url = Rails.configuration.spectra.nmriumwrapper.url
-              nmrium_uri = URI(nmrium_url)
-              protocol = nmrium_uri.scheme
-              host = nmrium_uri.host
-              port = nmrium_uri.port == 443 ? '' : nmrium_uri.port
-              { protocol: protocol, host: host, port: port }
+              { nmrium_url: nmrium_url }
             end
```
ptrxyz, line 208: Starting from 203 following, couldn't this simply be:

```rb
{ nmrium_url: Rails.configuration.spectra.nmriumwrapper.url.presence || '' }
```

Maybe even without the `|| ''`?

ptrxyz (changes requested):

app/api/chemotion/chem_spectra_api.rb
baolanlequang: @ptrxyz I updated

PiTrem (approved):

PR #1342, opened 2023-06-07
Element tag refactor

mehreenmansur:
- update element tag UI
- Migration of old element tag data
- update SQL for export functions


mehreenmansur:
Screenshot from 2023-06-07 08-31-19

PR #1350, opened 2023-06-16
updating group admin ui

mekkyz:
- Updated ui admin rights

- All admins can now delete/add members from/to the group or delete the groups

- Normal users can only delete themselves

- The "admin by" field now have all the admins and not only the creator

- Creating change to admin button for admins is complicated and is like a new feature. Was not implemented at all? Should it be implemented?

Fixes ComPlat/chemotion#10


PiTrem (approved):

mekkyz:
Fixes ComPlat/chemotion/#10

Issue #1340, opened 2023-06-05
Download of annotated image not working


FabianMauz:
- I am a developer

# Expected Behavior

The annotated image should be part of the zip file.

# Current Behavior

When an dataset in the analysis tab of an element with annotated image as attachment is downloaded only the original image is present in the zip file. The annotated image should also be in there.

## Steps to Reproduce

1. create a sample
2. add an image
3. annotate the image
4. save everything
5. click the download button in analyses context
6. zip file should contain the annotated image

PR #1358, opened 2023-06-20
1340 download of annotated image not working

FabianMauz:
This pull request will fix the bug https://github.com/ComPlat/chemotion_ELN/issues/1340 .

It will add annotated images into the zip folder if user downloads a container dataset.

### DIFF
--- a/app/api/chemotion/attachment_api.rb
+++ b/app/api/chemotion/attachment_api.rb
@@ -44,8 +44,8 @@ module Chemotion
       error!(message, 404)
     end
 
-    resource :attachments do # rubocop:disable Metrics/BlockLength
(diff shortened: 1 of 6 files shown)


JanCBrammer:
I could reproduce the bug on the current `main` branch according to the instructions in #1340 (I uploaded a .png, and annotated it; the subsequent download only contained the original .png, the annotated .png was missing).
Subsequently, I switched to the `1340-download-of-annotated-image-not-working` branch, and confirmed that the bug is fixed (i.e., the annotated .png is now downloaded alongside the original .png).

@PiTrem, can I merge?

JanCBrammer (approved):
Looks good to me.

PR #1362, opened 2023-06-22
notification timestamps and formatting notification button

mekkyz:
Fixes https://github.com/ComPlat/chemotion/issues/12
Fixes https://github.com/ComPlat/chemotion/issues/20

- used "prettier" to reformat the code.
- solved eslint issues
- getting the timezone String from backend to correctly present the local time
- DidMount has now startActivityDetection to make the notification button always available
- Did some ui change on the notification button because its size and centering were off

What is new for users:
- the notification button is now a bell instead of the messaging icon and be always shown in the default status when notification = 0.
- it will change to yellow and notification counter will start counting whenever there are notifications
- the text "message" is now "notification"
- the button has now the right size comparable to the button next to it.
- the icon is now centered in the button.

This is how it was when there were no notifications.

And now:

When there are notifications

And now:

### DIFF
--- a/app/packs/src/utilities/timezoneHelper.js
+++ b/app/packs/src/utilities/timezoneHelper.js
@@ -1,0 +1,34 @@
+const formatDate = (dateString) => {
+  const [datePart, timePart] = dateString.split(', ');
+  const [day, month, year] = datePart.split('.');
+  const [hours, minutes, seconds] = timePart.split(':');
+
+  // format date to js readable date
+  const formattedDate = `${year}-${month}-${day}T${hours}:${minutes}:${seconds}`;
+
+  const creationDateObj = new Date(formattedDate);
+
+  // local timezone offset
+  const timeZoneOffset = creationDateObj.getTimezoneOffset();
+  const offsetHours = Math.floor(Math.abs(timeZoneOffset / 60));
+  const offsetMinutes = Math.abs(timeZoneOffset) % 60;
+
+  // apply time offset
+  creationDateObj.setHours(
+    creationDateObj.getHours()
+    + (timeZoneOffset < 0 ? offsetHours : -offsetHours)
+  );
+  creationDateObj.setMinutes(
+    creationDateObj.getMinutes()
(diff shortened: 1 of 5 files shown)


app/packs/src/utilities/timezoneHelper.js
```diff
@@ -0,0 +1,34 @@
...
+  const [hours, minutes, seconds] = timePart.split(':');
+
+  // format date to js readable date
+  const formattedDate = `${year}-${month}-${day}T${hours}:${minutes}:${seconds}`;
+
+  const creationDateObj = new Date(formattedDate);
+
+  // local timezone offset
+  const timeZoneOffset = creationDateObj.getTimezoneOffset();
+  const offsetHours = Math.floor(Math.abs(timeZoneOffset / 60));
```
ptrxyz, line 13: to calculate the offset, simply use `Math.trunc` or `parseInt`. Those do the same but preserve the sign so you can get rid of the comparisons in line 19 and 23.

spec/javascripts/utils/timezoneHelper.spec.js
```diff
@@ -0,0 +1,50 @@
...
+  it('should correctly format the date', () => {
+    const input = '23.06.2023, 12:00:00';
+    const expectedOutput = 'Friday, June 23, 2023 at 12:00 PM';
+
+    const result = formatDate(input);
+
+    expect(result).toEqual(expectedOutput);
+  });
+
+  it('should handle different date formats', () => {
```
ptrxyz, line 14: Isnt this the same test as the one before just with a different date but the same format?

ptrxyz (changes requested):

app/packs/src/utilities/timezoneHelper.js
mekkyz: corrected

spec/javascripts/utils/timezoneHelper.spec.js
mekkyz: deleted the second one. I was testing with different dates

PiTrem (approved):

PR #1366, opened 2023-06-23
Fix bug in `Export reaction from selection`

JanCBrammer:
How to reproduce the bug:
* select a collection as well as one or multiple reactions from that collection
* click `Export reactions from selection` in the upper task bar / menu
* click `Smiles Export` in the  `Reaction Smiles Export` modal that's popping up
* the generated file is empty except for "{"error":"columns is invalid"}"

Once the bug is fixed (see diff in this PR), repeat the procedure above.
Now the generated file contains the expected Smiles output.


FabianMauz (approved):
Can confirm the fix.

The reason was a failed validation at 
https://github.com/ComPlat/chemotion_ELN/blob/c5904765808c9cad090c3cc759ce2900762eb8f9/app/api/helpers/report_helpers.rb#L8

PR #1380, opened 2023-06-28
Notification timestamps

mekkyz:
- Now using timezone info from backend for notification and adjusting the time accordingly


PiTrem:
superseded by https://github.com/ComPlat/chemotion_ELN/pull/1389

Issue #1384, opened 2023-06-29
user select in Admin-UI feature  to display full user info


PiTrem:
chemotion_ELN v1.6.1

analogue to  #1360 

![image](https://github.com/ComPlat/chemotion_ELN/assets/15799181/3d69f20c-45c2-4a3b-b485-a765c02ffa4d)

PR #1385, opened 2023-06-29
User select in UI feature

PiTrem:
Unify user-find-by-name api endpoint:

- remove duplicated endpoint in admin api
- extract option-formatting from components to a utility function
- user api endpoint  to also use simple user entity

resolve https://github.com/ComPlat/chemotion_ELN/issues/1384


app/api/chemotion/user_api.rb
```diff
@@ -5,14 +5,14 @@ class UserAPI < Grape::API
...
       get 'name' do
-        unless params[:name].nil? || params[:name].empty?
-          { users: User.where(type: %w[Person Group]).by_name(params[:name]).limit(3)
-                       .select('first_name', 'last_name', 'name', 'id', 'name_abbreviation', 'name_abbreviation as abb', 'type as user_type') }
-        else
-          { users: [] }
-        end
+        return { users: [] } if params[:name].blank?
+
+        users = User.where(type: params[:type]).by_name(params[:name]).limit(3)
```
FabianMauz, line 14: type is optional, i think it would return always [ ] if no type is given as argument

FabianMauz, line 14: maybe limit amount again as rest parameter

app/packs/src/fetchers/AdminFetcher.js
```diff
@@ -185,6 +185,15 @@ export default class AdminFetcher {
       .catch((errorMessage) => { console.log(errorMessage); });
   }
 
+  static fetchUsersByNameType(name, type, limit = 5) {
+    return fetch(`/api/v1/admin_user/listUsers/byname.json?type=${type}&name=${name}&limit=${limit}`, {
```
FabianMauz, line 189: Maybe use URLSearchParams for better clarity, but i think the current code is clear enough for me

app/packs/src/utilities/selectHelper.js
```diff
@@ -0,0 +1,26 @@
+// Format data for react-select component when using with users data as returned by the API
+// { users: [{ id: 1, name: 'John Doe', initials: 'JD', type: 'Person' }, ...]}
+
+const filterLabel = (user, withType = false) => (
+  withType ? `${user.name} (${user.initials} - ${user.type})` : `${user.name} (${user.initials})`
+);
+
+const filterData = (data, currentUserId = null) => {
```
FabianMauz, line 8: maybe rename method to something that makes more clear what it does like: getUserFromData

FabianMauz (changes requested):
Looks good to me except for one maybe bug. Lets have a talk about it.

app/api/chemotion/user_api.rb
PiTrem: non admin query are caped

PiTrem: default value added

app/api/chemotion/admin_user_api.rb
```diff
@@ -98,8 +114,9 @@ class AdminUserAPI < Grape::API
             end
           end
 
-          if params[:reconfirm_user].present?
-            user.update_columns(email: user.unconfirmed_email, unconfirmed_email: nil) if params[:reconfirm_user] == true
+          if params[:reconfirm_user].present? && (params[:reconfirm_user] == true)
+            user.update_columns(email: user.unconfirmed_email,
```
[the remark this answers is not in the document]
PiTrem: maybe

PR #1389, opened 2023-07-03
Unify datetime format

PiTrem:
- use formatDate utility to format displayed timestamps
 (reduce the number of  'import  moment.js' )

- Entity: return time zone in timestamp

- more entities are now using  the same exposure (expose_timestamp) for timestamps

- shoudl resolve timezone diff in comment-feature  and notification-feature


FabianMauz (approved):
Looks good

PR #1399, opened 2023-07-10
add max line length eslint rule

FabianMauz:
add one new eslint rule:

max line length max 120 signs, ignoring comments


JanCBrammer (approved):
LGTM.

Issue #1382, opened 2023-06-29
White screen for research plan after deleting linked sample/reaction


BenjaminGolub:
- I am a ELN admin

# Expected Behavior

If a sample/reaction which is linked in a research plan, is deleted, it should still be possible to open the research plan.

# Current Behavior

If a sample/reaction which is linked in a research plan (with the tool "drop sample/reaction"), clicking on the research plan leads to a white screen:
https://cloud.tu-braunschweig.de/s/GLaz2aZZ2b6J95P

## Steps to Reproduce

1. create a research plan and link a sample/reaction to it with the drag-and-drop
2. delete the sample/reaction from your eln 
3. now click on the research plan with the linked sample/reaction -> white screen

 - Windows 10
 - Chrome Browser 
 - Chemotion ELN v1.6.0

PR #1469, opened 2023-08-09
fix: display the not-accessible panel for 401 status on sample fetched by id

PiTrem:
* fix navigation to existing sample without permission:
 when a query to  fetch sample by id returns 401, the _not accessible_ panel should be rendered

image

* now also render the info panel for record (sample/reaction) not found (404)

image


PiTrem:
seems to handle #1382 
image

FabianMauz (approved):
Looks good for me

mekkyz (approved):

Issue #950, opened 2022-10-05
Vessel Element


dpatelev:
- I am a developer

# Description
Ongoing implementation of a new element to record reaction vessels and other equipments in the ELN.

The purpose of this element is to allow users to create a vessel template (similar to molecules) for their reaction vessels / equipment, and assign a vessel entity (akin to samples) to a reaction. 

The basis of this comes from a generic element implementation, that allows for the user to pick between vessels (various flasks / beakers), reactors (flow reactors, electrochemical cells) and other equipment (TBD), and state the material details and volume.

Back end - database model - https://github.com/ComPlat/chemotion_ELN/pull/1548
API - functional
Front end - functional, needs refining & CSS

jhmegorei:
Can you add a description what reaction vessels are and how you would model them in the context of the ELN?

endevor-music:
> Can you add a description what reaction vessels are and how you would model them in the context of the ELN?

I hope the following information helps,
Devang

endevor-music:
For developers: please see PR #1476 for the initial model and API implementation for this element.

cbuggle:
The  "Back end - database model" part is provided by [#1548] which when merged eventually renders [#1476] obsolete.

PR #1476, opened 2023-08-15
Vessel Model + API Implementation

endevor-music:
This PR includes the main implementation of the back-end models and API for the new vessel element, which is currently in development. The primary aim for this PR is to have the fixed back end in the codebase, so those working on the reaction process editor can continue their development.

Details:
- Migration file
- Models
- API + Usecases
- Specs

Once merged into the codebase, development of the main element will continue: 
- #950 
- https://github.com/ComPlat/chemotion_ELN/tree/vessel-element


cbuggle:
I think this PR is no longer required. 

The main purpose was to provide a provisional excerpt of `vessel-element` defining the vessel models as a common code base for the side project  'ReactionProcessEditor'.  This has now been accomplished in [#1548 ].

I think it is not ready to merge yet anyway as 2 of my comments are IMHO blockers (schema diverging from migrations, missing/overridden `User#has_many :vessels` as inverse of `Vessel#belongs_to :creator`). Also it contains way more than needed, particularly a lot of ELN-only frontend javascript which then again seems to be work in progress given the amount of related commits in the parallel `vessel-element` branch. Also the API endpoints and -entities will be tailored to the 'ReactionProcessEditor'  in their own module anyway so we don't need them right away).

endevor-music:
Closing this as unneeded due to stripped back implementation in PR https://github.com/ComPlat/chemotion_ELN/pull/1548

PR #1485, opened 2023-08-17
Bugfix: replace toSorted with manual sorting in SampleTaskInbox

jhmegorei:
toSorted is only available in really new browsers, which is unfortunately not a given within academia.
Therefore the function call was replaced with another sorting routine that uses basic JS array sort()


FabianMauz (approved):
Tested it on my machine with firefox 114 and it is working

jhmegorei:
@PiTrem can this be merged?

Issue #1497, opened 2023-08-23
Delete of literature not working


FabianMauz:
- I am a developer

When deleting an entry in the references tab of a sample, a red popup appears with "Add Literature"

# Current Behavior

![image](https://github.com/ComPlat/chemotion_ELN/assets/23172219/e214a174-980d-4427-a747-baba22ea75e6)

The reference seems not to be deleted. 
After refreshing the page ore reloading the sample the reference was removed.

Tested with: 

 - Firefox Browser 116.0.3 (64-bit)
 - Chrome Version 116.0.5845.97 (Offizieller Build) (64-Bit)

PR #1502, opened 2023-08-24
1497 delete of literature not working

FabianMauz:
This PR solves issue https://github.com/ComPlat/chemotion_ELN/issues/1497. 

The problem was an empty response body which was tried to transform to JSON format. I changed the literature API, so it will return now an status code and fixed the Fetcher that it reacts to the status code and only tries to convert the body to JSON if there is another status code than 200.

I also added a UI feedback in form of a pop up if the literal was successfully removed.

There was one accidentally commit https://github.com/ComPlat/chemotion_ELN/pull/1502/commits/9e390a982441a5aa4b198ff4ef5c3114404e04b1, but i reverted it with https://github.com/ComPlat/chemotion_ELN/pull/1502/commits/4002688acd786fe5dada15f38925209c9ecd03a5, so just ignore it.

### DIFF
--- a/spec/api/literature_api_spec.rb
+++ b/spec/api/literature_api_spec.rb
@@ -18,8 +18,8 @@ describe Chemotion::LiteratureAPI do
         ref: {
           is_new: true,
           title: 'testpost',
-          url: 'test'
-        }
+          url: 'test',
+        },
       }
     end
 
@@ -34,7 +34,7 @@ describe Chemotion::LiteratureAPI do
         end
 
         it 'responds 200' do
-          expect(response.status).to be 200
+          expect(response).to have_http_status :ok
         end
 
         it 'is able to get literatures by reaction Id' do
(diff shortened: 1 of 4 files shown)


spec/api/literature_api_spec.rb
```diff
@@ -1,93 +1,103 @@
...
+            .literatures.find_by(title: 'testpost')
+        expect(l).not_to be_nil
       end
     end
+  end
 
-    describe 'POST /api/v1/literatures' do
-      context 'with valid parameters' do
-        before { post '/api/v1/literatures', params: params }
+  describe 'DELETE /api/v1/literatures' do
```
JanCBrammer, line 74: Maybe also test for `400` (trying to delete missing literal)?

JanCBrammer (approved):
I can confirm #1497 on the latest main.
I locally confirmed that deleting literature is now properly handled in the UI (literature disappears, and a pop-up confirms deletion) on `1497-delete-of-literature-not-working`.

@PiTrem, this can be merged as far as I'm concerned.

PR #1520, opened 2023-09-08
Add annotation support for sample svg

jhmegorei:
I tried (and failed) to refactor the ImageAnnotationModalSVG component to no longer use an iframe to display the editor.

But I managed to refactor the component so it no longer does the fetch of the annotation within.
Now we can supply SVG Code via prop and handle the modified SVG code as a return value.

This allows reuse of the component in other contexts (our target is to implement annotation on the molecule structure SVG files of the sample).


app/packs/src/apps/mydb/elements/details/samples/SampleAnnotationDeleteButton.js
```diff
@@ -0,0 +1,116 @@
...
+              this.props.horizontalAlignment ? this.props.horizontalAlignment : ""
+            }
+            onClick={() => this.setState({ showTooltip: true })}
+            ref={(button) => { this.target = button; }}
+          >
+            <i className="fa fa-trash" aria-hidden="true" />
+          </Button>
+        </OverlayTrigger>
+        <Overlay
+          containter={this}
```
jhmegorei, line 47: Typo

PR #1574, opened 2023-10-11
feat: Table of data types and chemspectra layouts in the ELN Admin

f-idiris:
- chemspectra admin page with a table of data types and chemspectra layouts
- Entries into the table are stored in a json file in the ELN public folder which is sent along with the JCAMP files to chemspectra
image


app/api/chemotion/chem_spectra_api.rb
```diff
@@ -196,6 +197,70 @@ def raw_file(att)
...
+            }, body: formatted_data.to_json)
+          if response.code == 400
+            error_response = JSON.parse(response.body)
+            error_message = error_response['message']
+            error!(error_message, 400)
+          end
+        end
+      end
+
+      resource :spectra_layouts do
```
baolanlequang, line 224: Why don't we just use the same endpoint for get spectra type?

app/packs/src/fetchers/ChemSpectraFetcher.js
```diff
@@ -0,0 +1,42 @@
+import 'whatwg-fetch';
+
+export default class ChemSpectraFetcher {
```
baolanlequang, line 3: This should be update after using the same endpoint.

app/api/chemotion/chem_spectra_api.rb
```diff
@@ -196,6 +234,69 @@ def raw_file(att)
         end
       end
 
+      resource :spectra_layouts do
```
baolanlequang, line 237: Cannot use new datatype.
For example, when adding a new NMR type as `Test NMR`, it cannot be used after match.
This is the test file, just change `.txt` to `.jdx`
[test.txt](https://github.com/ComPlat/chemotion_ELN/files/13039673/test.txt)

app/packs/src/fetchers/ChemSpectraFetcher.js
```diff
@@ -0,0 +1,49 @@
...
+  static fetchSpectraLayouts() {
```
baolanlequang, line 4: Is there any test for this?

app/packs/src/apps/admin/ChemSpectraLayouts.js
```diff
@@ -0,0 +1,263 @@
+import React, { Component } from 'react';
+import ChemSpectraFetcher from 'src/fetchers/ChemSpectraFetcher';
+import {
+  Table, Button, Form, FormControl, Modal, Panel, FormGroup, ControlLabel, Popover, OverlayTrigger, ButtonGroup, Alert
+} from 'react-bootstrap';
+import Select from 'react-select';
+
+export default class ChemSpectraLayouts extends Component {
```
baolanlequang, line 8: Is there any test for this?

baolanlequang (approved):

baolanlequang (approved):

PR #1639, opened 2023-11-27
fix: collection management right click on the add button to not drag things around

FabianMauz:
### This PR solves 2 different problems in the Collection management component:

- It makes the root collection undraggable to mitigate the bug in the used library mentioned in the linked issue
- It makes the buttons undraggable as mentioned in the issue

### Remaining problems

- There are still some draggable areas in the tree component. I tried to minimize them but they are still there. In my opinion instead of fixing these remaining issues we should look for another tree library which is not as old and undocumented than the current one.

Because it is mainly frontiest frontend stuff and css i skipped js-unit testing as i would only test trivial attributes.

### DIFF
--- a/app/assets/stylesheets/collection_management.scss
+++ b/app/assets/stylesheets/collection_management.scss
@@ -92,9 +92,10 @@
       margin-left: 15px;
       float: left;
       padding-top: 4px;
+      padding-right: 0px;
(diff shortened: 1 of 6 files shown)


endevor-music (approved):
Looks good to me. :)

PR #1664, opened 2023-12-11
refactor: extract` SpectraEditorButton` to dedicated component

JanCBrammer:
I moved the `SpectraEditorButton` component to a dedicated file.
Before, the component was (re-)defined in three other component files (presumably copy-pasted):
`ReactionDetailsContainer`, `ResearchPlanDetailsContainer`, and `SampleDetailsContainer`.

FYI, the button looks like this (in the `Analyses` tab):
screenshot

### DIFF
--- a/app/packs/src/apps/mydb/elements/details/researchPlans/analysesTab/ResearchPlanDetailsContainers.js
+++ b/app/packs/src/apps/mydb/elements/details/researchPlans/analysesTab/ResearchPlanDetailsContainers.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { PanelGroup, Panel, Tooltip, Button, OverlayTrigger, SplitButton, ButtonGroup, MenuItem } from 'react-bootstrap';
+import { PanelGroup, Panel, Button } from 'react-bootstrap';
 import Container from 'src/models/Container';
 import ContainerComponent from 'src/components/container/ContainerComponent';
 import QuillViewer from 'src/components/QuillViewer';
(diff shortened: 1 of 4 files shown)


baolanlequang (commented):
Here are some issues that I found:
- NMRium button is disabled on analysis of Sample and Research plan

https://github.com/ComPlat/chemotion_ELN/assets/9006482/60811bdd-05c6-4ee2-b703-cb0edb6ce8bf

- When Chemspectra backend is disabled, `Reprocess` button also is disabled

https://github.com/ComPlat/chemotion_ELN/assets/9006482/d9d2f7c0-57b8-4073-a0ec-f97a0ee0127b

app/packs/src/apps/mydb/elements/details/researchPlans/analysesTab/ResearchPlanDetailsContainers.js
```diff
@@ -14,101 +14,8 @@ import SpectraActions from 'src/stores/alt/actions/SpectraActions';
...
-        >
-          <ButtonGroup className="button-right">
-            <Button
-              id="spectra-editor-split-button"
-              pullRight
-              bsStyle="info"
-              bsSize="xsmall"
-              onToggle={(open, event) => { if (event) { event.stopPropagation(); } }}
-              onClick={toggleNMRDisplayerModal}
-              disabled={!hasJcamp}
```
JanCBrammer, line 82: Note that the `disabled` flag of the NMRium button did not depend on `element.can_update` until now.
For `Reaction` and `Sample` the availability of the NMRium button *does* depend on `element.can_update`. 
Is there a reason for treating `ResearchPlan`s as a special case?

JanCBrammer (commented):
The `element` argument to `SpectraEditorButton` was missing in `SampleDetailsContainersAux`.
I forgot to change `sample` to `element` (see ffc4e07ee28a57801165780f685ed8f58137413b), good catch!

Could you try again?

app/packs/src/apps/mydb/elements/details/researchPlans/analysesTab/ResearchPlanDetailsContainers.js
baolanlequang: @JanCBrammer The `NMRium` button is still disabled on `ResearchPlan`. 
For the `element.can_update`, only `SampleEntity` and `ReactionEntity` have the value `can_update` while `ResearchPlanEntity` doesn't have it. You can check the `sample_entity.rb`, `reaction_entity.rb`  and  `research_plan_entity.rb` to see that.

JanCBrammer: > The `NMRium` button is still disabled on `ResearchPlan`

Ok, so the `NMRium` button is *supposed* to be disabled for research plans?

app/packs/src/apps/mydb/elements/details/researchPlans/analysesTab/ResearchPlanDetailsContainers.js
```diff
@@ -14,101 +14,8 @@ import SpectraActions from 'src/stores/alt/actions/SpectraActions';
...
-          </MenuItem>
-        </SplitButton>
-      </ButtonGroup>
-      ) : (
-        <Button
-          bsStyle="warning"
-          bsSize="xsmall"
-          className="button-right"
-          onClick={confirmRegenerate}
-          disabled={!hasJcamp || !element.can_update || !hasChemSpectra}
```
JanCBrammer, line 61: @baolanlequang, note that the `element.can_update` property is used for research plans.

app/packs/src/apps/mydb/elements/details/researchPlans/analysesTab/ResearchPlanDetailsContainers.js
baolanlequang: Yes, in the case that there is not `NMRium` service. So you can ignore the `can_update` value for `ResearchPlan` because it's always `true`

app/packs/src/apps/mydb/elements/details/researchPlans/analysesTab/ResearchPlanDetailsContainers.js
baolanlequang: Maybe it was implemented before to guarantee the data model of `ResearchPlan`, but to be honest, I'm not sure.

Issue #1615, opened 2023-11-09
Copy research plan


nicolejung:
One should be able to copy all elements in chemotion ELN. The copy feature is currently missing for the research plan.

PiTrem:
resolved with #1667

PR #1667, opened 2023-12-13
feat: Copy research plan 

mehreenmansur:
Feature for [1615](https://github.com/ComPlat/chemotion_ELN/issues/1615)

### DIFF
--- a/app/api/chemotion/research_plan_api.rb
+++ b/app/api/chemotion/research_plan_api.rb
@@ -76,6 +76,7 @@ module Chemotion
           body: params[:body]
         }
 
+        attributes.delete(:can_copy)
         research_plan = ResearchPlan.new attributes
         research_plan.creator = current_user
         research_plan.container = update_datamodel(params[:container])
--- a/app/api/chemotion/research_plan_api.rb
+++ b/app/api/chemotion/research_plan_api.rb
@@ -83,7 +83,7 @@ module Chemotion
         research_plan.save!
         research_plan.save_segments(segments: params[:segments], current_user_id: current_user.id)
         clone_attachs = params[:attachments]&.reject { |a| a[:is_new] }
-        research_plan.clone_existing_attachments(clone_attachs, research_plan.id, current_user.id) if clone_attachs
+        research_plan.clone_existing_attachments(clone_attachs, current_user.id) if clone_attachs
 
         if params[:collection_id]
           collection = current_user.collections.where(id: params[:collection_id]).take
--- a/app/api/chemotion/research_plan_api.rb
+++ b/app/api/chemotion/research_plan_api.rb
@@ -68,6 +68,7 @@ module Chemotion
         optional :collection_id, type: Integer, desc: 'Collection ID'
         requires :container, type: Hash, desc: 'Containers'
         optional :segments, type: Array, desc: 'Segments'
+        optional :attachments, type: Array, desc: 'Attachments'
       end
       post do
         attributes = {
@@ -80,7 +81,8 @@ module Chemotion
         research_plan.container = update_datamodel(params[:container])
         research_plan.save!
         research_plan.save_segments(segments: params[:segments], current_user_id: current_user.id)
-
+        clone_attachs = params[:attachments]&.reject { |a| a[:is_new] }
+        research_plan.clone_existing_attachments(clone_attachs, research_plan.id, current_user.id) if clone_attachs
 
         if params[:collection_id]
           collection = current_user.collections.where(id: params[:collection_id]).take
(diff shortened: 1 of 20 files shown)


FabianMauz:
I tried to copy a researchplan on uncharted branch, but rp did not appear in the target collection:
image
image
image

FabianMauz:
> I tried to copy a researchplan on uncharted branch, but rp did not appear in the target collection: image image image

Fixed in latest commit

FabianMauz:
While testing on uncharted branch, some thingies jumped to my attention:
- Copying a RP with analyses -> the **analyses** are not in the new, copied RP
- Copying a RP with references-> the **references** are not in the new, copied RP
- Copying a RP with wellplates-> the **wellplates** are not in the new, copied RP
- Copying a RP with metadata-> the **metadata** is not in the new, copied RP

Another problem seems to be the copy action for **attachments**:
The copied RP has the same attachment (linked via id) as the original RP.If the user is then annotating the e.g. an image in the original RP or editing a xlsx document the changes are automatically shipped to all copies of the RP.

app/packs/src/models/ResearchPlan.js
```diff
@@ -267,4 +268,37 @@ export default class ResearchPlan extends Element {
...
+    copy.container = Container.init();
+    copy.is_new = true;
+    copy.is_copy = true;
+    copy.can_update = true;
+    copy.can_copy = true;
+
+    return copy;
+  }
+
+  static copyFromResearchPlanAndCollectionId(research_plan, collection_id) {
```
FabianMauz, line 285: i think here are some more elements of the RP missing: metadata, references, wellplates, ...

And of course the generic segments

app/packs/src/models/Attachment.js
```diff
@@ -36,6 +37,11 @@ export default class Attachment extends Element {
     this._preview = preview;
   }
 
+  static buildCopy(_attachments) {
+    const newAttachments = cloneDeep(_attachments);
```
FabianMauz, line 41: Cloning of an attachment is a little bit tricky because you do not only want to copy the JS element but maybe also the corresponding file. I think here we should discuss about the future approach/goal of the copy method.

app/packs/src/apps/mydb/elements/details/researchPlans/ResearchPlanDetails.js
```diff
@@ -481,7 +483,16 @@ export default class ResearchPlanDetails extends Component {
   } /* eslint-enable */
 
   renderPanelHeading(researchPlan) {
+    const { currentCollection } = UIStore.getState();
+    const defCol = currentCollection && currentCollection.is_shared === false &&
```
FabianMauz, line 487: maybe extract as method with a nice name or rename variable to make more clear that this is the collection we want to copy the RP into.

app/models/research_plan.rb
```diff
@@ -98,15 +99,27 @@ def svg_files
     svg_files
   end
 
+  def clone_existing_attachments(attachments, attachment_id, current_user_id)
```
FabianMauz, line 102: It seems that the attachments are copied in terms of a database entry but not on the filesystem which leads to the problem i mentioned in the comments section

app/api/chemotion/research_plan_api.rb
```diff
@@ -68,19 +68,22 @@ class ResearchPlanAPI < Grape::API
         optional :collection_id, type: Integer, desc: 'Collection ID'
```
FabianMauz, line 68: I would not reuse the API endpoint for creating but instead create a new one which only handles the copying of a RP to avoid increasing of complexity

FabianMauz (changes requested):
I noted down my remarks in the comment section of the PR and at the files.

mehreenmansur:
> app/packs/src/apps/mydb/elements/details/researchPlans/ResearchPlanDetails.js

It was discussed that the initial scope of the feature will not include copying of analyses, references, wellplates, metadata, generic segments within the research plan.

app/api/chemotion/research_plan_api.rb
mehreenmansur: This is to maintain consistency of the similar use copying methods of samples & reactions. Refactor and separate API points can be created for all of them as another task.

FabianMauz:
After the latest commit it seems that after a copy action the attachments in the database are not the same anymore but on file system only one attachment is found. 
image
image
It seems that at creation time of the missing file parts the new uuid is not present

FabianMauz:
Another problem seems to be the link between attachments and the entry inside the researchplan body (it models images in the main part of the RP). The link (UUID) there is not replaced with the new one.

FabianMauz (approved):

PiTrem (approved):

PR #1680, opened 2024-01-09
fix: temperature conversion

JanCBrammer:
Prior to this bug fix, if  `fromUnit` and `toUnit` are identical, `convertTemperature` throws:

```
TypeError
conversionTable[fromUnit][toUnit] is not a function
```


FabianMauz (approved):
looks good

PR #1702, opened 2024-01-25
feat: activate segment visibility field in generic elements

StarmanMartin:
Update user profile: 

Added Layout details for generic elements

Added ElementDetailSortTab to the generic elements. This allows to sort and hide segment tabs in generic elements.

During development some bugs were fixed that were not part of this problem:

1) The rfb lib (in the CnC component) is now loaded after the document is ready https://github.com/ComPlat/chemotion_ELN/pull/1896

2) The 'class' property was changed to className in several elements

3) The popover for hiding and sorting the segments was placed correctly under its button


phuang26:
@PiTrem could you please help to review the CnC part, thank you.

phuang26:
the CnC change has been moved to https://github.com/ComPlat/chemotion_ELN/pull/1896

PR #1711, opened 2024-01-30
fix: data cannot be removed from segment of element

phuang26:


cllde8 (approved):

cllde8:
https://github.com/ComPlat/chemotion_ELN/pull/1711

PR #1771, opened 2024-02-20
fix: devcontainer

soerenPeters:
With https://github.com/ComPlat/chemotion_ELN/pull/1665 the node installation is moved from `Dockerfile.chemotion-dev` to `run-ruby-dev.sh`, which is not executed by the devcontainer. That's why the devcontainer would not start. Running `run-ruby-dev.sh` as part of the devcontainer startup would introduce a lifetime depdendency from the rails server (running in the `app` container) to the devcontainer. This is why we extracted the installation of node into a new script, which is now called from `prepare-ruby-dev.sh` as well as `postCreateCommand` command in devcontainer.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -53,11 +53,11 @@
   ],
   "postCreateCommand": [
     "/bin/bash",
-    ".devcontainer/post_create.sh"
+    "prepare-ruby-dev.sh"
   ],
   "remoteUser": "chemotion-dev", // see https: //aka.ms/vscode-remote/containers/non-root
   "shutdownAction": "stopCompose", // stop compose when quitting
-  "overrideCommand": true, // override the commands in the compose file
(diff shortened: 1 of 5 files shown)


JanCBrammer (approved):
I've re-build my devcontainer from scratch and it works fine. Everything is built correctly, and once the containers are up, I can (manually, as intended) start the rails server.

However, we incorrectly assumed that we can cleanly start `webpacker` by re-trying (in conjunction with having it depend on `app` startup). Turns out, 5 re-tries aren't sufficient because `bundle install` in `app` takes too long when building `app` from scratch.
I've fixed this by having `webpacker` wait until all gems are installed in `app`. See 83a2086e3cef22de84657a621d3062ffcd68a8e2.

FabianMauz:
I also rebuild the container with the new script and it worked. I had no trouble with 5 retry times approach but the later one seems to be more convinient.

FabianMauz (approved):

soerenPeters:
Jan, I checked it again with your additions in devcontainer and with docker compose and both are still working as expected. Also agree that It's now much cleaner with the health check.

Just wondering why the tests are failing...

JanCBrammer:
I believe the failing tests are fixed here: https://github.com/ComPlat/chemotion_ELN/pull/1772.

PR #1830, opened 2024-03-25
refactor: svg scrubber

PiTrem:
- adjust svg scrubber to handle ketcher newer format with replacing css rgb function to black hex value


FabianMauz (approved):
Had a look together with Pierre and discussed some open questions.

PR #1838, opened 2024-03-26
refactor: quill to plain text call-backs 

PiTrem:
convert quill to plain text in background job to speed up saving containers or reactions

### DIFF
--- a/app/models/container.rb
+++ b/app/models/container.rb
@@ -28,7 +28,8 @@ class Container < ApplicationRecord
   belongs_to :containable, polymorphic: true, optional: true
   has_many :attachments, as: :attachable
 
-  before_save :content_to_plain_text
+  around_save :content_to_plain_text,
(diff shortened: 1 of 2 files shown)


PiTrem:
```rb

# frozen_string_literal: true

require 'benchmark'

time = Benchmark.measure do
  r = Reaction.create(created_by: 1, description: { 'ops' => [] }, observation: { 'ops' => [] })
  # description = r.description
  # observation = r.observation
  100.times do
    r.description['ops'] << { 'attributes' => { 'script' => 'sub' }, 'insert' => '2' }
    r.observation['ops'] << { 'insert' => '2' }
    r.save!
  end
  # r.description = description
  # r.observation = observation
  # r.save!
  r.really_destroy!
end

puts "Time elapsed: #{time.real} seconds"

```

save between 50% to 90% time

Issue #1847, opened 2024-04-04
alignment of buttons in element lists


nicolejung:
In element lists, the buttons for shared collections and reactions (symbol) are not aligned anymore. Would be great if someone could have a look

PR #1853, opened 2024-04-08
Fix: sample entry alignment and spacing

mekkyz:
Fixes: https://github.com/ComPlat/chemotion_ELN/issues/1847

I showed all the buttons and symbols that I saw for the aligment (see picture)

Some quick and dirty techniques were used because most of the components use css classes that are not universal and serve specific needs. Changing css would break other things

Before: 
Screenshot from 2024-04-08 10-17-03

After:
Screenshot from 2024-04-08 10-13-48

### DIFF
--- a/app/packs/src/apps/mydb/elements/list/ElementsTableSampleEntries.js
+++ b/app/packs/src/apps/mydb/elements/list/ElementsTableSampleEntries.js
@@ -364,18 +364,20 @@ export default class ElementsTableSampleEntries extends Component {
             onClick={showDetails.bind(this, sample.id)}
           >
             {sample.title(selected)}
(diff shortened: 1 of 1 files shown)


nicolejung:
Hi Mostafa, thanks, looks very good - the only feedback that I have: there is a white button without content - could you check what this is used for?

mekkyz:
> Hi Mostafa, thanks, looks very good - the only feedback that I have: there is a white button without content - could you check what this is used for?

Hi @nicolejung , this is a generic element label. The reason why it is white, that I have no access on the dev server on the the generic element things. 
However I added the button as a placeholder to align it correctly.

When there is an actual generic element label, it will show correctly.

PR #1860, opened 2024-04-10
refactor: remove unused part as this is obsolet due to .tool-versions…

soerenPeters:


JanCBrammer (approved):

JanCBrammer:
@PiTrem, ok if we merge this?

soerenPeters:
With .tools-version the required version is installed using asdf. However, it is still necessary to add the corresponding plugin before running `asdf install` with `asdf plugin add XXX`.

Issue #1878, opened 2024-04-18
Copy cell lines


FabianMauz:
In analogy to the samples, a copy function for elements of the type cell lines is to be created. The functionality is needed for further developments regarding to generic elements and the ability to link cell lines to them.

- [x] Backend REST-API Endpoint
- [x] UseCase for Copy functionality
- [x] FrontEnd - Fetcher method to get easy access to function
- [x] UI Integration - Really neccessary?
- [ ] Documentation

PR #1923, opened 2024-05-16
1878 copy cell lines

soerenPeters:
This PR solves https://github.com/ComPlat/chemotion_ELN/issues/1878


app/packs/src/stores/alt/actions/ElementActions.js
```diff
@@ -485,6 +485,17 @@ class ElementActions {
     return cellLineSample;
   }
 
+  copyCellLineFromId(id,collectionId ) {
+    return (dispatch) => {
+      CellLinesFetcher.copyCellLine(id,collectionId)
+        .then((result) => {
+          dispatch(result);
+        }).catch((errorMessage) => {
+          console.log(errorMessage);
```
FabianMauz, line 494: I think we should add a response to the user that something went wrong

soerenPeters: Is this an error that can occur normally or should it not actually fail? In the latter case, I think that logging is enough. Especially since all other operations in the `ElementAction` only log to the console in the case of an error. What do you think?

PR #1937, opened 2024-05-24
fix: limit welcome email attempts

PiTrem:
- set max attempt to 2 to avoid unnecessary sus traffic is the email undeliverable.

- disable call-back for dev seed users


FabianMauz (commented):
Somehow it seems that the override of max_attempts is not working:

1. I changed the perform method to provoke an error in the job:
image
2. Created a new user
3. Watched at delayed_job process and observed database
image
image


Maybe something related to https://github.com/collectiveidea/delayed_job/issues/715

PiTrem:
:weary: 

Indeed seems  not to work as dreamed of but I found another solution...

FabianMauz (commented):
Tested again in development mode. Delayed job was removed from database after first run failed. 
TO DO: set up a simple mail server and test if email was sent

FabianMauz (approved):
Tested the positive and negative case, all working fine now.

PR #1942, opened 2024-05-24
chore: Remove `Chrome` from devcontainer

JanCBrammer:
`Chrome` was used to drive our end-to-end tests. Since we no longer run end-to-end tests (for now) this PR removes the `Chrome` installation to reduce bloat and speed up build time. Confirmed with @PiTrem.


soerenPeters (approved):
Nice!

Issue #1875, opened 2024-04-15
Reaction Variations Refactoring


JanCBrammer:
During previous feature development I collected a few refactorings.
I don't want to pollute feature PRs with these, that's why I'm bundling them in a separate PR.

## ReactionVariations.js

- [x] update material-related state the way analyses-related state is updated, i.e., get rid of `reactionMaterialsChanged`
- [x] extract and test `copyVariationsRow` and `updateVariationsRow`
- [x] extract material-related components and functions into `ReactionVariationsMaterials.js` (akin to `ReactionVariationsAnalyses.js`)

## ReactionVariationsUtils.js

- [x] inline short functions that are used in one place only (if it's not critical to test them in isolation)

PR #1946, opened 2024-05-27
Reaction variations refactoring

JanCBrammer:
This PR addresses #1875.


FabianMauz (changes requested):
We have gone through the commits -> seems to be ok.

At live testing if found some issues:

1. Changing a value after sorting in a column changes the wrong row: ~~
20240529_reaction-variations-1

2. I think when in full screen mode all of the space should be used for the table

3. The import of collections (exported a synced collection -> imported it again) did not import the variations of reactions. Checked it -> also not working on uncharted branch. I think that should be a seperate issue.

FabianMauz (approved):
Talk with @JanCBrammer . 
1) The table is resorted after change. -> not an issue at all
2) Will be dressed in a separate issue
3) Will be dressed in a separate issue

Issue #1948, opened 2024-05-27
Reaction Variations: Allow comma and period as decimal separators


JanCBrammer:
Currently, the table displays `NaN` on entering a comma as decimal separator.

Auto-convert comma to period? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

PR #1959, opened 2024-06-03
feat: Allow comma and period as decimal separators in reaction variations table

JanCBrammer:
This PR addresses #1948.

The reaction variations table accepts only a period (i.e., ".") as decimal separator for numeric entries. The goal is to allow a comma and period as decimal separators. In the current codebase, there are at least two ways to achieve this.

## 1. Using the JavaScript package `numeral`

The package is used in the `NumeralInput` component (which in turn is used in "Elemental composition" section of `SampleDetails` component via `ElementalCompositionCustom` and `ElementalCompositionGroup`components). I decided not to use `numeral`:

- The package is dead (last commit 7 years ago: https://github.com/adamwdraper/Numeral-js).
- We're limited to an outdated version (of the dead package).

## 2. Custom logic in the `handleInputValueChange` method of the `NumeralInputWithUnitsCompo` component

I decided against using the custom logic:
- It's not re-usable outside of the component and its not trivial to extract since it's tightly coupled to the component.
- It's not applicable to my needs since it parses characters as they are entered rather than the entire string on submission.
- It breaks when using multiple commas or periods (e.g., as thousands separator).
		
Since the JavaScript standard library doesn't offer locale-specific number parsing / parsing different decimal separators (see https://stackoverflow.com/a/42000120 and https://stackoverflow.com/questions/55364947/is-there-any-javascript-standard-api-to-parse-to-number-according-to-locale), I'm adding a dependency-free solution that could ideally be re-used to replace solutions 1. and 2. (in a separate PR).

### DIFF
--- a/app/packs/src/utilities/MathUtils.js
+++ b/app/packs/src/utilities/MathUtils.js
@@ -60,8 +60,6 @@ function parseNumericString(numberString) {
   if (typeof numberString !== 'string') {
     return NaN;
   }
-
-  const numberIsNegative = numberString.startsWith('-');
   let sanitizedNumberString = numberString;
 
   // Remove all characters that aren't digits, commas, or periods.
@@ -81,7 +79,7 @@ function parseNumericString(numberString) {
       sanitizedNumberString.slice(finalPeriodIndex + 1)}`;
   }
 
-  if (numberIsNegative) {
+  if (numberString.startsWith('-')) {
     sanitizedNumberString = `-${sanitizedNumberString}`;
   }
 
--- a/app/packs/src/utilities/MathUtils.js
+++ b/app/packs/src/utilities/MathUtils.js
@@ -56,6 +56,17 @@ const formatBytes = (bytes, decimals = 2) => {
   return `${parseFloat((bytes / (k ** i)).toFixed(dm))} ${sizes[i]}`;
 };
 
+/**
+ * Parse a string into a number.
+ *
+ * All characters other than digits, commas, and periods are ignored,
+ * with the exception of an optional leading dash to indicate a negative number.
+ * The string may contain a decimal separator, which can be either a comma or a period.
+ * All other periods or commas (such as thousands separators) are ignored.
+ *
+ * @param {string} numberString - The string to parse.
+ * @returns {number|NaN} - The parsed number or NaN if parsing fails.
+ */
 function parseNumericString(numberString) {
   if (typeof numberString !== 'string') {
     return NaN;
--- a/app/packs/src/utilities/MathUtils.js
+++ b/app/packs/src/utilities/MathUtils.js
@@ -56,9 +56,42 @@ const formatBytes = (bytes, decimals = 2) => {
   return `${parseFloat((bytes / (k ** i)).toFixed(dm))} ${sizes[i]}`;
 };
 
+function parseNumericString(numberString) {
+  if (typeof numberString !== 'string') {
+    return NaN;
+  }
+
+  const numberIsNegative = numberString.startsWith('-');
+  let sanitizedNumberString = numberString;
+
+  // Remove all characters that aren't digits, commas, or periods.
+  sanitizedNumberString = sanitizedNumberString.replace(/[^0-9,.]/g, '');
+  if (sanitizedNumberString === '') {
+    return NaN;
+  }
+
+  // Decimal separator can be comma or period. Convert to period.
+  sanitizedNumberString = sanitizedNumberString.replaceAll(',', '.');
+  // Keep only final (non-terminal) period under the assumption that it's meant as the decimal separator.
+  // Assume that preceding periods were meant as thousands separators.
+  const finalPeriodIndex = sanitizedNumberString.lastIndexOf('.');
+  if (finalPeriodIndex !== -1) {
+    sanitizedNumberString = `${sanitizedNumberString.slice(0, finalPeriodIndex).replaceAll('.', '')
+    }.${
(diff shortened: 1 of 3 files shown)


app/packs/src/utilities/MathUtils.js
```diff
@@ -56,9 +56,42 @@ const formatBytes = (bytes, decimals = 2) => {
   return `${parseFloat((bytes / (k ** i)).toFixed(dm))} ${sizes[i]}`;
 };
 
+function parseNumericString(numberString) {
```
soerenPeters, line 59: As this a potential useful function in many contexts, It would be nice to see how this function behave without reading the tests or looking into the implementation. fo example with some JSDoc documentation?

app/packs/src/utilities/MathUtils.js
```diff
@@ -56,9 +56,42 @@ const formatBytes = (bytes, decimals = 2) => {
...
+  if (typeof numberString !== 'string') {
+    return NaN;
+  }
```
soerenPeters, line 62: It could be confusing, that you pass in a valid number (which is not a string), but the function returns `NaN`. 
This a discussion about expectation and could be resolved having a doc-string. 

But, what if every type is accepted and cast to a `string`? Alternatively, if a number is passed, this number could return immediately. 

I'm not sure what the best way to handle this is either and your implementation might be the best JS solution. This is just my thoughts on this.

app/packs/src/utilities/MathUtils.js
```diff
@@ -56,9 +56,42 @@ const formatBytes = (bytes, decimals = 2) => {
...
+
+  const numberIsNegative = numberString.startsWith('-');
```
soerenPeters, line 64: Move this variable next to its usage.

app/packs/src/utilities/MathUtils.js
```diff
@@ -56,9 +56,42 @@ const formatBytes = (bytes, decimals = 2) => {
...
+  const finalPeriodIndex = sanitizedNumberString.lastIndexOf('.');
+  if (finalPeriodIndex !== -1) {
+    sanitizedNumberString = `${sanitizedNumberString.slice(0, finalPeriodIndex).replaceAll('.', '')
+    }.${
+      sanitizedNumberString.slice(finalPeriodIndex + 1)}`;
+  }
+
+  if (numberIsNegative) {
+    sanitizedNumberString = `-${sanitizedNumberString}`;
+  }
```
soerenPeters, line 86: I would split this block into three sub-functions:
Advantage: you can perhaps get rid of the comments and it is clear in the later refactoring that essentially three steps are applied: 
1. remove all non-numeric characters
2. replace commas with periods and declare only the last period as a decimal point.
3. set the sign

Its a matter of taste, you might argue that you want to have this behavior as close as possible together. I am also fine with that in this case.

app/packs/src/utilities/MathUtils.js
JanCBrammer: Good point, see 59e2d74.

app/packs/src/utilities/MathUtils.js
JanCBrammer: I feel like the function name indicates that it expects a string? 
Together with the type signature form 59e2d74, that's even more clear.
I'd find it more confusing if it *would* accept an argument of type `number`.

app/packs/src/utilities/MathUtils.js
JanCBrammer: Good point, see d20fc28.

app/packs/src/utilities/MathUtils.js
JanCBrammer: I see your point. In other cases, I'd probably agree, however, in this case I'd prefer not to fragment the function because:

- I don't see much utility / re-usability in the sub-functions. At least at the moment. I'd defer extraction until I (or someone else) wants to re-use the logic partially.
- It's a matter of taste, but to my eyes, fragmenting an already short(ish) function hurts readability rather than helps it.
Even if the sub-functions are well-named, I'd still read them (which requires jumping around) just to make sure they're doing what I think they do.
- I'm not interested in testing any of the sub-functions. Atm I only care for their aggregate behavior. I.e., I'm fine treating the function as a black box. In fact, I *want* it to be a black box (hiding complexity).

app/packs/src/utilities/MathUtils.js
soerenPeters: Yes, you are right - i agree. Actually, I was not sure about this either. So lets keep this!

app/packs/src/utilities/MathUtils.js
soerenPeters: Sorry for being annoying. Isn't the doc-string usually above the function signature?

app/packs/src/utilities/MathUtils.js
soerenPeters: As i said in my first comment, i agree, it's a matter of taste. From my point of view, the readability and understandability would increase, because I can read through the algorithm of this function, without having to care about the details of the implementation. It would be only "private" helper functions, so it would be still fine to just test the main function as a blackbox. 

 In this specific case, the algorithm is quite clear, so I don't have problems with keeping it as it is.

app/packs/src/utilities/MathUtils.js
JanCBrammer: Good catch, I didn't know the convention. Fixed with 0665a60.

soerenPeters (approved):

PR #1971, opened 2024-06-11
feat: split cell lines

FabianMauz:
This PR implements the feature to create **sub cell lines** of a cell line. The inheritance is implemented by the [ancestry](https://github.com/stefankroes/ancestry) gem as already done in samples. 

The PR also includes the **copy** functionality which had an own PR but will be closed in favour of this one.


spec/api/chemotion/cell_line_api_spec.rb
```diff
@@ -249,5 +249,122 @@
...
+
+  describe 'POST /api/v1/cell_lines/copy' do
+    let(:collection) { create(:collection, label: 'other collection') }
+    let!(:user) { create(:user, collections: [collection]) }
+    let!(:cell_line) { create(:cellline_sample, collections: [collection]) }
+    let(:allow_creation) { true }
+    let(:container_param)  do
+      { 'name' => 'new',
+        'children' =>
+[{ 'name' => 'new',
```
JanCBrammer, line 261: Indentation seems off?

spec/usecases/cell_lines/copy_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      it 'cell line sample label was created' do
+        expect(loaded_cell_line_sample_copied.short_label).not_to be cell_line_sample_to_copy.short_label
+      end
+
+      it 'user cell line amount was changed' do
+        loaded_cell_line_sample_copied
+        old_value = user.counters['celllines']
+        expect(user.reload.counters['celllines']).not_to be old_value
```
JanCBrammer, line 50: Could be more specific by using `.to be >` instead of `.not_to be`?

spec/usecases/cell_lines/copy_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+# rubocop:disable RSpec/MultipleExpectations
+
+require 'spec_helper'
+
+RSpec.describe Usecases::CellLines::Copy do
```
JanCBrammer, line 7: Would it make sense to add tests for `Usecases::CellLines::Split` as well?

JanCBrammer (commented):
Finished first round of review. Will test UI next.

JanCBrammer:
Are there plans to document the copy and split features over at https://www.chemotion.net/docs/eln/ui/elements/cell_lines?

JanCBrammer (commented):
I confirm that the "Copy Cell line" and "Split Cell line" buttons work analogous to the corresponding buttons for the other elements.

However, the naming of copied cell line samples is different from other elements.
For samples and reactions, the copies get new, (seemingly) unique names.
That is not the case for cell line samples.
For the latter, irrespective of the name of the original (i.e., the copied cell line sample), the copy is called "C".

Another (potential) naming issue is that the cell line sample names aren't unique.
When I create two cell lines samples (within the same cell line), both are called "C0".

PR #1973, opened 2024-06-14
fix: attachment permission validation to allow access on shared items

FabianMauz:
This PR will fix a problem if the user tries to **download** an **attachment** from e.g. a research plan if the plan is in a **synchronized** or shared **collection**.

### DIFF
--- a/app/api/chemotion/attachment_api.rb
+++ b/app/api/chemotion/attachment_api.rb
@@ -103,11 +103,19 @@ module Chemotion
                           ElementPermissionProxy.new(current_user, element, user_ids).read_dataset?
             end
           elsif @attachment
+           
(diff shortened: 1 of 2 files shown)


JanCBrammer (changes requested):
Ideally, we'd have a single test that demonstrates the issue (pre-PR) and verifies the fix (post-PR).

JanCBrammer (approved):
@FabianMauz walked me through the problem in the UI (I cannot link to the original issue since it's on a private board). I could reproduce the issue on the current `main` branch (HEAD at 8c235edcee9968663196221fd3e7a8468bef742d) and confirm that this PR fixes the issue. The same goes for the tests: they fail on the current `main` branch, and pass in `download-attachment-from-sharedsync-research-plan`.

Issue #1928, opened 2024-05-21
Possibility of image selection for preview in datasets


FabianMauz:
It should be possible for users to choose which image should be used for the preview thumbnails in analyses.

E.g.

Analysis has 2 datasets

dataset-1 
  -- Cat-1.jpg
  -- Cat-2.jpg
dataset-2
 --  Cat-3.jpg

![image](https://github.com/ComPlat/chemotion_ELN/assets/23172219/2aff6dbd-57bc-473c-a4dd-53fb55a3ec97)

The current behavior is taken the last image from the first dataset. 
Maybe the user can also set if he likes to see the original or the annotated image if one exists.

Issue #1927, opened 2024-05-21
Image of dataset preview does not match magnified image


FabianMauz:
In the analysis area of samples and/or reactions there is an thumbnail of the most recent image of files in the dataset.
The thumbnail is based on the annotated image. If i click at the thumbnail a magnified version of the image appears but it is the original image without the annotation.

It should always be the same image base, either annotated or original. 

This issue is a sequel issue from https://github.com/ComPlat/chemotion_ELN/issues/1900 .

![image](https://github.com/ComPlat/chemotion_ELN/assets/23172219/b208081e-dd2d-4567-a806-1caac76256a1)

After clicking on the thumbnail image:

![image](https://github.com/ComPlat/chemotion_ELN/assets/23172219/13f6c43c-5140-4094-8eb4-8db1831376ad)

PR #1978, opened 2024-06-18
fix: dataset thumbnail image to match magnified preview when annotated

FabianMauz:
This PR fixes the problem that in an container image preview the thumbnail image does not match the magnified image if the user clicks on it if there is an annotated version of the image (see #1927).

It also fixes that for cell lines the magnify function was not enabled.

As discussed with the reviewer the fix does not fully solve the problem. After annotating the image in the container the preview image is instantly updated but the magnified image still needs a reload of the element. It would need a relative big refactoring of the logic to achieve an instant updating. Since there is an issue that requires this modificationanyway, I would like to plead in favor of solving this remaining problem in issue https://github.com/ComPlat/chemotion_ELN/issues/1928

### DIFF
--- a/app/packs/src/apps/mydb/elements/details/cellLines/analysesTab/EditModeHeader.js
+++ b/app/packs/src/apps/mydb/elements/details/cellLines/analysesTab/EditModeHeader.js
@@ -115,8 +115,10 @@ export default class EditModeHeader extends React.Component {
 
     return (
       <div className="preview">
+        
         <ImageModal
-          hasPop={false}
+          hasPop={true}
+          disableClick={false}
           previewObject={{
             src: previewImg
           }}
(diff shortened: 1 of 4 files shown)


JanCBrammer (approved):
I could reproduce issue #1927 on the current main branch (HEAD at 8c235edcee9968663196221fd3e7a8468bef742d).
I can confirm that this PR fixes the issue.
I agree with @FabianMauz's assessment that the current fix is "good enough" for now. It's solving most of the problem with a small change.

FabianMauz:
After some discussion with @PiTrem we decide to also check, if the problem of magnifying the preview in direct attachments is also related to the same problem and if it easy to solve.

FabianMauz:
The fix above solves partially the issue with attachment annotated previews. The preview will only be correct if the side was reloaded. Again a rework would be needed to solve it completely. This will be done in https://github.com/ComPlat/chemotion_ELN/issues/1928

PR #1981, opened 2024-06-21
chore: rm deused chemscanner and unused  js dep

PiTrem:
- rm broken implementation of chemscanner pending availability of newer, standalone, version

- clean unused js dep


Gemfile
```diff
@@ -18,7 +18,7 @@ gem 'bootstrap-sass'
 gem 'charlock_holmes'
 
 # gem 'chem_scanner', git: 'git@git.scc.kit.edu:ComPlat/chem_scanner.git'
```
JanCBrammer, line 20: Why comment those out instead of removing them?

JanCBrammer (commented):
TODO for me:

- [x] Verify that all dependencies that were removed from `package.json` aren't imported in `.js` files that aren't related to `chemscanner`.
  - there seems to be a left-over reference to `raw-loader`: https://github.com/ComPlat/chemotion_ELN/blob/c29054ae248b7f5c8ce73fa91ee8276c01e3435a/config/webpack/custom.js#L34
  - `chem-spectra-client` and `chem-spectra-editor` seem to depend on `react-redux` (see `yarn.lock`)
- [x] Verify that all dependencies that were imported exclusively in `chemscanner`-related `.js` files were removed in `package.json`.

PR #1999, opened 2024-07-03
Fix generic designer

beque:
Fix navigation of generic designer


JoachimTillessen (changes requested):
@beque This looks good. I have two suggestions:

1. the class .vw-90 should be renamed to .w-90vw to better align with bootstrap sizing class names (see https://getbootstrap.com/docs/5.3/utilities/sizing/)
2. I would align the button with another positioning method, float usually has the most unwanted side effects (I will update our handbook accordingly).

PR #2002, opened 2024-07-05
Refactor sample task inbox to newer bootstrap

jhmegorei:


app/packs/src/components/sampleTaskInbox/SampleTaskCard.js
```diff
@@ -155,35 +154,39 @@ const SampleTaskCard = ({ sampleTask }) => {
...
-        <div className="row">
-          <div className="col-sm-6">
-            {contentForSample(sampleTask, dropRef)}
-          </div>
-          <div className="col-sm-6">
-            {contentForSampleTask(sampleTask)}
-          </div>
+      </Card.Header>
+      <Card.Body className="d-flex">
+        <div className="w-50">
```
JoachimTillessen, line 176: You can use the Col component directly in a d-flex container.
If you need extra DOM-elements for layout purpose I would suggest using the dedicated components (Container, Row, Col).
Apply the corresponding utility classes, when they reduce nesting depth.
Col components (and the simple .col class) automatically take up equal parts of their parent container – no "w-50" class needed.
https://react-bootstrap.netlify.app/docs/layout/grid

app/packs/src/components/sampleTaskInbox/SampleTaskInbox.js
```diff
@@ -77,62 +77,69 @@ const SampleTaskInbox = ({}) => {
...
-                size="sm"
-                onClick={sampleTasksStore.hideSampleTaskInbox}
-              >
-                <i className="fa fa-times" />
-              </Button>
-            </div>
-          </div>
-        </Panel.Heading>
-        <Panel.Body>
-          <div className="row sampleTaskCreationDropzones">
```
JoachimTillessen, line 123: I notice a lot of class names, that do not correspond to css rules.
They are especially weird because they are in camel-case.
I would remove them where ever possible.

JoachimTillessen (changes requested):
Hi Johannes,
hier ein paar Empfehlungen für knotenfreies Schnürsenkeln.

app/packs/src/components/sampleTaskInbox/SampleTaskCard.js
jhmegorei: The result is pretty ugly markup-wise, but okay 🙈

app/packs/src/components/sampleTaskInbox/SampleTaskInbox.js
jhmegorei: The classes are for code navigation. otherwise I have no proper way of knowing which Component renders which part of the generated html. (I had a horrible experience using the React DevTools, so this is my preferred way at the moment)

JoachimTillessen: Yeah I like finding my stuff in the DOM inspector too (it's way faster). Let's take that topic to our next team meeting. I think it's important, that we establish best practices that are not only useful for the original authors but also transparent for others.

app/packs/src/components/sampleTaskInbox/SampleTaskCard.js
```diff
@@ -172,13 +172,17 @@ const SampleTaskCard = ({ sampleTask }) => {
...
       </Card.Header>
-      <Card.Body className="d-flex">
-        <div className="w-50">
-          {contentForSample(sampleTask, dropRef)}
-        </div>
-        <div className="w-50">
-          {contentForSampleTask(sampleTask)}
-        </div>
+      <Card.Body>
+        <Container>
```
JoachimTillessen, line 176: You do not have to have a container in order to use row. No extra nesting needed.

jhmegorei: Interestingly enough though you are right in this case, the Footer of the modal (SampleTaskInbox.js) needs a container around the row, otherwise the Dropzones don't use the full width of the row... no idea why...
So I have two different places where I want to achieve the same thing (2 content blobs that should span a full row next to each other) with different behaviour. That being the way, I would either understand why this happens or have at least consistent markup :D

JoachimTillessen: CSS is awesome

JoachimTillessen: If you want them side by side, why does the footer have an flex-column class?

jhmegorei: Some remains from experimenting with the utility classes if I remember correctly.
I updated the corresponding codepaths, so it looks a bit better now :)

JoachimTillessen (approved):

PR #2005, opened 2024-07-08
Update bootstrap: Collection Views

maiwald:


app/packs/src/apps/mydb/collections/CollectionSubtree.js
```diff
@@ -205,48 +162,51 @@ export default class CollectionSubtree extends React.Component {
...
+          id={`tree-id-${root.label}`}
+          className={`title ${selected ? 'selected' : ''} d-flex align-items-baseline`}
+          onClick={this.handleClick}
+        >
+          {showGatePushButton && (<GatePushBtn collection_id={root.id} />)}
+          <span className="me-auto">{label}</span>
+          {this.canTakeOwnership() && (
+            <i
+              className="fa fa-exchange"
+              onClick={e => this.handleTakeOwnership(e)}
```
jhmegorei, line 184: handleTakeOwnership takes no parameters, so you could just use
```javascript
onClick={this.handleTakeOwnership}
```

app/packs/src/apps/mydb/collections/CollectionSubtree.js
```diff
@@ -205,48 +162,51 @@ export default class CollectionSubtree extends React.Component {
...
-          {this.expandButton()}
-          {this.synchronizedIcon()}
-          {gated}
-          {label}
+        <div
+          id={`tree-id-${root.label}`}
+          className={`title ${selected ? 'selected' : ''} d-flex align-items-baseline`}
+          onClick={this.handleClick}
+        >
+          {showGatePushButton && (<GatePushBtn collection_id={root.id} />)}
```
jhmegorei, line 179: I am not too fond of this `{conditional && (renderSomething)}` pattern.
My preferred solution is creating a function
```javascript
const gatePushButton = () => {
  showButton = some_conditional
  if (!showButton) return null;

  return (<GatePushBtn ...>)
}
```

This makes the render method way easier to read.

What's your opinion? (and maybe the whole team's opinion)

app/packs/src/apps/mydb/collections/CollectionTree.js
```diff
@@ -263,26 +246,32 @@ export default class CollectionTree extends React.Component {
...
-
     return (
-      <div className='collection-tree'>
+      <div className="collection-tree">
         <div className="tree-view">
-          {this.collectionManagementButton()}
+          <div className="take-ownership-btn">
+            <Button id="collection-management-button" size="xsm" variant="danger"
+              title="Manage & organize collections: create or delete collections, adjust sharing options, adjust the visibility of tabs based on the collection level"
+              onClick={() => this.handleCollectionManagementToggle()}>
```
jhmegorei, line 255:
```javascript
onClick={this.handleCollectionManagementToggle}
```

jhmegorei (commented):
just some minor comments, please respond as you see fit

app/packs/src/apps/mydb/collections/CollectionSubtree.js
JoachimTillessen: onClick={this.handleTakeOwnership} is better

app/packs/src/apps/mydb/collections/CollectionSubtree.js
JoachimTillessen: I have very strong feelings against having a four lines of code in a remote location somewhere in the class. What I found in a lot was that people added additional functionality and that tiny function driftet further and further away from the render method obscuring what the contained DOM actually consists of. Having the complete DOM/component tree within the render method is a nice separation of concern, this is where devs look if they want to change the markup. If you have a complex decision tree that determines what to render case a, b c,..., than yes it is likely bloating the render function. But if it's a simple condition then the one liner is way more readable to me.
The whole return null pattern is especially cumbersome to me, it emphasizes what not to do. I'm no react veteran and maybe my preferences are influenced by vue.js standards, but I'm pretty sure I haven't seen it anywhere outside of chemotion.

JoachimTillessen: We have already gone to great lengths to eliminate this kind of code elsewhere reducing the overall size of classes significantly.

app/packs/src/apps/mydb/collections/SyncWithMeCollections.js
```diff
@@ -11,7 +11,7 @@ export default class SyncWithMeCollections extends React.Component {
       tree: {
         label: 'Synchronized with me Collections',
         id: -1,
-        children: [{}],
+        children: [],
```
JoachimTillessen: What do we do with code duplication like this. Shall we open a refactoring backlog?

JoachimTillessen (changes requested):

app/assets/stylesheets/collection_management.scss
```diff
@@ -40,86 +52,13 @@
...
-
   .caret-right:before {
-    content: '\25B8';
+    content: "\25B8";
     font-size: 15px;
   }
 
   .caret-down:before {
-    content: '\25BE';
+    content: "\25BE";
```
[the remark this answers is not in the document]
maiwald: I'm leaving this as is for now. There is a way to use fa, but these classes are used in different parts of the app and changing this would increase the scope of this PR. I'll do it in a separate commit/PR for all Tree views.

Issue #2000, opened 2024-07-04
Using Reaction Variation leads to buggy display 


BenjaminGolub:
- I am a ELN admin

# Expected Behavior

The detail view of the reaction has a blue bar (on top). This bar should be dark blue in case very changes are stored. It switches to light blue in case some unsaved changes are in this tab. So if I just view a reaction, without changing anything, the top bar should always stay dark blue. 

# Current Behavior

When I add the "variation" tab to the reaction detail view, then the dark blue bar on top, will constantly change to light blue, even if nothing has been changed (see video below). As soon as I hide the "variation" tab again, the blue bar stays dark blue as expected. When working with "variation" this is confusing and slightly annoying. 

## Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Create a reaction and add enable the "variation" segment
2. Store the reaction and open another element in the ELN (another reaction or sample)
3. Open up the reaction again and the dark blue bar on top will change to light blue
4. If you disable "variation" again and do the same thing, the blue bar on top will stay dark blue

## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

 - OS: Windows 11
 - Chrome Browser 
 - ELN version: v1.9.3

## Failure Logs / Screenshots

https://github.com/ComPlat/chemotion_ELN/assets/107389350/d589cf3c-7cea-4954-9f4a-fc47f18f7e7b

JanCBrammer:
Thanks for the issue @BenjaminGolub! I'll look into it.

JanCBrammer:
@BenjaminGolub, I could reproduce and fix the issue. We'll let you know when the fix is released.

PR #2007, opened 2024-07-09
fix: Show save-status of reaction variations table

JanCBrammer:
Addresses issue #2000.

TLDR of the issue: a UI element shows the save-status of the reaction variations table by turning from dark-blue to light-blue when there are unsaved changes. Prior to this PR, the UI element was always light-blue, even when all changes in the reaction variations table had been changed.

Commits 3cc75ffb6c32401479a6e647724ab5f972308aad and 5731839b1af8908740316c2767d3b7cd0b5845ab fix the issue.
The remaining commits are related refactorings.


FabianMauz (approved):
Looked together with @JanCBrammer at code, made some small changes and tested locally. Everting works as intended.

PR #2024, opened 2024-07-17
Update home to new bootstrap version

maiwald:


app/packs/src/apps/home/Home.js
```diff
@@ -1,29 +1,20 @@
...
+    <Container fluid>
+      <Row className="card-navigation">
+        <Navigation isHidden />
+      </Row>
+      <Row>
+        <Col>
+          <WelcomeMessage />
+        </Col>
+      </Row>
+    </Container>
```
JoachimTillessen, line 18: Is this grid useful here? I don't see any cols...

JoachimTillessen (changes requested):
Thank you!!

PR #2018, opened 2024-07-12
feat: Printing configuration: customizable code type, layout, label options, and molecule display for PDF output

Nicolass67:
Issue : 
https://github.com/ComPlat/chemotion_ELN/issues/1910

Previous printing code feature : 
- Possibility to print two type of pdf : Small Size or Large Size 
- Qr code, bar, code and data matrix are all displayed in the pdf

Changes in the printing code feature : 
- Printing configuration is set up with a config file (public/json/printingConfig/defaultConfig.json)
- Possibility to choose the type of code printed (QR code, Barcode or Data matrix Code)
- Possibility to change the width of the printed pdf (size in mm)
- Possibility to choose which label print in the printed pdf (name/short label/external label/code log/molecule name)
- Possibility to choose the text position (below the code or on the right side of the code)
- Possibility to print the sample onto the printed pdf 

Config File : 
{
    "code_type": String [qr_code bar_code data_matrix_code] - Default is qr_code
    "code_image_size": Integer - Min value : 0 Max value = 100.
    "name": boolean - Display or not the name of the sample/molecule
    "short_label": boolean - Display or not the name of the sample
    "external_label": boolean - Display or not the name of the sample
    "code_log": boolean - Display or not the name of the sample/molecule
    "molecule_name": boolean - Display or not the name of the sample
    "displaySample": boolean - Display or not the Sample image
    "text_position" : String [below right] Default is "below". Fixed to "below" if code_image_size>30 or code_type="bar_code"
    "width": Integer
}

Example of Config File : 
{
    "code_type": "qr_code",
    "code_image_size": 30,
    "name": true,
    "short_label": true,
    "external_label": true,
    "code_log": true,
    "molecule_name": true,
    "displaySample": true,
    "text_position" : "below",
    "width": 100
}

https://github.com/user-attachments/assets/ef8815b8-f336-4c48-bb85-4a96bd362ab5

PR #2053, opened 2024-08-02
feat: customization of qr/bar/datamatrix-code print label

Nicolass67:
Changes in the printing code feature : 
- Printing configuration is set up with a config file
- Possibility to choose the type of code printed
- Possibility to change the width of the printed pdf
- Possibility to choose which label print in the printed pdf
- Possibility to print the sample onto the printed pdf


PiTrem:
duplicated #2018

PR #2059, opened 2024-08-07
Update Bootstrap: CellLine Details

maiwald:


app/packs/src/apps/mydb/elements/details/cellLines/analysesTab/EditModeRow.js
```diff
@@ -1,42 +1,39 @@
...
-            readOnly={readOnly}
-            disabled={false}
-            container={container}
-            onChange={() => parent.handleChange(container)}
-          />
-        </Panel.Body>
-      </Panel>
-    );
-  }
+export default function EditModeRow({
```
beque, line 7: If you change this to a functional component.
Prefer to write it like this:
`export default EditModeRow  = ({ container, parent, element, readOnly }) => {`
or you can write it like this:
`const  default EditModeRow  = ({ container, parent, element, readOnly }) => {`
and the `export default EditModeRow` at the end of the file

beque (commented):
Some minor fixes

PR #2068, opened 2024-08-12
feat: change history tracker with UI for auditing record changes

PiTrem:
Add a change history tracking with gem logidze.

Store a history which keeps track of every change made to certain
entities and add a "History" tab to the Reaction/Sample detail pages.
Related entities are displayed in the version history. E.g. Reaction ->
ReactionsSample.
Save timestamp and author of the version with the changes. 
A change diff to the previous version can be displayed.
Instead of one version per database, update versions are grouped by the
request id (since some requests can lead to multiple updates for the same record). 
These changes are merged to make it more intuitive for the user.

Reverting a record to previous state: 
the user can update selected fields of a record to a former values.
Note that this is not a roll-back and the applied reverting changes are tracked
as well.

Tracked entities:
- Sample
  - Residue
  - ElementalComposition
- Reaction
- ReactionsSample
- ResearchPlan
  - ResearchPlansWellplate
  - ResearchPlanMetadata
- Screen
- Wellplate
  - Well
- Container
- Attachment

(replace #1253)

-----

## Backend
- implement Logidze 
  -- catch whodunit and wrap it into each DB update/insert transaction 
- api for fetching/reverting history
- fetchers, reverters, and serializers in app/service/versioning

## DB
- add json column to store the history for each tracked model 
- build specific `gem-fx` versioned sql logging functions
- added triggers on update insert for tracked elements
 a `gem-fx` versioned function is used to generate the default trigger for each model
 instead of using `gem-fx` versioned individual triggers since all triggers are identical
- NB migration squashed and timestamped forward
- ⚠️ migration initiate history object for all records of the tracked models  
- migration rollback confirmed:
NB: back-rolling the logging function also roll-back all triggers since they depends on it
- added soft deletion on container
- some update on older migration
 
##  Frontend
- history accessible in an additional tab in each element-detail card
- specific svg formatter and json
- history tab syncs with properties - reloads on click
- change-table using AgGridReact

---------

Co-authored-by: Martin Schneider <m.schneider@cortona.de>
Co-authored-by: VadimKeller <v.keller@cortona.de>
Co-authored-by: Martin Schneider <mschneider85@users.noreply.github.com>
Co-authored-by: VadimKeller <VadimKeller@users.noreply.github.com>
Co-authored-by: Adrian Herrmann <adrian.herrmann@kit.edu>


public/jsmol
JanCBrammer, line 1: I believe this shouldn't be checked in either?

app/packs/src/stores/alt/actions/ElementActions.js
JanCBrammer, line 1: As far as I can tell there aren't any semantic changes in this file.
Could we revert the changes to make the PR smaller?

app/packs/src/apps/mydb/elements/details/VersionsTable.js
```diff
@@ -0,0 +1,208 @@
...
+import VersionsTableTime from 'src/apps/mydb/elements/details/VersionsTableTime';
+import VersionsTableChanges from 'src/apps/mydb/elements/details/VersionsTableChanges';
+import { elementShowOrNew } from 'src/utilities/routesUtils';
+import DetailActions from 'src/stores/alt/actions/DetailActions';
+import SamplesFetcher from 'src/fetchers/SamplesFetcher';
+import ResearchPlansFetcher from 'src/fetchers/ResearchPlansFetcher';
+import ScreensFetcher from 'src/fetchers/ScreensFetcher';
+import WellplatesFetcher from 'src/fetchers/WellplatesFetcher';
+
+export default class VersionsTable extends Component {
```
JanCBrammer, line 16: Maybe convert this to a functional component? I feel like introducing new components might be an opportunity to move the codebase away from class components.

app/packs/src/apps/mydb/elements/details/VersionsTable.js
```diff
@@ -0,0 +1,208 @@
...
+              background: white;
+            }
+          `}
+        </style>
+        <ul className="history-legend">
+          <li className="history-legend__item history-legend__item--old">before</li>
+          <li className="history-legend__item history-legend__item--new">after</li>
+          <li className="history-legend__item history-legend__item--current">current value</li>
+        </ul>
+        <BootstrapTable
```
JanCBrammer, line 187: Maybe use `ag-grid` instead? We already have it in our dependencies.  Even better would be to use a plain `Table` from `react-bootstrap`.

JanCBrammer (changes requested):
I made a few suggestions to revert the changes to semantically unchanged files.
This would make the PR a bit smaller which would make working with the PR in the GitHub UI a bit easier. Atm, the UI is quite unresponsive due to to its size (large diffs in many files).

I'll continue the review over the coming days.

app/packs/src/models/ResearchPlan.js
JanCBrammer, line 1: Are those changes related to the versioning feature? If not, please revert.

app/packs/src/models/Sample.js
JanCBrammer, line 1: Are there any semantic / behavioral changes in this diff? If not, please revert.

JanCBrammer (changes requested):

app/models/sample.rb
```diff
@@ -59,6 +59,7 @@
 class Sample < ApplicationRecord
   attr_accessor :skip_inventory_label_update
 
+  has_logidze
```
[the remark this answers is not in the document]
headri: yes

app/packs/src/apps/mydb/elements/details/VersionsTableChanges.js
```diff
@@ -0,0 +1,77 @@
...
+  const {
+    id, changes, handleRevert, isEdited
+  } = props;
+
+  const revertibleFields = () => {
+    if (isEdited) return -1;
+
+    let filteredFields = [];
+
+    changes.forEach(({ fields }) => {
```
JanCBrammer, line 17: Use `map` instead of `forEach`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

app/packs/src/apps/mydb/elements/details/VersionsTableChanges.js
```diff
@@ -0,0 +1,77 @@
...
+        ))}
+      </ol>
+      <VersionsTableFields fields={fields} renderRevertView={false} />
+    </React.Fragment>
+  ));
+
+  const isRevertible = revertibleFields();
+  let alertText = 'You cannot undo changes. ';
+  if (isRevertible === 0) {
+    alertText += 'Every change is either up to date or irreversible';
```
JanCBrammer, line 49: Now the first and last conditional branch both contain the "all changes are irreversible" case.

JanCBrammer (changes requested):
This feature needs tests :monocle_face:

JanCBrammer (approved):
Tests look ok. Nice that there's at least some basic coverage now.

JanCBrammer (commented):
I tested the latest implementation in the UI. I deleted a starting material from a reaction and saved the reactions. That resulted in two observations that puzzle me a bit:

1. I'd expect the deleted starting material to show up in the "previous version" column, and be absent in the "updated version" column. Not the other way around (like on the screenshot):
Screenshot from 2024-11-28 09-12-09


2. Although I have saved the reaction, a warning indicates that I cannot undo the change due to unsaved changes:
Screenshot from 2024-11-28 09-09-59

JanCBrammer (commented):
@headri and I looked into the the two issues I posted earlier:

1. Seems to be an issue with the UI, not the internal logic. The "previous version column" *does* contain data, but it's only visible on hover-over or when enlarging the column.

2. Happens because the reaction is incorrectly marked as change / unsaved. This problem does not seem to have been introduced with this PR though.

PR #2079, opened 2024-08-15
Update Bootstrap: Metadata

maiwald:


app/packs/src/components/metadata/MetadataGeneral.js
```diff
@@ -1,103 +1,104 @@
...
-      <ControlLabel>
+    </Form.Group>
+    <Form.Group className="mb-3">
+      <Form.Label>
         Description
-      </ControlLabel>
-      <FormControl
+      </Form.Label>
+      <Form.Control
         componentClass="textarea"
```
ashwiniHerle, line 50: I think the latest version uses `as="textarea"` instead of componentClass

PR #2094, opened 2024-08-21
update-bootstrap-sample-analyses-tab

JoachimTillessen:


app/packs/src/apps/mydb/elements/details/samples/analysesTab/SampleDetailsContainersCom.js
```diff
@@ -161,83 +53,118 @@ function RndEdit({
...
+                    container={container}
+                    mode={mode}
+                    handleUndo={handleUndo}
+                    readOnly={readOnly}
+                    isDisabled={isDisabled}
+                    handleRemove={handleRemove}
+                    handleSubmit={handleSubmit}
+                    toggleAddToReport={toggleAddToReport}
+                  />
+                </Accordion.Header>
```
beque, line 125: When you are using an `Accordion.Header` and put in the `AnalysesHeader` with all the content and buttons, it will throw a html error.
That's happened because the Accordion.Header generates an H2-Tag with the class 'accordion-header' and inside is a Button. In this Button is the hole content of the `AnalysesHeader`.

I have added a new accordion header at the screen analyses:
https://github.com/ComPlat/chemotion_ELN/blob/update-bootstrap-ui-library/app/packs/src/apps/mydb/elements/details/screens/analysesTab/ScreenDetailsContainers.js

With this new header the hole header is gray and has no white padding. It looks like the analyses on the uncharted server and has a correct html.

Maybe you want to adapt this?

beque (commented):
Just a suggestion for the analyses header.

And you have added Lucianos changes of the static pages into your branch.
Maybe a problem with rebasing?

app/packs/src/apps/mydb/elements/details/samples/analysesTab/SampleDetailsContainersCom.js
JoachimTillessen: Sounds good. I will check it out!

PR #2118, opened 2024-08-30
Fixes for screen

beque:
Fix analyses tab
Fix print code button and modal (modal size / centered)


JoachimTillessen (approved):
Thanks for further cleaning up the print modal...

PR #2119, opened 2024-08-30
prevent scrollbars in nested cards, convert elements_details styles t…

JoachimTillessen:
Having cards inside Cards for example in Sample details resulted in each nested card displaying scrollbars.
Furthermore I migrated the corresponding css file to scss, as it is the standard format in this project.
@ashwiniHerle please have a look.


ashwiniHerle:
Looks good. Could you please confirm if scroll works normally after the changes? In sample details, reaction details etc?

ashwiniHerle:
Obsolete PR. Implemented else where.

PR #2123, opened 2024-09-03
feat(wellplates): allow resizing after creation when no filled well is lost

### LINKED ISSUE #3144: wellplate element behavior
The wellplate size can only be changed before the element wellplate is saved. That's ok, but in consequence, the user should see the field without dropdown option once the element. There should also be a hover over information that the size cannot be changed if this status is reached.

headri:
### What this adds

A wellplate can be created without a size — the size dropdown offers **"Define later"** — and sized afterwards, from either a predefined option or the custom-size dialog.

The size is no longer frozen once set. A wellplate can be resized at any time, subject to one rule:

> **Growing is always allowed. Shrinking is refused when a well outside the new grid still holds data.**

"Holds data" means a sample, a readout value or unit, a label, a colour code, or an additive — anything a user put there. Untouched wells never block a resize, so shrinking an empty region of a plate just works.

When a smaller size *would* discard filled wells, the ELN does not ask for confirmation and does not silently delete them. It refuses, and says which wells are in the way:

- sizes that would drop filled wells are greyed out in the dropdown;
- the custom-size dialog disables **Apply** and names the blocking positions.

Clearing those wells and resizing are therefore two separate, separately-saved steps: empty the wells, save, then resize. The same order applies when growing to make room for new samples — save the larger size first, then place samples in the new wells.

### Why not keep "the size can only be set once"

That rule only ever existed in the browser, as a disabled dropdown. The API still accepted new dimensions, and nothing reconciled the `wells` table against the new grid, so a second resize — from a stale second tab, or any direct API call — left the plate holding wells outside its own grid. That state destroyed the affected samples on the next save and broke the Designer tab on reload, with no way back through the UI.

Enforcing the rule properly required moving it to the server. Once it lives there, "resize only when nothing is lost" is both safer and less restrictive than "resize only once", so the stricter rule was dropped in favour of it.

### Behaviour worth knowing

- **Resizing is its own save.** Choosing a size on a saved wellplate applies immediately and persists on its own; it is not part of the next Save. The size control is therefore locked while the wellplate has unsaved changes, with a tooltip explaining why — a resize replaces the wellplate with the reconciled one from the server, which would otherwise discard those edits. A wellplate that has not been created yet resizes in memory, as before.
- **Size changes cannot be undone from the version history.** Reverting dimensions would bypass the same guard, so `width`/`height` are shown in the history but are not revertible.
- **Importing a spreadsheet into an unsized wellplate is refused** with a message asking for a size first. Previously the position check validated nothing at 0×0 and created wells at positions no grid could hold.

### API changes

- **New:** `PUT /api/v1/wellplates/resize/:id` with `width` and `height` (each `0..100`). Validates the request, refuses a shrink that would drop a filled well (`422` naming the wells), then reconciles the well rows and returns the updated wellplate.
- **`PUT /api/v1/wellplates/:id` no longer accepts `width`/`height`.** They are ignored, as they were before this branch. This is deliberate: a client echoing cached dimensions must not be able to revert someone else's resize.
- `POST /api/v1/wellplates` dimensions are bounded at `100`, matching `WellPosition` and the xlsx template generator. `101` was previously accepted and produced a wellplate whose template endpoint returned a 500.

### Tests

New coverage for the resize use case and endpoint (including every way a well can count as "filled", the two-tab conflict, and out-of-grid repair), the `content?` predicate, the spreadsheet-import guard, and the first spec for a versioning reverter in the repo. On the frontend, coverage for the resize semantics, the pending-changes lock, the greyed-out sizes, and the custom-size validation boundaries — restoring the `describe.skip`ped boundary cases that had gone stale.

### DIFF
--- a/spec/javascripts/packs/src/apps/mydb/elements/details/wellplates/propertiesTab/WellplateSizeDropdown.spec.js
+++ b/spec/javascripts/packs/src/apps/mydb/elements/details/wellplates/propertiesTab/WellplateSizeDropdown.spec.js
@@ -116,4 +116,18 @@ describe('WellplateSizeDropdown (functional component)', () => {
     expect(matching.length).toEqual(1);
     expect(selectedOption(wrapper).props().label).toEqual('96 (12x8)');
   });
+
+  it('shows a plain read-only field with an explanatory tooltip once the size is locked', () => {
+    const wellplate = new Wellplate({ ...wellplate8x12EmptyJson, is_new: false });
+    const wrapper = mount(
+      <WellplateSizeDropdown wellplate={wellplate} updateWellplate={emptyFunction} />
+    );
+
+    expect(wrapper.find('select').length).toEqual(0);
+    expect(wrapper.find('input').first().props().value).toEqual('96 (12x8)');
+    expect(wrapper.find('input').first().props().disabled).toEqual(true);
+    expect(wrapper.find('button.create-own-size-button').props().disabled).toEqual(true);
+    const overlay = wrapper.find('OverlayTrigger').props().overlay;
+    expect(shallow(overlay).text()).toEqual('The size cannot be changed.');
+  });
 });
--- a/spec/javascripts/packs/src/apps/mydb/elements/details/wellplates/propertiesTab/WellplateSizeDropdown.spec.js
+++ b/spec/javascripts/packs/src/apps/mydb/elements/details/wellplates/propertiesTab/WellplateSizeDropdown.spec.js
@@ -2,7 +2,7 @@
 
 import React from 'react';
 import expect from 'expect';
-import { configure, shallow } from 'enzyme';
+import { configure, shallow, mount } from 'enzyme';
 import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
 import WellplateSizeDropdown from 'src/apps/mydb/elements/details/wellplates/propertiesTab/WellplateSizeDropdown';
 import wellplate2x3EmptyJson from 'fixture/wellplates/wellplate_2_3_empty';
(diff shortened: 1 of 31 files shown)


JanCBrammer (commented):
> After the size is set, it cannot be changed again.

Wasn't this what this PR was supposed to implement? I.e., wasn't the goal to enable changing wellplate-size even after creation?

app/packs/src/apps/mydb/elements/details/wellplates/propertiesTab/WellplateSizeDropdown.js
```diff
@@ -11,15 +11,24 @@ class Option {
   }
 
   get label() {
-    return `${this.height * this.width} (${this.width}x${this.height})`;
+    const size = this.height * this.width;
+    if (size === 0) {
+      return 'not yet chosen';
```
JanCBrammer, line 16: Could we change this to "Define later"?

spec/javascripts/packs/src/apps/mydb/elements/details/wellplates/propertiesTab/WellplateSizeDropdown.spec.js
JanCBrammer, line 1: Maybe add test case(s) for "not yet defined" option?

JanCBrammer (commented):
I believe the instructions in red are off by one? I.e., The red instructions should go away when I enter "0" or "100" ? 

wellplate_dimensions

Maybe also replace 0 with 1, since 0 doesn't make a lot of sense from a user perspective.

Other than that the changes seem to be working well in the UI. However, I'm not quite sure if this PR solves the original problem, i.e., to enable changing wellplate size at any time.

JanCBrammer (approved):
I confirm that the UI works as intended now.

JanCBrammer (approved):
The feature still works as specified.

PiTrem (approved):

PiTrem:
@headri @JanCBrammer — a heads-up before this lands, because the feature's contract changed since you last looked at it.

**What changed.** The original design enforced "the size can only be set once". That rule lived only in the browser, as a disabled dropdown — the API still accepted new dimensions, and nothing reconciled the `wells` table against the new grid. A second resize (a stale tab, or any direct API call) left the plate holding wells outside its own grid, which destroyed the affected samples on the next save and broke the Designer tab on reload.

Rather than harden that rule, it has been **replaced** by one that can actually be enforced server-side:

> Growing is always allowed. Shrinking is refused when a well outside the new grid still holds data.

So resizing is now permitted at any time, and "Define later" is no longer a one-shot. Clearing wells and resizing became two separately-saved steps, which is what makes the invariant checkable. The updated description covers the behaviour and the API changes in full.

**Six commits were added** on top of `2d9a6cf47e`: the rework itself, then fixes for issues found reviewing it — a crash on wellplates shared at a low detail level, a concurrent-resize race, a client/server disagreement about what counts as an occupied well, and an unscoped `Well.find` that let a payload update a well belonging to a different wellplate.

**Both approvals on this PR predate all of that.** @JanCBrammer's is on `b5230ddd99` and approves the *opposite* behaviour; mine is on `04732797f1` and predates the last three commits. GitHub doesn't dismiss stale approvals, so the PR reads as reviewed when the current head has not been. Flagging it explicitly rather than letting the green check speak for itself.

CI is green on `517a92989c`, and the full Ruby and JS suites pass locally. Happy to revisit any of it after the fact.

PR #2175, opened 2024-09-24
Update Bootstrap: Remove `clipboard` package

maiwald:


JoachimTillessen (approved):
Yes! We can replace this dependency with a super simple utility.

PR #2177, opened 2024-09-24
fix popover dimensions display popover header

JoachimTillessen:

- [ ] added code is linted


app/packs/src/components/common/SvgWithPopover.js
```diff
@@ -16,20 +16,22 @@ export default class SvgWithPopover extends Component {
       <Popover
         id="popover-trigger-hover-focus"
         title={popObject.title}
```
maiwald, line 18: I think these two properties can be removed. `Popover` does not accept a `title` and the id is not used anywhere.

JoachimTillessen: Good catch! This is where the title, that had been displayed in the original version, has gone missing. Also static IDs for components are generally not a good idea.

PR #2196, opened 2024-10-04
refactor: Remove UNSAFE_ react lifecycle hooks

maiwald:


app/packs/src/apps/mydb/elements/list/ElementsTableSampleEntries.js
```diff
@@ -189,34 +189,40 @@ export default class ElementsTableSampleEntries extends Component {
...
+  componentDidUpdate(prevProps) {
+    const { elements, moleculeSort } = this.props;
+    if (elements === prevProps.elements && moleculeSort === prevProps.moleculeSort) {
+      return
+    }
+
     const moleculelist = {};
-    elements.forEach((sample) => {
+    const moluculeList = elements.forEach((sample) => {
       let samples = [];
```
jhmegorei, line 210: this line is no longer necessary

app/packs/src/apps/mydb/elements/list/ElementsTableSampleEntries.js
```diff
@@ -189,34 +189,40 @@ export default class ElementsTableSampleEntries extends Component {
...
+
+  componentDidUpdate(prevProps) {
+    const { elements, moleculeSort } = this.props;
+    if (elements === prevProps.elements && moleculeSort === prevProps.moleculeSort) {
+      return
+    }
+
     const moleculelist = {};
-    elements.forEach((sample) => {
+    const moluculeList = elements.forEach((sample) => {
```
jhmegorei, line 209: why do you assign another variable that is never used?

app/packs/src/apps/mydb/elements/list/ElementsTableSampleEntries.js
```diff
@@ -189,34 +189,40 @@ export default class ElementsTableSampleEntries extends Component {
...
-    Object.keys(moleculelist).forEach((moleculeId, idx) => {
-      displayedMoleculeGroup.push(moleculelist[moleculeId]);
-      let numSamples = moleculelist[moleculeId].length;
-      if (nextProps.moleculeSort && numSamples > 3) { numSamples = 3; }
-      displayedMoleculeGroup[idx].numSamples = numSamples;
+
+    const displayedMoleculeGroup = Object.keys(moleculelist).map((molId) => {
+      const m = moleculelist[molId];
+      if (moleculeSort && m.length > 3) {
+        m.numSamples = 3;
```
jhmegorei, line 220: This logic isn't quite the same. In the old code, the numSamples property of the array is always set, in your new code, it is only set when there are more than 3 samples within. There is no default, which would break code down the line that relies on numSamples being a valid integer.

app/packs/src/components/QuillEditor.js
```diff
@@ -71,31 +71,14 @@ export default class QuillEditor extends React.Component {
...
-      if (sel) this.editor.setSelection(sel);
-    }
-  }
-
-  // eslint-disable-next-line camelcase
-  UNSAFE_componentWillUpdate() {
-    this.componentWillUnmount();
-  }
-
-  componentDidUpdate() {
```
jhmegorei, line 93: Can this be safely deleted?

jhmegorei (changes requested):

app/packs/src/apps/mydb/elements/details/researchPlans/ResearchPlanMetadata.js
```diff
@@ -1,85 +1,77 @@
...
-        url: '',
-        landing_page: '',
-        alternate_identifier: [],
-        related_identifier: [],
-        description: [],
-        dates: [],
-        geo_location: [],
-        funding_reference: []
-      }
+      researchPlanMetadata: defaultsDeep({}, researchPlan.research_plan_metadata, emptyMetadata),
```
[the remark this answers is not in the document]
maiwald: I don't understand what you mean regarding the rename. The metadata state is taken from the passed in `researchPlan` prop instead of additionally passing it in as a prop. Do the names not match anywhere?

I'll check regarding lodash. I did not investigate what the server returns for metadata. So maybe we can get rid of setting defaults entirely.

app/packs/src/apps/mydb/elements/details/screens/researchPlansTab/EmbeddedResearchPlanDetails.js
```diff
@@ -56,13 +56,15 @@ export default class EmbeddedResearchPlanDetails extends Component {
...
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    let { researchPlan, expanded } = nextProps;
-    if (!(researchPlan instanceof ResearchPlan)) {
-      const rResearchPlan = new ResearchPlan(researchPlan);
-      researchPlan = rResearchPlan;
+  componentDidUpdate(prevProps) {
+    let { researchPlan } = this.props;
+    if (researchPlan !== prevProps.researchPlan) {
+      if (!researchPlan instanceof ResearchPlan) {
```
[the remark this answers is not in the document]
maiwald: Good catch! I was not aware of this.

app/packs/src/apps/mydb/elements/list/ElementsTableSampleEntries.js
maiwald: That does look odd. I will check the code again.

app/packs/src/apps/mydb/elements/details/researchPlans/ResearchPlanMetadata.js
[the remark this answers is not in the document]
jhmegorei: ok you are partly right. I misread this line here. But still, you can remove the property researchPlanMetadata from the proptype definition in line 504 :)

app/packs/src/components/QuillEditor.js
maiwald: I should be. Although it seems that we don't properly clean up the editor. I will check this again.

maiwald: I checked the code. Before this change `componentDidUpdate` would call `initQuill` that is guarded by `if (!this.editor) …` which is set during `componentDidMount`. So `componentDidUpdate` would always hit the guard and skip the contents of `initQuill`.

jhmegorei (approved):

PR #2197, opened 2024-10-05
fix: development environment setup during the docker build process

Ramy-Badr-Ahmed:
This PR implements several improvements to the Chemotion ELN development environment for a dynamic build-up process:

- Added database preparation script `prepare_database.sh`:
     - Renames the exisiting configuration files (e.g., `database.yml`, `shrine.yml`, `storage.yml`, `datacollectors.sh`), it prevents failures in the `rake db:setup` process.
     - Checks existing pending migrations via `db:migrate:status` return for `down`.

- Added database migration step in `prepare-ruby-dev.sh` before starting the Rails server. This prevents `ActiveRecord::PendingMigrationError` due to pending migrations on server startup.

- Changes to `Dockerfile.chemotion-dev`:
  - Added dependency: `jq`. 
  - Set `GEM` in `PATH`. 
  - Added `bundle` & `gem` installation before `prepare-rubygems.sh`.
  - Created `node_module` directory as `root` then changed ownership. 
  - Restyled `Dockerfile.chemotion-dev` file.

- Shell Script Line Ending Management:
     - Added a `.gitattributes` file to enforce `LF` line endings for shell scripts and .tool-versions (prevent issues with `CRLF` line endings across different OS).
    - Updated `docker-compose.dev.yml` to ensure shell scripts are checked and converted to `LF` line endings before execution as expected by Docker.
- Changed `Webpacker` service to reference `.env.development` instead of `.env`
- Synchronise dynamically `engines.node` in package.json with `REQUIRED_NODE_VERSION` to avoid `yarn` failure during `js` package installation.


Ramy-Badr-Ahmed (commented):
In my forked version:

[Continuous Integration](https://github.com/Ramy-Badr-Ahmed/chemotion_ELN/actions/runs/11193794643/job/31119526823)

PR #2210, opened 2024-10-10
fix in bootstrap: color and label not saved on submit in wellplates

ashwiniHerle:


beque (approved):
Looks good

PR #2212, opened 2024-10-11
Fix position of select options div

beque:
- Fix position of select options
- Fix delete rule for style sheets at quill editor
- Fix breaking molecule name

### DIFF
--- a/app/packs/src/components/reactQuill/ToolbarDropdown.js
+++ b/app/packs/src/components/reactQuill/ToolbarDropdown.js
@@ -52,7 +52,7 @@ export default class ToolbarDropdown extends React.Component {
 
   componentWillUnmount() {
     try {
-      this.sheet.deleteRule(this.ruleId, this.sheet.cssRules.length - 1);
+      this.sheet.deleteRule(this.ruleId - 1);
     } catch (err) {
       console.log(err);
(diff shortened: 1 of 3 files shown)


app/packs/src/components/reactQuill/ToolbarDropdown.js
```diff
@@ -50,7 +50,7 @@ export default class ToolbarDropdown extends React.Component {
     });
   }
 
-  UNSAFE_componentWillMount() {
+  componentWillUnmount() {
     try {
       this.sheet.deleteRule(this.ruleId, this.sheet.cssRules.length - 1);
```
maiwald, line 55: `deleteRule` does not take a second argument:
https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/deleteRule

maiwald (commented):
I did not verify the css changes because I could not reproduce the problem locally. The rest looks good, apart from a small note. 👍

app/packs/src/components/reactQuill/ToolbarDropdown.js
beque: I have found an other error at the deleteRule and fixed it now

PR #2215, opened 2024-10-15
fix: selecting all pages with advanced/detail search corrected

headri:
this.state.currentSearchSelection is not set in detail / advanced search
therefore condition removed; the other condition is sufficient
closes https://github.com/ComPlat/chemotion/issues/217

### DIFF
--- a/app/packs/src/stores/alt/stores/UIStore.js
+++ b/app/packs/src/stores/alt/stores/UIStore.js
@@ -1,4 +1,5 @@
-import { List, fromJS } from 'immutable';
+/* eslint-disable no-param-reassign */
+import { List } from 'immutable';
 import alt from 'src/stores/alt/alt';
 
 import UIActions from 'src/stores/alt/actions/UIActions';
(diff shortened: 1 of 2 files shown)


app/packs/src/stores/alt/stores/UIStore.js
```diff
@@ -188,31 +188,29 @@ class UIStore {
   handleCheckAllElements(params) {
     this.waitFor(ElementStore.dispatchToken);
 
-    let { type, range } = params;
-    let { elements } = ElementStore.getState();
+    const { type, range } = params;
+    const { elements } = ElementStore.getState();
 
-    if (range == 'all') {
-      if (this.state.currentSearchSelection && elements[type + "s"].ids) {
```
headri, line 195: relevant change

JanCBrammer: Just to be sure, the only behavioral change in this file's diff is the removal of `this.state.currentSearchSelection` from the guard clause? For the future please change behavior and formatting in separate commits.

JanCBrammer (changes requested):
If this is a bug fix, please add a test that replicates (or at least approximates) the bug and demonstrates how this PR fixes it.

PiTrem:
to be ported to main with test

PR #2218, opened 2024-10-15
fix: quill editor and other warnings

beque:
- Remove add and delete css at quill editor
- Remove required at ContainerDatasetField
- Remove pullRight and add margin for icons at SpectraEditorButton
- Remove refs at selects of quill editor custom toolbar
- Set key for quill editor custom toolbar

### DIFF
--- a/app/packs/src/components/common/Select.js
+++ b/app/packs/src/components/common/Select.js
@@ -1,24 +1,27 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 import RSelect from 'react-select';
 import RAsyncSelect from 'react-select/async';
 import RCreatableSelect from 'react-select/creatable';
 
 /* eslint-disable react/jsx-props-no-spreading */
 
- // deactivate the default styling and apply custom class names to enable bootstrap styling
- // see https://react-select.com/styles#the-unstyled-prop
- // see https://react-select.com/styles#the-classnameprefix-prop
+// deactivate the default styling and apply custom class names to enable bootstrap styling
+// see https://react-select.com/styles#the-unstyled-prop
+// see https://react-select.com/styles#the-classnameprefix-prop
 
-export const Select = ({className, ...props}) => (
-  <RSelect
-    className={["chemotion-select", className].join(' ')}
-    classNamePrefix="chemotion-select"
-    unstyled
-    {...props}
-  />
-);
+export const Select = forwardRef(function Select(props, ref) {
+  return (
+    <RSelect
+      className={["chemotion-select", props.className].join(' ')}
+      classNamePrefix="chemotion-select"
+      ref={ref}
+      unstyled
+      {...props}
+    />
+  );
+});
 
-export const AsyncSelect = ({className, ...props}) => (
+export const AsyncSelect = ({ className, ...props }) => (
   <RAsyncSelect
     className={["chemotion-select", className].join(' ')}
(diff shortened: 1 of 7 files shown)


app/packs/src/components/reactQuill/ToolbarDropdown.js
```diff
@@ -30,14 +30,6 @@ export default class ToolbarDropdown extends React.Component {
   }
 
   componentDidMount() {
-    const cssRule = ".ql-snow .ql-picker." + this.ddQuillId + " .ql-picker-label::before, ." + this.ddQuillId + " .ql-picker.ql-size .ql-picker-item::before { content: attr(data-label); line-height: 25px; }";
```
maiwald, line 33: 🥳 

Can we now get rid of `this.ddQuillId` as well?

app/packs/src/components/textTemplateToolbar/ToolbarTemplateCreator.js
```diff
@@ -172,7 +171,6 @@ export default class ToolbarTemplateCreator extends React.Component {
             />
             <Select
               className="me-2 col-10 f-5"
-              ref={selectRef.ref}
```
maiwald, line 175: I think we need those refs. They are used to pull the values out of the components when saving. I have thought about rewriting the whole state management in this component but decided against it, because it seemed like too much work.

maiwald (changes requested):

app/packs/src/components/textTemplateToolbar/ToolbarTemplateCreator.js
beque: I removed this refs because they throw warnings
`Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`
But I haven't tried to save the toolbar.
This doesn't work anymore (with or without the refs)

So I will look for an other solution now

app/packs/src/components/reactQuill/ToolbarDropdown.js
beque: Yes, I' ll try to remove it

maiwald (changes requested):

app/packs/src/components/common/Select.js
```diff
@@ -1,24 +1,27 @@
...
 
-export const Select = ({className, ...props}) => (
-  <RSelect
-    className={["chemotion-select", className].join(' ')}
-    classNamePrefix="chemotion-select"
-    unstyled
-    {...props}
-  />
-);
+export const Select = forwardRef(function Select(props, ref) {
```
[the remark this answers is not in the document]
beque: I was thinking about this too.
But this is the only case where ref is used

app/packs/src/components/textTemplateToolbar/ToolbarTemplateCreator.js
```diff
@@ -116,7 +116,7 @@ export default class ToolbarTemplateCreator extends React.Component {
     const { updateTextTemplates } = this.props;
     if (!updateTextTemplates) return;
 
-    const iconTemplates = this.toolbarSelectRef.current.state.value;
+    const iconTemplates = this.toolbarSelectRef.current.state.selectValue;
```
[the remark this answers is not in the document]
beque: I have fixed this

maiwald (approved):

PR #2236, opened 2024-11-08
refactor: migrate to shakapacker

jhmegorei:
upgrade from retired [rails/webpacker](https://github.com/rails/webpacker?tab=readme-ov-file#webpacker-has-been-retired-) to [shakacode/shakapacker ](https://github.com/shakacode/shakapacker)

!! Move JS file hierarchy to location defined in shakapacker.yml (app/javascript)


config/initializers/eln_features.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-# pack version
-pack_path = Webpacker.manifest.send(:data)&.fetch('application.js', nil)
-ENV['VERSION_ASSETS'] = pack_path && File.basename(pack_path)
+# TODO: clarify what this is and why it is used in the message api
```
PiTrem, line 3: somehow that should notify the user to reload the page/assets if their client app is running an older version

PiTrem (approved):
🎉

PR #2255, opened 2024-12-06
fix: creating molecules with invalid molfile using faulty smiles

adambasha0:
ensure the status of request to pubchem for getting mol from smiles is OK.
Otherwise return nil


adambasha0:
- This bug fix introduces a routine to do the following:

1. prevent the creation of faulty molecule (meaning invalid/corrupt molfile) using molecule :smiles API endpoint. 

Routine: use MolfileValidation to validate molfile which is fetched using PubchemService

PiTrem (approved):

PR #2266, opened 2024-12-11
ci: testing asset precompilation

PiTrem:
see


PiTrem:
replaced bz https://github.com/ComPlat/chemotion_ELN/pull/2268

PR #2274, opened 2024-12-20
User storage space

headri:
https://github.com/ComPlat/chemotion/issues/111


app/models/attachment.rb
```diff
@@ -297,13 +277,27 @@ def delete_file_and_thumbnail
...
+  end
+
   def attach_file
     return if file_path.nil?
     return unless File.exist?(file_path)
 
     attachment_attacher.attach(File.open(file_path, binmode: true))
     raise 'File to large' unless valid?
 
+    raise 'User quota exceeded' unless check_user_quota
```
JanCBrammer, line 299: Rename `check_user_quota` to `user_quota_exceeded?` (and invert return values). Then this line becomes `raise 'User quota exceeded' if user_quota_exceeded?`

db/functions/calculate_collection_space_v01.sql
JanCBrammer, line 1: Copilot tells me this could be refactored to

```
CREATE OR REPLACE function calculate_collection_space(collectionId integer)
returns bigint
language plpgsql
as $function$
declare
    used_space bigint default 0;
    element_types text[] := array['Sample', 'Reaction', 'Wellplate', 'Screen', 'ResearchPlan'];
    element_table text;
    element_space bigint;
begin
    foreach element_table in array element_types loop
        execute format('select sum(calculate_element_space(id, $1)) from collections_%s where collection_id = $2', lower(element_table))
        into element_space
        using element_table, collectionId;
        used_space := used_space + coalesce(element_space, 0);
    end loop;
    return coalesce(used_space, 0);
end;
$function$;
```

:robot:

JanCBrammer (commented):
Finished the first pass of my review.
Next I'll try the new features in the UI.

app/javascript/src/apps/admin/UserManagement.js
```diff
@@ -1461,22 +1492,26 @@ export default class UserManagement extends React.Component {
...
           {g.email}
         </td>
-        <td className="col-md-1 py-3">
+        <td className="py-3">
           {g.type}
         </td>
-        <td className="col-md-2 py-3">
+        <td className="py-3">
+          {g.available_space === 0 ? ''
+            : `${Math.round((g.used_space / g.available_space) * 100)}% of ${g.available_space / 1024 / 1024}MB`}
```
JanCBrammer, line 1509: Add space before "MB", otherwise the unit is difficult to see.
E.g., trying the feature in the UI I got `0% of 0.09999942779541016MB`.

app/javascript/src/apps/admin/AdminDashboard.js
```diff
@@ -28,30 +33,59 @@ export default class AdminDashboard extends React.Component {
...
+            defaultValue={`${diskPercentUsed}%` || ''}
+            readOnly
+          />
+          <InputGroup.Text>Default User Available Space (MB)</InputGroup.Text>
+          <InputGroup>
+            <Form.Control
+              type="number"
+              min="0"
+              defaultValue={availableUserSpace || ''}
+              onChange={(event) => this.setState({ availableUserSpace: event.target.value })}
```
JanCBrammer, line 78: Doesn't seem to accept floats that start with `0`?
When I enter `0.5` and hit "Save", the entry seems to fail silently. I.e., `0.5` isn't shown in "Disk Usage" column in "User Management" tab of the admin interface.

Also, I can only increase the value (e.g., from `1` to `2`). Decrements (e.g., from `2` back to `1`) don't seem to be accepted and fail silently.

app/javascript/src/components/container/ContainerDatasetModalContent.js
```diff
@@ -625,16 +629,21 @@ export class ContainerDatasetModalContent extends Component {
...
+                && renderGroup(attachmentGroups.Pending, 'Pending')}
+              {attachmentGroups.Original.length > 0 && renderGroup(attachmentGroups.Original, 'Original')}
+              {attachmentGroups.BagitZip.length > 0 && renderGroup(attachmentGroups.BagitZip, 'Bagit / Zip')}
+              {hasProcessedAttachments && Object.keys(attachmentGroups.Processed)
+                .map((groupName) => attachmentGroups.Processed[groupName].length > 0
+                  && renderGroup(attachmentGroups.Processed[groupName], `Processed: ${groupName}`, groupName))}
+              {attachmentGroups.Combined.length > 0 && renderGroup(attachmentGroups.Combined, 'Combined')}
+            </div>
+            <Alert variant="warning" show={UserStore.isUserQuotaExceeded(filteredAttachments)}>
+              Uploading attachments will fail; User quota will be exceeded.
```
JanCBrammer, line 644: Everywhere this message is shown, it would be good to inform the users how much space is available. Otherwise users have to find out with trial and error how big of a file they can upload?
`isUserQuotaExceeded` could give back the string including information regarding available space in case the quota is exceeded (e.g., `Uploading attachment(s) will fail; User quota (2 MB) will be exceeded.`).
In case the quota isn't exceeded `isUserQuotaExceeded` could give back an empty string.

JanCBrammer (commented):
Do we have documentation for this feature?

I tried the feature in the UI and have a few comments (see below).

app/models/attachment.rb
```diff
@@ -297,13 +277,27 @@ def delete_file_and_thumbnail
     attachment_attacher.destroy
   end
 
+  def user_quota_exceeded?
+    user = User.find(created_for.nil? ? created_by : created_for)
+    if (user.used_space + attachment_data['metadata']['size']) > user.available_space &&
+       !user.available_space.zero?
+      return false
```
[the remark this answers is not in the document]
headri: Correct, I change it.

app/javascript/src/apps/admin/AdminDashboard.js
headri: Decrements works for me. It only decreases the default value, users will not be decremented; lower value only counts for new users. Float Values are not intended.

JanCBrammer: So changes in the default value are only applied for new users, not existing users, got it.
If floats aren't intended, could the input be restricted to ints?

headri: Changes in the default value are applied for existing users, but only if higher than its current quota.

JanCBrammer: > "Could the input be restricted to ints": I don't know how.

Copilot suggests:

"To restrict the input of the `Form.Control` to integers, you can use the `step` attribute with a value of `1`. This ensures that the input only allows integer values. Here is the modified code:

```jsx
<Form.Control
  type="number"
  min="0"
  step="1" // This line ensures only integer values are allowed
  defaultValue={availableUserSpace || ''}
  onChange={(event) => this.setState({ availableUserSpace: event.target.value })}
/>
```
"

JanCBrammer (approved):

PR #2281, opened 2025-01-10
feat: element device description

beque:
Add a new  element manageable by users though the Element-list : Device Description 

it belongs to a device and defines:  
- the purpose of the device using ontology (CHMO)
- how the device is setup (device components description)
- current configuration
- responsible entities
...

image
image

Backend: 
- AR model DeviceDescription (belongs to Device)
- manageable through collection (has_many collections_device_descriptions, element_permission ...)
- Grape api endpoints and Usecases for CRUD operations
 (NB Labimotion::SegmentKlass dependence /dedicated sql queries for ontologies)
 - DB seed Labimotion::ElementKlass where name: 'device_description'
 
 Frontend:
 - integration for full management through element-list/element-detail UI
 - mobx store (element-details)
 - fetchers CRUD, split


PiTrem (commented):
LGTM 

Points for next step:
- migrations to be squashed (eg rename columns)
- check serialisation of attachment thumbnail
  performance: thumbnail not be send with attachment on fetching the parent Element,  but on separate request if the rendering is needed
- should the vanilla SQL for Labimotion:segment be placed somewhere else

PR #2286, opened 2025-01-15
fix: references typo

thvranken:
This fixes a typo in the Reference Manager.


JanCBrammer (approved):
Thanks for the fix @thvranken.

PR #2321, opened 2025-02-07
fix: device email editable

headri:


db/schema.rb
PiTrem, line 1: should not be there

PiTrem (changes requested):

PiTrem (approved):

PR #2342, opened 2025-02-19
fix: find_each order by on view notify_message

PiTrem:
introduced by rubocop formating

since: #2322

- NotifyMessage is a view with no primary key

- [find_each](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each) automatically `order by ` primary key

ref: #2322


headri:
Seems fixed, cannot reproduce any more.

headri (approved):

PR #2396, opened 2025-03-18
fix(ui): improve unit switch behavior for reaction scheme gase phase inputs

adambasha0:
Problem:
- Switching units for gas products input fields in the gaseous reaction scheme is not available with invalid values
- This was causing unnecessary UI clutter and potentially confusing users

Solution:
- Modified the `gaseousInputFields` function to only show tooltips when:
  1. The value is not defined (null, undefined, or 'n.d')
  2. The unit is not in the list of non-switchable units (ppm, TON): Some fields like 'ppm' and 'TON' should not show unit switch tooltips at all
- Added a clear message "Unit switch only active with valid values" to guide users
- Extracted the input component to avoid code duplication

Changes:
- Added `noSwitchUnits` array to identify units that don't support switching
- Updated tooltip display logic to check both value existence and unit type
___________________________________________________________________________

- [ ] added code is linted

### DIFF
--- a/app/javascript/src/apps/mydb/elements/details/reactions/schemeTab/Material.js
+++ b/app/javascript/src/apps/mydb/elements/details/reactions/schemeTab/Material.js
@@ -324,19 +324,36 @@ class Material extends Component {
     const readOnly = this.isFieldReadOnly(field);
 
     const updateValue = this.getFormattedValue(value);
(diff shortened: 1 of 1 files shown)


haditariq (approved):
Thank you. Field and tooltips are behaving as expected.

PR #2416, opened 2025-03-24
fix: Limit maximum SampleTaskInbox height

maiwald:
to keep draggable modal content in the viewport and prevent offscreen un-interactable content


alizaib1217 (approved):
The modal now stays within the viewport when more sample content is added, preventing overflow issue.

PR #2423, opened 2025-03-27
feat: track chemicals history

headri:


alizaib1217:
Issues Found During Testing

1. Revert Functionality Not Restoring Correctly
Steps to reproduce:
  1. Change vendor field to 12, then save.
  2. Change vendor to 123, then save again.
  3. Go to the revert section and revert the last change.
Expected: Vendor should revert to 12.
Actual: Vendor field becomes empty.

---------------

2. Crash on Status Dropdown Selection

Steps to reproduce:
  1. Click on the Status dropdown.
  2. Select any value.
Behavior: App crashes immediately.
Additional: I’ve attached a screenshot below for reference.
image

alizaib1217:
Hello.
In the ChemicalReverter, the handle_json method doesn’t handle parsing errors. If the JSON is malformed or not in the expected format, it could crash during revert. Shouldn't we handle this?

alizaib1217 (approved):
Looks good.

PiTrem (approved):

PR #2428, opened 2025-03-31
fix: import collection with faulty svg

PiTrem:
add check on svg file content

refactor: extract images from zip input to tmp a directory instead of keeping each one as Tempfile instances (avoid ulimit issue)


harivyasi:
@PiTrem I tested this on `next-release` server and can confirm that it works.

harivyasi (approved):

headri (approved):

PR #2431, opened 2025-04-04
chore(db): fix molecules and sample with faulty molfiles

PiTrem:
- pubchem info request: ensure success status to process the response 
- process to handle molecule with corrupted molfiles and associated
  samples:
  - try to regenerate according to sample molfile or molecule smiles
- add migration and rake task data:molefile:fix

refs: #2255

### DIFF
--- a/lib/tasks/support/molecule_structure_curation.rb
+++ b/lib/tasks/support/molecule_structure_curation.rb
@@ -1,0 +1,308 @@
+# frozen_string_literal: true
+
+require 'logger'
+
+# MoleculeStructureCurationError is a custom error class for molecule structure curation errors.
+class MoleculeStructureCurationError < StandardError
+  def initialize(message = nil)
+    super(message || 'Molecule structure curation error')
+  end
+end
+
+# MoleculeStructureCuration is responsible for finding and fixing problematic molecules
+# It will find all molecules that have a corrupted molfile due to wrong parsing of pubchem response
+# and try to create correct molecules anew and update associated samples.
+# @example: MoleculeStructureCuration.new.process
+
+class MoleculeStructureCuration
+  LOGFILE = Rails.root.join('log/data_curation.log')
+  PATTERN = 'Status: '
+
+  attr_reader :logger, :pattern, :responsible
+
+  def initialize(logfile: LOGFILE, logger: nil, pattern: PATTERN, responsible: User.default_admin)
+    @logger = logger || Logger.new(logfile)
+    @pattern = pattern
+    @responsible = case responsible
+                   when User
+                     responsible.id
+                   when Integer
+                     User.find_by(id: responsible)&.id
+                   end
+  end
+
+  # Main function to process the molecules
+  # @note It logs the processing steps in log/data_curation.log.
+  # @note: process
+  #   faulty_molecules - find all molecules with faulty molfile and loop through them
+  #     reindex_molecule - update inchikey column to avoid index collision
+  #     samples_empty? - check if associated samples present and destroy molecule if not
+  #     process_by_sample_molfile - try to create new molecules using each sample molfile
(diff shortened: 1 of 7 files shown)


lib/tasks/support/molecule_structure_curation.rb
```diff
@@ -0,0 +1,260 @@
...
+
+require 'logger'
+
+# Find problematic molefile and recreate molecules /update samples
+# @example: MoleculeStructureCuration.new.process
+# @note: process
+# find problematic molecules                                                                  get_faulty_molecules
+#   for each molecule
+#     -> find associated samples:                                                             get_samples
+#        -> if sample has molfile, try to recreate the Molecule                               get_or_create_molecule
```
adambasha0, line 11: find or create the Molecule

lib/tasks/support/molecule_structure_curation.rb
```diff
@@ -0,0 +1,260 @@
...
+#           -> if molecule is found, update the Sample
```
adambasha0, line 12: update the sample only if a new molecule is created successfully

lib/tasks/support/molecule_structure_curation.rb
```diff
@@ -0,0 +1,260 @@
...
+#
+#     -> reload associated samples                                                            get_samples
+#       -> if samples left, try to create a new molecule using the molecule cano_smiles       get_new_molecule
```
adambasha0, line 15: if samples with faulty molfiles are left

lib/tasks/support/molecule_structure_curation.rb
```diff
@@ -0,0 +1,260 @@
...
+    return if new_molecule.blank?
+
+    get_samples(molecule_id).each { |sample| update_sample(sample, new_molecule) }
+  end
+
+  # update the given sample with the new Molecule
+  # @param [Sample]
+  # @param [Molecule]
+  # @return [void]
+  def update_sample(sample, molecule)
```
adambasha0, line 85: What if the newly created molecule has a different molecular weight and other different attributes (specifically: molecular_mass in Sample model which is mirrored from molecular_weight check Sample.js line 902)?, is there an update needed? or shall we add this to the log info?

adambasha0 (changes requested):

lib/tasks/support/molecule_structure_curation.rb
```diff
@@ -0,0 +1,260 @@
...
+    molecule_id = molecule.id
+    attributes = { molecule_id: molecule_id }
+    attributes[:molfile] = molecule.molfile if faulty_molfile?(sample.molfile)
+    sample.update_columns(**attributes) # rubocop:disable Rails/SkipsModelValidations
```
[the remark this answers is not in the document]
PiTrem: IMO safer and should avoid unforeseen chain reaction

Although the only callback that should be used is to update the molecule_name
I will update this

lib/tasks/support/molecule_structure_curation.rb
PiTrem: I will add it to the log

lib/tasks/support/molecule_structure_curation.rb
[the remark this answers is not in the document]
PiTrem: reaction scheme or sample svg is not updated either
 but I d let the use to check it
I could add a notification to the user

adambasha0 (approved):

lib/tasks/support/molecule_structure_curation.rb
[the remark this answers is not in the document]
PiTrem: ok now the changes are tracked with the history tracker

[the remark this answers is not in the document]
PiTrem: image

PR #2440, opened 2025-04-10
refactor(UI): rework list and group rendering

maiwald:
This PR adds the `ElementListRenderer` and `ElementGrousRenderer` components that consolidate rendering of element lists and groups. This helps us keep the design and functionality consistent for all element types.

Currently the components contain checkboxes and drag handles for all items. The `ElementGroupsRenderer` also contains the logic grouping the list of elements by a passed-in groupKey and for expanding and collapsing groups.

The components expose render props for customising how individual elements are displayed.
- `ElementListRenderer` exposes `renderItem(item, showDetails)`
- `ElementGroupsRenderer` exposes `renderGroupHeader(items, toggleGroupCollapse)` and `renderGroupItem(item, showDetails)`.

This PR also updates all element lists to use the new components, extracting parts into separate components where sensible. Where possible, the layout of individual element list/group items has been kept as it was. Redesigning the contents of element list items is not part of this PR.

This PR also removes the keyboard navigation feature from some lists. The previous implementation was barely usable as the keyboard selection would move outside of the view because the list would not scroll with the selection. Apart from the obvious fix, a unified concept for keyboard navigation is necessary to implement the feature again.


PiTrem (approved):
Thx

PR #2444, opened 2025-04-15
fix(db): molecule index on inchkey-sum formula to bypass openbabel  formula issue

PiTrem:
__Change the uniq index on molecule inchikey to inchikey + sum_formula__
and adapt the molecule creation by molfile accordingly

In case a molecule was created with the openbabel calculation for the sum formula not being correct, this allows the user to create correct molecule by enforcing the valence value in the molfile

Temporary solution while more work is in progress.


spec/api/molecule_api_spec.rb
```diff
@@ -117,6 +111,18 @@
           })
         end
       end
+
+      context 'when a molecule with a corrupt molfile is already present', skip: 'present for reproducibility' do
+        it 'cannot creates the correct molecule if one with corrupt molfile is there' do
```
adambasha0, line 116: Ambiguous Test Description:
Suggestion: it 'does not create a molecule with the same ID as the corrupt one' do

adambasha0 (approved):

spec/api/molecule_api_spec.rb
PiTrem: thx I will revisit those tests... 
but as discussed... later😅

PR #2464, opened 2025-04-24
chore(api): remove unused endpoind

PiTrem:


app/jobs/message_incoming_data_job.rb
```diff
@@ -25,7 +25,7 @@ def correspondent(arg)
     when User
       arg.id
     when Device
-      arg.groups.any? ? arg.groups.first.id : Admin.first.id
+      arg.groups.any? ? arg.groups.first.id : User.default_admin.id
```
headri, line 28: It should be Admin.default_admin here like everywhere else.

headri (changes requested):

app/jobs/message_incoming_data_job.rb
PiTrem: default_admin is only defined atm for the parent class User, so to me it's the same 🤷‍♂️

headri (approved):

app/jobs/message_incoming_data_job.rb
PiTrem: actually since the subclasses are scoped default_admin can only be present for User and Admin but not for Person or Group

PR #2472, opened 2025-04-30
Track reference

headri:
Adds tracking to the reference tab. The data is tracked in the literatures table.

### DIFF
--- a/app/models/literature.rb
+++ b/app/models/literature.rb
@@ -3,14 +3,14 @@
 # Table name: literatures
 #
 #  id         :integer          not null, primary key
+#  deleted_at :datetime
+#  doi        :string
+#  isbn       :string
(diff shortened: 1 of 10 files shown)


alizaib1217 (changes requested):
@headri 
Everything looks good to me overall -- Just a few questions/concerns:

1. It looks like we can’t edit a reference once it’s created — is that expected?
2. When a reference is edited and moved between categories (e.g. “cited own work”, “cited reference”, “referring to”), I don’t see that change being captured in version history.
3. Similarly, when a reference is deleted, its history doesn't seem to be recorded either.

Thanks

headri:
@alizaib1217 Fixed 2. and 3. : categories change and deletion is now shown in history and revertible
About 1. Editing references is currently not planned.

PR #2491, opened 2025-05-08
added log_data to collections

headri:


app/services/versioning/fetchers/collection_fetcher.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class Versioning::Fetchers::ResearchPlanFetcher
```
JanCBrammer: Is the class supposed to be named `CollectionFetcher` (i.e., `ResearchPlanFetcher` left over from copy-paste)?

JanCBrammer (approved):

PR #2494, opened 2025-05-09
fix: drag-n-drop preview drift in collection management

maiwald:
The library we are using assumes all items in the tree have the same height and does not display the preview properly when the don't.

Because we were displaying sync information below a collection in `ManageCollections`, individual collection divs would have more height than others when they were shared with others. This made rearranging collections in a long list with many synced collections very hard and at times practically impossible.

To fix this, sync information is now shown in a Button/Popover instead of as a separate line below the collection. This gives every collection in the tree the same height.


PiTrem:
image

PiTrem (approved):

PiTrem:
## preview
image

PR #2519, opened 2025-05-20
refactor(variations): change data structure from array to object

headri:
Changed how variation are saved. Replaced the outer array with an object with uuid key. Also added uuid to inner object for identification:

## Old state:
```json
[
  {"id": 2, "metadata": { }, "products": { }},
  {"id": 1, "metadata": { }, "products": { }},
  ....
]
```
## New state with uuid

```json
{ 
  "_uuid_a_": {"uuid": "_uuid_a_", "id":1, "metadata": { } "products": {} }
  "_uudi_b_": {"uuid": "_uuid_b_", "id":1, "metadata": {}, "products": {} }
...
}
```

- Added migration for old db entries
- Changed reading of this value, so that it stay an array in the code like before
- Added before_save hook, to write the value as object in db for update / insert.


StarmanMartin (approved):

app/models/reaction.rb
```diff
@@ -298,6 +305,15 @@ def scrubber(value)
     Chemotion::Sanitizer.scrub_xml(value)
   end
 
+  def transform_variations
+    return unless variations.is_a?(Array)
```
JanCBrammer, line 309: Under which circumstances are `variations` not an array?

PR #2537, opened 2025-05-28
fix: prevent null cas search and show instrument modal on top of Gene…

alizaib1217:
### Description

This PR fixes two issues:
- Added null check for input before calling validateCas while smile search in sample
- Ensured instrument modal shows on top of GenericDSDetails 

refs: #2511

### DIFF
--- a/app/javascript/src/components/container/ContainerDatasetModalContent.js
+++ b/app/javascript/src/components/container/ContainerDatasetModalContent.js
@@ -699,7 +699,10 @@ export class ContainerDatasetModalContent extends Component {
               rootClose
               onHide={() => this.abortAutoSelection()}
             >
-              <ListGroup className="mt-0 w-100 overflow-auto" style={{ maxHeight: '200px' }}>
+              <ListGroup
+                className="mt-0 w-100 overflow-auto rounded-3 shadow-lg"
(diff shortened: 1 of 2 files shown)


app/javascript/src/components/container/ContainerDatasetModalContent.js
```diff
@@ -699,7 +699,10 @@ export class ContainerDatasetModalContent extends Component {
               rootClose
               onHide={() => this.abortAutoSelection()}
             >
-              <ListGroup className="mt-0 w-100 overflow-auto" style={{ maxHeight: '200px' }}>
+              <ListGroup
+                className="mt-0 w-100 overflow-auto rounded-3 shadow-lg"
+                style={{ maxHeight: '200px', zIndex: 1 }}
+              >
```
maiwald, line 705: Responding to @PiTrem's request for comment on the inline style:
Looking at the code, it seems to me that this construct allows selecting items from a list. Does it also allow creating new items? Either way, if thats the case, it should probably be replaced with a standard `Select` component from `app/javascript/src/components/common/Select.js` altogether, rather than fixing individual styles on a non-standard component.

PiTrem (approved):
refactoring needed as stated by @maiwald

PR #2551, opened 2025-06-16
refactor: instrument field to use standard select component

alizaib1217:
simplify logic by using  a standard Select component from app/javascript/src/components/common/Select.js
solve style issues with the dropdown list

refs: #2537

### DIFF
--- a/app/javascript/src/components/container/ContainerDatasetModalContent.js
+++ b/app/javascript/src/components/container/ContainerDatasetModalContent.js
@@ -51,7 +51,7 @@ export class ContainerDatasetModalContent extends Component {
     this.thirdPartyApps = thirdPartyApps;
     this.state = {
       datasetContainer,
-      instruments: null,
+      instruments: [],
       timeoutReference: null,
       attachmentEditor: false,
       extension: null,
@@ -338,7 +338,7 @@ export class ContainerDatasetModalContent extends Component {
     const { datasetContainer } = this.state;
     this.setState({
       value: '',
-      instruments: null,
+      instruments: [],
     });
     datasetContainer.extended_metadata.instrument = '';
   }
@@ -361,7 +361,7 @@ export class ContainerDatasetModalContent extends Component {
           newState.instruments = result;
           newState.showInstruments = show;
         } else {
-          newState.instruments = null;
+          newState.instruments = [];
           newState.error = '';
           newState.showInstruments = false;
         }
@@ -571,7 +571,7 @@ export class ContainerDatasetModalContent extends Component {
   }
 
   renderMetadata() {
-    const { datasetContainer, instruments = [] } = this.state;
+    const { datasetContainer, instruments } = this.state;
     const { readOnly, disabled, kind } = this.props;
     const termId = absOlsTermId(kind);
     const klasses = (UserStore.getState() && UserStore.getState().dsKlasses) || [];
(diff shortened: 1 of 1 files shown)


app/javascript/src/components/container/ContainerDatasetModalContent.js
```diff
@@ -667,6 +580,7 @@ export class ContainerDatasetModalContent extends Component {
     if (idx > -1) {
       klass = klasses[idx];
     }
+    console.log({ instruments });
```
haditariq: Is it required?

app/javascript/src/components/container/ContainerDatasetModalContent.js
```diff
@@ -658,7 +571,7 @@ export class ContainerDatasetModalContent extends Component {
   }
 
   renderMetadata() {
-    const { datasetContainer, showInstruments } = this.state;
+    const { datasetContainer, instruments = [] } = this.state;
```
haditariq: initialise instruments as [] but not null?

app/javascript/src/components/container/ContainerDatasetModalContent.js
```diff
@@ -679,33 +593,33 @@ export class ContainerDatasetModalContent extends Component {
...
+                const value = selectedOption ? selectedOption.value : '';
+                this.handleInstrumentValueChange({ target: { value } }, this.doneInstrumentTyping);
+              }}
+              onInputChange={(inputValue, { action }) => {
+                if (action === 'input-change') {
+                  this.handleInstrumentValueChange({ target: { value: inputValue } }, this.doneInstrumentTyping);
+                }
+              }}
+              options={instruments?.map((item) => ({
+                label: item.name || item,
```
haditariq: inconsistent rendering for label and value. Item could be an obeject?

app/javascript/src/components/container/ContainerDatasetModalContent.js
```diff
@@ -679,33 +593,33 @@ export class ContainerDatasetModalContent extends Component {
...
-                event,
-                this.doneInstrumentTyping
-              )}
-              ref={(form) => { this.instRef = form; }}
-              autoComplete="off"
+            <CreatableSelect
+              isClearable
+              className="w-100"
+              value={
+                datasetContainer.extended_metadata.instrument
```
haditariq: datasetContainer?.extended_metadata?.instrument optional chaining?

app/javascript/src/components/container/ContainerDatasetModalContent.js
```diff
@@ -679,33 +593,33 @@ export class ContainerDatasetModalContent extends Component {
...
+                  ? {
+                    label: datasetContainer.extended_metadata.instrument,
+                    value: datasetContainer.extended_metadata.instrument
+                  }
+                  : null
+              }
+              isDisabled={readOnly || disabled}
```
haditariq, line 612: can we ensure that both varaibles are always boolean?

haditariq (changes requested):

app/javascript/src/components/container/ContainerDatasetModalContent.js
```diff
@@ -38,6 +38,7 @@ import { formatDate } from 'src/utilities/timezoneHelper';
 import UIStore from 'src/stores/alt/stores/UIStore';
```
haditariq, line 38: image

Enter a space in instrument field and it crashes!
My case: no instrument exists before to be searched!

haditariq (approved):
@alizaib1217 the crash is not happening anymore. A suggestion would be if you can trim the instrument query field that could skip some calls and make it more efficent as debounce is being used.

Everything looks good.

alizaib1217:
@haditariq , If we trim the query we wont be able to get results against. Instruments having space like "Bruker 1000".

PR #2568, opened 2025-07-01
feat: make creatableselects editable across all elements

alizaib1217:
**Description**

This PR standardizes and enables editable (creatable) select inputs across the application for a consistent user experience.

For example, as it was in v1.9, the user can select a molecule name for a sample and edit the value and save it. Same for a dataset instrument. 

**Key changes:**

- Updated all relevant CreatableSelect and Select components to support input editing, clearing, and creation of new options.
- Unified props usage (isClearable, isInputEditable, inputValue, onInputChange, allowCreateWhileLoading, formatCreateLabel) across SampleForm, CellLineName, GeneralProperties, Affiliations, ComputedPropsGraphContainer, and SampleDetails.
- Improved state management and value matching for selects to ensure correct display and editing.
- Fixed issues where selected values were not editable or labels were not displayed correctly.
- Enhanced user experience by allowing direct editing and creating select options everywhere needed.

------------------------------

### DIFF
--- a/app/javascript/src/apps/mydb/elements/details/cellLines/propertiesTab/CellLineName.js
+++ b/app/javascript/src/apps/mydb/elements/details/cellLines/propertiesTab/CellLineName.js
@@ -12,7 +12,8 @@ export default class CellLineName extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      nameSuggestions: []
+      nameSuggestions: [],
+      cellLineNameInputValue: props.name || ''
     };
   }
 
@@ -22,6 +23,15 @@ export default class CellLineName extends React.Component {
       .then((data) => {
         this.setState({ nameSuggestions: data });
       });
+  }
+
+  componentDidUpdate(prevProps) {
+    // Sync cellLineNameInputValue when name prop changes
+    if (prevProps.name !== this.props.name) {
+      this.setState({
+        cellLineNameInputValue: this.props.name || ''
+      });
+    }
   }
 
   static renderNameSuggestion(name, src) {
@@ -38,7 +48,7 @@ export default class CellLineName extends React.Component {
 
   render() {
     const { cellLineDetailsStore } = this.context;
-    const { nameSuggestions } = this.state;
+    const { nameSuggestions, cellLineNameInputValue } = this.state;
     const { id, name, readOnly } = this.props;
 
     if (readOnly) {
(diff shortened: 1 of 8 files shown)


maiwald (changes requested):
I think the general approach of extending React Select is viable. However, it seems to me that we are confusing two similar looking use cases for this type of UI element.

1. **Input for selecting option(s) from a fixed list of options, with the ability to add new options to that (possibly shared) list**
This is what CreatableSelect provides. But not necessarily what we want in this situation.

2. **Text input with auto-completed suggestions**
I think this is what we actually want here. We want to allow users to either provide their own input or select from a list of suggestions. In both cases the value would be editable.

So my suggestion would be to create a new component that uses the non-creatable version of React Select, add the ability to freely edit the text, and use that instead of configuring CreatableSelect here. Then that new component would provide a clear API where users of the component specify at minimum something like this pseudo code:

```js
<InputWithSuggestions
  value={"foo"}
  suggestions={["foo", "bar", "baz"]}
  onChange={(newValue) => { … }}
/>
```

If something like this already exists in the code base, even better.

As for the other uses of CreatableSelect, we should check whether they call for option 1 with CreatableSelect or option 2 and use the new component.

/cc @PiTrem

maiwald:
@alizaib1217 Here are two issues I have found with this version.

https://github.com/user-attachments/assets/098a1e44-700b-49d1-ab85-e67af3b7473c

https://github.com/user-attachments/assets/ddbacc64-90bf-48de-a23b-b64794fc4996

This makes me wonder whether this should rather be a new component (potentially based on React-Select but not necessarily) that functions like a regular text input with additional suggestions instead of extending the Select components that only allow selecting from a list (or adding to a global list of options in case of the creatable variant).

PiTrem:
@maiwald 
> This makes me wonder whether this should rather be a new component (potentially based on React-Select but not necessarily) that functions like a regular text input with additional suggestions instead of extending the Select components which are meant to only allow selecting from a list (or adding to a global list of options in case of the creatable variant). Then you'd be free to change the (React-Select) behaviour more drastically.

it has to show multiple `suggestions`  or `options`  even without text input from the user
and the user should be able to create a text input from scratch, or by modifying one of the suggestion.
so yes it is a 🧟‍♂️franken-select!

PiTrem (approved):

PR #2571, opened 2025-07-01
Fix PrintCodeButton dropdown menu being cut off by `overflow: hidden`

### LINKED ISSUE #2564: Could not see preview of print code label dropdown
- I am an ELNer
- I am a developer
- I am a system/ELN admin

# Expected Behavior

When we click on Print Code Label option in the element header, the dropdown component should appear on top of below component.

# Current Behavior

When we click on Print Code Label option in element header the dropdown component hides behind below component.

## Steps to Reproduce

Image

Please see the attached screenshot

## Issue Branch

Current main

maiwald:
Fixes #2564


alizaib1217 (changes requested):
Hi @maiwald , Found few issues
1. **Dropdown Cutoff**
On the Sample Details > Analyses tab, and also in the Device Description > Analyses section, the dropdown menu is being cut off on the right side. Please refer to the attached screenshot.

image

2. **Sample Tab – Reaction Details Analyses**
In the Reaction Details > Analyses section, under the Sample tab, the layout seems misaligned or broken.

image


3. **Wide Layouts**
The following tabs seem too wide and could benefit from improved spacing or responsiveness:
Reaction Details > Reaction tab
Wellplate Detail > Analyses tab
Screen Detail > Analyses tab

image

4. **Responsive Mode – Print Code Label Missing**
In Screen Details and wellplate,  when switching to responsive view (Toggle Device Toolbar), the “Print Code” label dropdown is not visible.
Please see the attached video

https://github.com/user-attachments/assets/50f0d2e3-e678-43d1-9d40-e064fb4f9447

maiwald:
@alizaib1217 I have adapted the fix for the other instances you mentioned, except 4 (Responsive Mode – Print Code Label Missing), since I believe this is an issue with the responsive mode. With smaller window sizes the dropdown is visible.

I have also discussed the issue with @JoachimTillessen, since this is a systemic issue with how react bootstrap implements dropdown menus. He will take over and fix another instance but a general fix is beyond this PR and needs further discussion first.

alizaib1217:
> @alizaib1217 Good catch!
> 
> I see the broken dropdown menus as well. The problem is that the library for handling menu positioning used in react-bootstrap does not seem to work reliably in contexts with `overflow: hidden`. Worst case would be replacing all Dropdown Menus. I'll see what I can do.
> 
> As for item 3, I'd say redesigning the analyses tab is beyond the scope of this PR. Or am I misunderstanding something?

In point 3, I meant that the drop-down menu in point 3 is very wide. Like, there is a lot of empty space next to "Small label" or "Large Label" text. I believe you made the latest commit to fix the issue, but it has not been fixed yet. Can you please share a screenshot if it's fixed at your end?

maiwald:
> @maiwald In point 3, I meant that the drop-down menu in point 3 is very wide. Like, there is a lot of empty space next to "Small label" or "Large Label" text. I believe you made the latest commit to fix the issue, but it has not been fixed yet. Can you please share a screenshot if it's fixed at your end?

Ah, I understand. Bootstrap defines a [min-width for Dropdown menus](https://github.com/twbs/bootstrap/blob/3bee7c15d4f301b03098f6ac534ae54f48b59240/site/src/scss/_navbar.scss#L91). I'd rather not override that and stick with standard components unless absolutely necessary.

Regarding the width-change, @PiTrem asked me to look at a similar issue with Select components and I offered to fix that issue in this PR. Here are videos for the Select issue

Before

https://github.com/user-attachments/assets/168ccb5c-05b6-4494-9137-8ef9825fad70

Overflow fix

https://github.com/user-attachments/assets/3b5a72ca-8674-4a18-b0ec-ce7127cac001

Spacing improvement

https://github.com/user-attachments/assets/c586ce41-b68a-4e57-86ff-d0996306f40d

PR #2581, opened 2025-07-08
Only strip last extension in fname_wo_ext to preserve internal dots

### LINKED ISSUE #2580: Generated spectrum image is deleted when source filename contains multiple dots
# Description
When uploading a raw data file whose name includes more than one dot (for example raman_jcamp.2.jdx, the post-processing routine identifies the freshly generated .peak.png image as "matching" the base name of the source file and immediately deletes it. This happens because the fname_wo_ext helper (/app/models/concerns/attachment_jcamp_aasm.rb) strips all segments after the first dot instead of only removing the final extension.

### Steps to Reproduce
- Upload a file named raman_jcamp.2.jdx to an analysis container.
- Trigger peak generation.
- Observe that the .peak.png image is created and then instantly removed.

### Expected Behavior
The generated image (raman_jcamp.2.peak.png) should remain in the container and not be deleted.

### Actual Behavior
The new image is deleted because its “base name” (computed by removing all extensions) matches the source file’s base, so delete_related_imgs treats it as a duplicate.

### File for testing (remove .txt)
[raman_jcamp.2.jdx.txt](https://github.com/user-attachments/files/21122015/raman_jcamp.2.jdx.txt)

## Suggested Fix
Update fname_wo_ext to only strip the last file extension (everything after the final .), preserving any other dots in the filename.

```
  def fname_wo_ext(target)
    parts = target.filename_parts
    if parts.length >= 2
      parts[0..-2].join('.')
    else
      parts[0]
    end
  end
```

Nicolass67:
## What
Update `fname_wo_ext` so that it only removes the very last file extension, leaving any other dots in the original filename intact.

## Why
Previously, filenames with multiple dots (e.g. `raman_jcamp.2.jdx`) had all extensions stripped, causing the base name of generated images (`.peak.png`) to match the source and be deleted by `delete_related_imgs`. This change prevents unintended deletion of freshly generated spectrum images.

## How
Replace the old logic with:

```ruby
def fname_wo_ext(target)
  parts = target.filename_parts
  if parts.length >= 2
    parts[0..-2].join('.')
  else
    parts[0]
  end
end
```
## Impact
Generated images for multi-dot filenames will now remain in place as expected.

Closes #2580

### DIFF
--- a/app/models/concerns/attachment_jcamp_aasm.rb
+++ b/app/models/concerns/attachment_jcamp_aasm.rb
@@ -503,8 +503,11 @@ module AttachmentJcampProcess
 
   def fname_wo_ext(target)
     parts = target.filename_parts
-    ending = parts.length == 2 || parts.length == 3 ? -2 : -3
-    parts[0..ending].join('_')
+    if parts.length >= 2
+      parts[0..-2].join('.')
+    else
+      parts[0]
+    end
   end
 
   def delete_related_imgs(img_att)


haditariq:
When user clicks on the empty thumbnail in the add analysis area there is a bug. 

@Nicolass67 Is this part of this PR 

https://github.com/user-attachments/assets/c9c961cd-adec-471a-80c9-6fd00b502da1

haditariq:
Cases:
- raman_jcamp.jdx (pass)
- file.2.2..abc..jdx (fail)
- raman_jcamp.Copy.jdx(pass)
- raman_jcamp.  124.456.jdx (pass)

@Nicolass67 I have tested some cases. Only one case is failing!

Thanks

PiTrem:
@haditariq @Nicolass67 
> 
> I have tested some cases. Only one case is failing!
> 
Added to the refactoring task.
I will accept the proposed solution for now.

PR #2595, opened 2025-07-15
fix(UI): Fix select menus in modals

maiwald:
Before
Screenshot 2025-07-15 at 14 40 56

After
Screenshot 2025-07-15 at 14 40 29


alizaib1217:
Hi @maiwald 
I just tested the branch and it's working as expected on My Labels.

image


But it's not working right on Well Details for the well. Please see ss. 

image

PR #2618, opened 2025-07-24
bugfix: no attachment handling in ImageModal to prevent crashes

alizaib1217:
**Description**

This PR addresses a bug in the ImageModal component where accessing properties of a null attachment caused a TypeError. The following changes were made:

- Added null checks in fetchImage() and fetchImageThumbnail() methods.
- Set fallback images when attachment is unavailable.
---------------------

- [ ] added code is linted

### DIFF
--- a/app/javascript/src/components/common/ImageModal.js
+++ b/app/javascript/src/components/common/ImageModal.js
@@ -90,6 +90,13 @@ export default class ImageModal extends Component {
 
   fetchImage() {
     const { attachment } = this.props;
+    if (!attachment || !attachment.id) {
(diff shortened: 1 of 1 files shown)


haditariq (commented):
Funcation is not crashing looks good, Would it be possible to: 

- use a try/catch block
- use the already defined function to set fallback state [handleImageError](https://github.com/ComPlat/chemotion_ELN/blob/9abe6b388663254ec657922961ee6025a2c6f67e/app/javascript/src/components/common/ImageModal.js#L71C3-L71C19)

PR #2620, opened 2025-07-25
style: improve yield/conversion rate toggle button implementation

adambasha0:
**Problem:**
1.  Inconsistent UI styling - Toggle buttons used custom color properties instead of standard Bootstrap variants
2. Poor user experience - Missing explanatory tooltips for yield field calculations 

**Solution:**
1. Standardized styling system - Replaced custom onColor/offColor props with Bootstrap variant prop
2. Enhanced user guidance - Added informative tooltip explaining yield calculation basis
3. Improved component consistency - Updated ToggleButton to use buttonClass and buttonTitleClass props
4. Better text formatting - Implemented JSX elements for cleaner tooltip presentation
__________________________________________________________________________________________________________________

- [ ] added code is linted

### DIFF
--- a/app/javascript/src/apps/mydb/elements/details/reactions/schemeTab/Material.js
+++ b/app/javascript/src/apps/mydb/elements/details/reactions/schemeTab/Material.js
@@ -275,17 +275,32 @@ class Material extends Component {
 
   yieldOrConversionRate(material) {
     const { reaction, displayYieldField } = this.props;
+    const yieldMessage = (
+      <>
+        The final yield value calculated upon saving the reaction
+        is based on the real amount field value of this product.
+      </>
+    );
     if (displayYieldField === true || displayYieldField === null) {
       return (
         <div>
-          <Form.Control
-            name="yield"
-            type="text"
-            bsClass="bs-form--compact form-control"
-            size="sm"
-            value={this.calculateYield(material, reaction) || 'n.d.'}
-            disabled
-          />
+          <OverlayTrigger
+            placement="top"
+            overlay={(
+              <Tooltip id="yield-tooltip">
+                {yieldMessage}
+              </Tooltip>
+            )}
+          >
+            <Form.Control
+              name="yield"
+              type="text"
+              bsClass="bs-form--compact form-control"
+              size="sm"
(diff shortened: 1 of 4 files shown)


adambasha0:
[Chemotion dev (35).webm](https://github.com/user-attachments/assets/4e1cbb29-d134-401b-b45b-c3dcc1801879)

alizaib1217 (changes requested):
Hi @adambasha0,
Overall, the code looks good! Just a few things I noticed:

1. In the "Conv." field, if I enter a value like 100, I can't fully clear it — I can delete two digits, but not the last one. The field visually still shows the value until I focus on another field, at that point it resets to 0.
2.  The same issue occurs with the "Amount" field — it doesn't immediately reflect the cleared value unless I shift focus to another input.
3. When I click the "Save" button, I see a duplication of all product elements. Please check the attached video for reference:


https://github.com/user-attachments/assets/e745fdac-7b60-4222-8cfa-d484b4342762


4. Instead of the grey button, can we use an Ant Design Switch component?

Thanks!

alizaib1217:
> > Hi @adambasha0, Overall, the code looks good! Just a few things I noticed:
> > 
> > 1. In the "Conv." field, if I enter a value like 100, I can't fully clear it — I can delete two digits, but not the last one. The field visually still shows the value until I focus on another field, at that point it resets to 0.
> > 2. The same issue occurs with the "Amount" field — it doesn't immediately reflect the cleared value unless I shift focus to another input.
> > 3. When I click the "Save" button, I see a duplication of all product elements. Please check the attached video for reference:
> > 
> > PR.2620.mp4
> > 4. Instead of the grey button, can we use an Ant Design Switch component?
> > Thanks!
> 
> Hi Ali, thanks for the review, for points 1, and 2, these are related to the NumeralInputWithUnitsCompo component, which is used for many fields in the reaction scheme. Thus, not related to this PR. You can create a new issue, shall it is confirmed to be an issue. 3. could be, due to buggy old dummy seed data, try to create a new reaction.

Hi Adam,
Overall looks good to me. For point 3 - I tested with a new reaction and it works fine, but old reactions still show the duplication issue.
Thanks!

PR #2639, opened 2025-08-05
fix: tabs navigation in advanced search result

beque:
Changing tabs in search result not always working.
You can't change back to the first active tab


alizaib1217:
Hi, @beque 
Can you share a video of what was done?

beque:
Hi @alizaib1217 
here is a video of the not working search result tabs
https://github.com/user-attachments/assets/145e75e9-3aab-416a-9fbf-d0622e26afe8

I can change the tab contents but I can't return to the sample tab.
Only the sample tab was highlighted.

To fix this, I have to move `SearchResultTabContainer` back to `SearchResult`.
I also have to fix some variables and class names

alizaib1217:
Hi, Looks Good.

PR #2642, opened 2025-08-08
feat: add history trace to cell-line elements

headri:
Added tracking for cellline sample and materials to database.
Added History Tab to Celllines to view and revert changes to cellline sample.
Celline Material changes cannot be viewed because there is no dedicated view for materials.


JanCBrammer (approved):
I've tried the feature in the UI and it seems to work as intended.
The implementation looks good to me.

PR #2667, opened 2025-08-27
feat: partial data migration between 2 instances

headri:
1. restore user data of selected users from backup
    a. Users, profiles, devices, labels
    b. collections (synced & shared), inventory
    c. containers, attachments
    d. samples, reactions, wellplates, wells, screens, research_plans, device_descriptions
    e. elements, segments
    
 restorer = RecoveryDB::PartialMigration::RestoreUsers.new(user_ids: [], file: 'tmp/db_backups/backup.sql')
restorer.run


lib/tasks/support/recovery_db/partial_migration.rb
```diff
@@ -0,0 +1,617 @@
...
+      end
+
+      def restore_user_group(user_id_map)
+        RecoveryDB::Models::UsersGroup.find_each do |ua|
+          new_group_id = user_id_map[ua.group_id]
+          new_user_id  = user_id_map[ua.user_id]
+
+          next unless new_group_id && new_user_id
+
+          UsersGroup.create!(admin_id: new_group_id, user_id: new_user_id)
```
StarmanMartin: Seems like admin id is wrong

lib/tasks/support/recovery_db/partial_migration.rb
```diff
@@ -0,0 +1,617 @@
...
+        raise ArgumentError, 'User Ids are required' if @user_ids.nil?
+        raise ActiveRecord::RecordNotFound, 'No user found' if rec_users.empty?
+
+        user_id_map = {}
+        created_elements = {}
+
+        @mount.log_event "Restoring users: found #{rec_users.size} of #{@user_ids.size} user(s) to restore"
+        restore_element_klasses(created_elements)
+        rec_users.each do |recovery_user|
+          new_user = restore_user(recovery_user, created_elements)
```
StarmanMartin: -  Many restore_* methods run DB operations without wrapping them in transactions.
- If part of a restore fails, you may end up with half-restored, inconsistent state.
- Suggestion: Use transactions at critical levels:
```ruby
ActiveRecord::Base.transaction do
  new_user = restore_user(...)
  restore_profile(...)
  restore_collections(...)
end
```

lib/tasks/support/recovery_db/partial_migration.rb
```diff
@@ -0,0 +1,617 @@
...
+
+      def run
+        raise ArgumentError, 'User Ids are required' if @user_ids.nil?
+        raise ActiveRecord::RecordNotFound, 'No user found' if rec_users.empty?
+
+        user_id_map = {}
+        created_elements = {}
+
+        @mount.log_event "Restoring users: found #{rec_users.size} of #{@user_ids.size} user(s) to restore"
+        restore_element_klasses(created_elements)
```
StarmanMartin, line 45: **Avoid inconsistent state:**
Ensure that all database actions are performed within transactions and that, if an error occurs, all previous steps revert to their original state and the process is terminated. In such a case, it is important to display a meaningful error message, for example 'Chemotion versions are not compatible'.

lib/tasks/support/recovery_db.rb
```diff
@@ -0,0 +1,257 @@
...
+    MODEL_NAMES = %w[
+      User Profile Collection Sample Reaction Wellplate Well
+      Screen ResearchPlan DeviceDescription Attachment
+      Container ContainerHierarchy CollectionsSample
+      CollectionsReaction CollectionsWellplate CollectionsScreen
+      CollectionsResearchPlan CollectionsDeviceDescription Labimotion::ElementsSample
+      SyncCollectionsUser UsersDevice Device Labimotion::CollectionsElement
+      Labimotion::Element Labimotion::ElementKlass Labimotion::ElementsElement
+      Labimotion::SegmentKlass Labimotion::Segment UserLabel UsersAdmin UsersGroup
+    ]
```
StarmanMartin: add .freeze: to makes it reusable and easier to maintain.

StarmanMartin (changes requested):

PR #2679, opened 2025-09-19
feat: improve sample class and components: molecular weight, labels, and mixture handling

TasnimMehzabin:
- Added additional input fields for mixtures
- Included molar mass and relative molecular weight information
- Fixed an issue where switching the sample type after drawing a structure caused inconsistencies
- Fixed an issue with SVG images when adding or deleting components in a specific use case
- Improved report export logic for mixtures

-------------------------------------------
- Show components' label with molecular weight.

image

- Fix the export of components when **_All pages_** is selected.

https://github.com/user-attachments/assets/07a8bf70-2e1d-49f2-a86a-0c96f9a6c61b

- Fix the component creation when user switches the sample type from _**Single molecule**_ to **_Mixture_**, as the switch was previously not functioning correctly.

https://github.com/user-attachments/assets/2dc1d713-b71b-41fc-b665-cb6d6527d69c

- When a molecule structure is drawn in the editor, dragging and dropping a sample into the components list previously did not update the SVG. The same issue occurred when a component was deleted from the list. Both issues have now been fixed.

https://github.com/user-attachments/assets/ef407cd8-77a9-4869-af41-c176b9ec56a9

### DIFF
--- a/spec/javascripts/packs/src/models/Component.spec.js
+++ b/spec/javascripts/packs/src/models/Component.spec.js
@@ -32,17 +32,23 @@ describe('Component', () => {
 
   describe('calculateVolumeForLiquid', () => {
     it('should calculate volume from density when density is provided', () => {
-      component.amount_mol = 1; // 1 mol
-      component.amount_l = 1; // 1 L
-      component.calculateVolumeForLiquid(1); // Call with purity 1
-      expect(component.amount_l).toBe(1);
+      component.density = 1.0; // g/mL
+      component.amount_mol = 2; // mol
+      const purity = 1;
+
+      component.calculateVolumeForLiquid(purity);
+
+      // amount_l = (amount_mol * molecular_weight) / (density * 1000 * purity)
+      const expectedAmountL = (component.amount_mol * component.molecule.molecular_weight)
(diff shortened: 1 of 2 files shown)


app/javascript/src/models/Sample.js
```diff
@@ -908,6 +947,38 @@ export default class Sample extends Element {
       this.updateTONPerTimeValue(value, gasPhaseTime);
     }
   }
+
+  /**
+   * Calculates the volume (in liters) for mixture samples based on total mass and density/molarity.
+   *
+   * @param {number} totalMassGrams - The total mass of the mixture in grams
```
adambasha0: param name is not uptodate with param name of the function (amount_g)

app/javascript/src/models/Sample.js
```diff
@@ -1510,6 +1731,7 @@ export default class Sample extends Element {
   /**
    * Sets the reference component in the mixture by index and updates equivalents.
    * @param {number} componentIndex - The index of the component to set as reference.
+   * @throws {Error} If componentIndex is invalid or components array is empty.
```
adambasha0: where does it throw Error?

PR #2697, opened 2025-10-07
ci(test): optimize test workflow

PiTrem:
Reduce testing run duration for JS-only PR from currently 15~20 min to ~2 min 

- extract js testing into a separate workflow
- use actions/nodejs with version set from .tool-versions
- cache js packages for main
- only run on PR when js code base has changes. Watched paths:
  - app/javascript
  - spec/javascripts
  - packages.json
  - yarn.lock
  - .tool-versions

- only run ruby test if backend code base has changes
  (paths-ignore)


.github/workflows/ci-rb.yml
JanCBrammer, line 1: Rename this file `ci-ruby.yml` to make the separation more explicit?

.github/workflows/ci-rb.yml
```diff
@@ -1,8 +1,15 @@
-name: Continuous Integration
+name: CI - Ruby Lint, RSpec, and code coverage
 
 on:
   workflow_dispatch:
   pull_request:
+    paths-ignore:
```
JanCBrammer, line 6: Maybe ignore all JS files (`- '**/*.js'`) to future-proof this (in case new JS dirs are added)?

.github/workflows/ci-js.yml
```diff
@@ -0,0 +1,98 @@
...
+
+      - name: Enable Corepack (Yarn)
+        run: corepack enable
+
+
+      # Get the yarn.lock from origin/main and hash it
+      - name: Compute main yarn.lock hash
+        id: mainlock
+        run: |
+          echo "hash=$(sha256sum tmp/main/yarn.lock | cut -d' ' -f1)" >> $GITHUB_OUTPUT
```
JanCBrammer: Couldn't the `key` used on lines 73 and 93 be pre-computed here?

I.e., `yarn-main-${{ runner.os }}-${{ steps.node_version.outputs.NODE_VERSION }}-${{ hashFiles('tmp/main/yarn.lock') }}` (Note the use of GitHub's built-in [hashFiles](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#hashfiles)).

.github/workflows/ci-rb.yml
PiTrem: ✅

.github/workflows/ci-js.yml
PiTrem: done

PR #2708, opened 2025-10-13
ci(test): optimize test workflow

PiTrem:
- use slimmer image for ruby testing to improve `Initianlize containers` step 

- use `actions/setup-node@v4` for ruby testing instead of asdf-vm packed nodejs

- bump `actions/checkout` from v4 to v5

- bump `actions/setup-node@v5` . NB global cache not used in rb workflow

- TODO:
  - fix sharing nodejs cachec between `js` and `ruby` workflows to improve the  `yarn install` step in `ruby` workflow
  - `linting` step: define pronto/rubocop dependencies in a dedicated Gemfile block to be able to use the bundler cache setting from `action/setup-ruby`

ref: #2697

### DIFF
--- a/.github/workflows/ci-rb.yml
+++ b/.github/workflows/ci-rb.yml
@@ -56,7 +56,10 @@ jobs:
       run:
         shell: bash
 
-    container: complat/chemotion_eln_runner:latest
+    container:
+      image: complat/chemotion_eln_runner:latest
+      ## options: --user root
+
 
     services:
       postgres:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-rb.yml
```diff
@@ -73,10 +73,50 @@ jobs:
...
+      YARN_CACHE_FOLDER: /home/runner/.cache/yarn
 
     steps:
       - name: checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
+
+      - name: Setup Nodejs
+        id: setup_node
+        uses: actions/setup-node@v4
```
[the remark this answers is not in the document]
PiTrem: For some reasons, yarn is not found during the ruby workflow at `Setup Nodejs` step, and the cache settings in v5 that are by default enabled leads to error since `yarn cache dir` cannot be executed

.github/workflows/ci-rb.yml
JanCBrammer, line 1: Most likely I'm missing sth obvious. Why do we need node in this workflow at all?

JanCBrammer: Turns out, we do (some Ruby tests rely on node).

PR #2736, opened 2025-10-29
fix(UI): prevent duplicate analysis

haditariq:
PR resolves: Issue with duplicate analysis 

### WHAT:
Creating multiple analyses for a sample and saving it using "Save and close" or "Sample Save" creates duplicate analyses.

### HOW:
 Make sure to add the dataset and use "Save and Close", then "SAVE" doesn't duplicate the analysis using the current container.
Also, fixed props update causing recursive rendering


alizaib1217 (approved):
Looks good. Thanks.

PR #2783, opened 2025-11-14
refactor!: collection sharing - merge shared and sync collections

jhmegorei:
This PR reworks the collection synching/sharing mechanism.

Before, there was collection sharing, where a collection was created, that was assigned to the user it was shared with, and the owner of the collection was saved in another field. Additionally, the synching system, which was basically a join table between a user and a collection.

Over time this has become cumbersome, so it became necessary of reworking it.

Now, the collection system has been simplified a lot:

Collections have one owner (who has by definition all rights and permissions to do anything with the elements of this collection). The owner is signified by the Collection->User relation. All collections will be migrated, so the owner of the collection is always `collection.user`.

Collections can be shared to any amount of other users. For this, we created a new model named CollectionShare (we intentionally did not use the SyncCollectionsUser model which did basically the same but was in our opinion badly named), which joins Users/Groups and Collections. All permission/detail level related fields are now in this model.

Conceptually, we can now differentiate between a user's own collections (i.e. where collection.user == user) and the collections that have been shared to the user (i.e. collection_share.shared_with == user). We added the corresponding scopes and other utility methods to work with the new concept.

There are still some things that need improvement (distinction between user and group, different ways of accessing elements of collections via a lot of joins, specs, readability, lot of logic issues concerning transitive sharing or ownership transfer) which we were not able to address due to time/budget constraints. We will see how those areas can be improved in the future.

Known issues:
- Generic Element API is still a work in progress, as we have no access to the Labimotion gem
- Feature Specs don't run locally, so we basically edited them blindly and now need to use the CI for further work
- Currently it is not possible to change the ownership of collections. This is due to the abovementioned logic issues in the collection system

### DIFF
--- a/spec/features/copy_reaction_spec.rb
+++ b/spec/features/copy_reaction_spec.rb
@@ -11,7 +11,7 @@ describe 'Copy reaction' do
       last_name: 'Complat',
       name_abbreviation: 'US1',
       account_active: true,
-      confirmed_at: Time.now
+      confirmed_at: Time.current,
     )
   end
   let!(:user2) do
@@ -21,7 +21,7 @@ describe 'Copy reaction' do
       last_name: 'Complat',
       name_abbreviation: 'US2',
       account_active: true,
-      confirmed_at: Time.now
+      confirmed_at: Time.current,
     )
   end
 
@@ -42,7 +42,7 @@ describe 'Copy reaction' do
       name: 'Material2',
       target_amount_value: 7.15,
       molecule: m1,
-      collections: [shared_collection_with_high_detail_level]
+      collections: [shared_collection_with_high_detail_level],
     )
   end
   let(:product2) do
@@ -51,7 +51,7 @@ describe 'Copy reaction' do
       name: 'Product2',
       real_amount_value: 4.671,
       molecule: m2,
-      collections: [shared_collection_with_high_detail_level]
+      collections: [shared_collection_with_high_detail_level],
     )
   end
   let(:reaction2) do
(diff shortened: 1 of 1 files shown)


app/api/entities/collection_share_entity.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Entities
+  class CollectionShareEntity < ApplicationEntity
+    expose! :id
+    expose! :collection_id
+    expose! :shared_with
+    expose! :shared_with_id
+    expose! :shared_with_type
+    expose! :shared_with_id
```
headri: line is duplicated

app/javascript/src/utilities/collectionUtilities.js
```diff
@@ -0,0 +1,78 @@
...
+      label: 'Shared with me collections',
+      options: makeList(shared),
+    },
+  ];
+}
+
+const collectionHasPermission = (collection, permissionLevel) => {
+  if (collection.permission_level === undefined) { return true }
+
+  return collection.collection_share_id && collection.permission_level <= permissionLevel;
```
headri: Because higher level means more is allowed, the condition should be ">=" instead of "<=". E.g. if I have permission_level 10, which allows all, this function should return true for lower permissionLevel calls.
Also check calls.

app/javascript/src/stores/alt/actions/ElementActions.js
```diff
@@ -301,9 +300,9 @@ class ElementActions {
     };
   }
 
-  fetchReactionsByCollectionId(id, queryParams = {}, collectionIsSync = false) {
+  fetchReactionsByCollectionId(id, queryParams = {} = false) {
```
headri: Should be "fetchReactionsByCollectionId(id, queryParams = {}) {" as in the function above.

PR #2837, opened 2025-12-18
feat: auto transfer inbox item to reaction analyses

headri:
Add Transfer from Inbox to Reaction
Add Auto Transfer for sample and reaction
Add config for auto (default on) and manual (default off) transfer in user profile
Add Link to Variation using (-v[variation_id]) in name - Example: CU1-R1-v1

### DIFF
--- a/app/api/chemotion/admin_user_api.rb
+++ b/app/api/chemotion/admin_user_api.rb
@@ -11,7 +11,8 @@ module Chemotion
       namespace :users do
         desc 'Find all users'
         get do
-          present User.order(type: :desc, id: :asc), with: Entities::UserEntity, root: 'users'
+          users = User.includes(:profile).order(type: :desc, id: :asc)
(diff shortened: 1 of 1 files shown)


app/javascript/src/apps/userSettings/AccountProfile.js
```diff
@@ -148,6 +174,51 @@ function AccountProfile({ currentUser }) {
 
       <InventoryLabelSettings />
 
+      <Card>
+        <Card.Header>Inbox to Element</Card.Header>
+        <Card.Body>
+          <Form onSubmit={handleProfileInboxSubmit}>
+            <Row className="mb-3">
+              <Form.Label column className="col-form-label col-3 offset-3">
+                Enable Auto Transfer Inbox to Element
```
StarmanMartin: A help text would be nice. It should explain to the user exactly what happens when they change these settings.

StarmanMartin (approved):

PR #2858, opened 2026-01-09
feat: SBMM calculation enhancements and usage in reaction scheme

TasnimMehzabin:
- Refactored activity-based calculations into modular methods for clarity and efficiency.
- Added support for purity-based recalculations and adjusted concentration handling for accuracy.
- Introduced methods for calculating volume and mass based on activity and purity.
- Enhanced purity integration with getter/setter and condition checks for recalculations.
- Added the functionality to use of the SBMM samples in the reaction scheme
- Report generations

### DIFF
--- a/lib/export/export_sdf.rb
+++ b/lib/export/export_sdf.rb
@@ -97,7 +97,10 @@ module Export
     end
 
     def sdf_molfile_for(sample)
-      sample['molfile'].presence
+      return sample['molfile'] if sample['molfile'].present?
+      return EMPTY_MOLFILE if sample['source_type'] == 'sbmm'
+
+      nil
     end
 
     def extract_reference_values(raw_value)
--- a/app/api/helpers/report_helpers.rb
+++ b/app/api/helpers/report_helpers.rb
@@ -820,7 +820,7 @@ module ReportHelpers
           , 'sbmm'::text as source_type
           , r_s.reaction_id
           , r_s.reference
-          , sbms.equivalent
+          , r_s.equivalent
           , null::float as conversion_rate
           , r_s.position
           , 'ReactionsReactantSbmmSample'::varchar as type
--- a/app/javascript/src/models/SequenceBasedMacromoleculeSample.js
+++ b/app/javascript/src/models/SequenceBasedMacromoleculeSample.js
@@ -431,7 +431,9 @@ export default class SequenceBasedMacromoleculeSample extends Element {
         break;
 
       case 'purity':
-        if (this.purity > 0) {
+        if (this.purity !== null && this.purity !== undefined && this.purity !== '') {
+          // Keep client-side calculations stable: treat 0/invalid purity as 1.0.
+          this._purity = this.normalizedPurityForCalculations();
           // Recalculate volume from mass: amount_l = amount_g / (concentration × purity)
           this.calculateVolumeByMass();
           // Recalculate amount_mol from volume: amount_mol = amount_l * molarity * purity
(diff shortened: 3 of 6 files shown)


db/schema.rb
```diff
@@ -1112,6 +1112,21 @@
     t.index ["rxno"], name: "index_reactions_on_rxno", order: :desc
   end
 
+  create_table "reactions_reactant_sbmm_samples", force: :cascade do |t|
```
adambasha0, line 1115: Is there a strong reason why do we need to create a new table for SBMM reactions reactant samples? maybe you can add sequence_based_macromolecule_samples_id to reactions_samples table and use mutual exclusive between sample_id &  (https://github.com/ruby-grape/grape?tab=readme-ov-file#mutually_exclusive or exactly_one_of).

db/schema.rb
```diff
@@ -1493,6 +1508,10 @@
     t.string "purification_method", default: ""
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.float "equivalent"
```
adambasha0, line 1513: equivalent & weight percentage belong to reactions_samples not the samples table. Although logic here might work, it is not consistent with core existing logic

db/migrate/20260219120000_add_reference_to_reactions_reactant_sbmm_samples.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddReferenceToReactionsReactantSbmmSamples < ActiveRecord::Migration[6.1]
+  def change
+    add_column :reactions_reactant_sbmm_samples, :reference, :boolean, default: false, null: false
```
adambasha0: this migration could be combined with CreateReactionsReactantSbmmSamples migration

lib/export/export_sdf.rb
```diff
@@ -87,6 +96,13 @@ def filter_with_permission_and_detail_level(sample)
       data.concat("\$\$\$\$\n")
     end
 
+    def sdf_molfile_for(sample)
+      return sample['molfile'] if sample['molfile'].present?
+      return EMPTY_MOLFILE if sample['source_type'] == 'sbmm'
```
adambasha0: why does an empty molfile needs to be added for sbmm? - should not mofile property be skipped for sbmm sample instead?

app/models/concerns/reactable.rb
```diff
@@ -34,14 +34,17 @@ def update_equivalent
...
 
-    ref_record = ReactionsSample.find_by(reaction_id: reaction_id, reference: true)
-    return if ref_record.nil? ||
-              sample&.sample_type == Sample::SAMPLE_TYPE_MIXTURE
+    # The reference material must always keep Eq = 1.
+    # Guard early to avoid transient recalculation paths during reaction save.
     if reference && gas_type != 'gas'
-      ## set equivalent to 1 for reference sample
-      update!(equivalent: 1.0)
+      update!(equivalent: 1.0) unless (equivalent.to_f - 1.0).abs < 1e-6
```
adambasha0, line 40: Why?

app/usecases/reactions/update_materials.rb
```diff
@@ -35,7 +35,28 @@ def is_split
   end
 
   def to_boolean(string)
-    !!"#{string}".match(/^(true|t|yes|y|1)$/i)
+    !!string.to_s.match(/^(true|t|yes|y|1)$/i)
+  end
+end
+
+# SBMM Sample Structure
+class OSbmmSample < OpenStruct
```
adambasha0, line 43: should be avoided if OSbmmSample is added as foreign key in reactions_samples using mutually_exclusive with sample_id

app/models/concerns/reactable.rb
TasnimMehzabin: This came as part of the copilot code review. It was suggested that in case the equivalent is already 1, then not to execute the update. So a float comparison has been added to make sure the equivalent is not already 1.

db/schema.rb
TasnimMehzabin: reactions_samples is built around Sample, so I preferred to keep a separate table for SBMM. Since mutually_exclusive in Grape is only for request validation, I think it will not be a small tweak at this situation, rather it will need a board refactor across the codebase.

PR #2870, opened 2026-01-15
chore: add converter config data output to ui_api

headri:


Gemfile.lock
```diff
@@ -408,8 +408,7 @@ GEM
     minitest (5.25.5)
     msgpack (1.8.0)
     multi_json (1.15.0)
-    multi_xml (0.8.1)
-      bigdecimal (>= 3.1, < 5)
+    multi_xml (0.6.0)
```
herrdivad, line 411: Why downgrading multi_xml and removing bigdecimal packages?

herrdivad (changes requested):
Check the Gemfile why downgrading package was done

Gemfile.lock
headri: I don't know. Gemfile.lock is automatically generated. Maybe a previous PR did not commit this file.

PR #2946, opened 2026-02-05
fix: ketcher api molfile fix with v2000-v3000 and solid surface SVGs

haditariq:
PR resolves:

- Opening solid supported sample and saving it crashes ELN
- Adding a solid supported sample into a reaction bugs 
- Macromoleucles Sample failing
- Bead and surface shapes same as ketcher-rails
- reaction svg atob error

ref: #2907

### DIFF
--- a/app/api/chemotion/molecule_api.rb
+++ b/app/api/chemotion/molecule_api.rb
@@ -188,7 +188,7 @@ module Chemotion
         decoupled = params[:decoupled]
         molecule = decoupled ? Molecule.find_or_create_dummy : Molecule.find_or_create_by_molfile(molfile)
         molecule = Molecule.find_or_create_dummy if molecule.blank?
-        ob = molecule&.ob_log 
+        ob = molecule&.ob_log
         svg_digest = "#{molecule.inchikey}#{Time.zone.now}"
 
         if svg.present? && svg.include?('epam-ketcher-ssc')


app/javascript/src/utilities/ketcherSurfaceChemistry/canvasOperations.js
```diff
@@ -894,13 +894,43 @@ const replaceAliasesWithIndexesAndCollectComponents = async (comboString) => {
...
-
-    const canvasDataMol = await editor.structureDef.editor.getMolfile('V2000');
+    let canvasDataMol = null;
+    let shouldSvg = true;
+    try {
+      canvasDataMol = await editor.structureDef.editor.getMolfile('V2000');
+      if (isMolfileEmpty(canvasDataMol)) {
+        console.log('onFinalCanvasSave: V2000 empty, trying V3000');
+        try {
+          canvasDataMol = await editor.structureDef.editor.getMolfile();
```
headri, line 916: Missing param, should be "getMolfile('V3000')"

docker-compose.dev.yml
```diff
@@ -112,20 +112,24 @@ services:
     working_dir: "/home/ubuntu/app"
     command: './run-js-dev.sh'
 
-  storybook:
```
headri, line 115: Why remove storybook?

app/javascript/src/utilities/ketcherSurfaceChemistry/canvasOperations.js
haditariq: If the parameter is missing, it generates a v2000 or v3000 ID by default, or crashes.

headri: But then you are doing the same thing as in the catch block below.

PR #2988, opened 2026-03-03
feat(search): add cell line

beque:
- Fix cell line routes after reloading page
- UI rework for detail view
- Add cell line to advanced and detail search
- Fix search for cell lines when press enter in search field (Added to PG Search)

To init PG Search
```
rake pg_search:multisearch:rebuild[CelllineSample]
rake pg_search:multisearch:rebuild[CelllineMaterial]
```


PiTrem (approved):

PiTrem:
- Collection-scoped search by cell-line name returns nothing. The new multisearchable on CelllineMaterial indexes the name, but by_collection_id scopes both sample and material rows through IN (SELECT cellline_sample_id FROM 
  collections_celllines …) — i.e. sample ids. Material searchable_ids come from a different (material) sequence, so every name match is filtered out before results are built. A collection-scoped "search all" by cell-line name finds
  nothing even when the cell line is in the collection. Needs a real material → sample → collection join. (app/usecases/search/all_elements_search.rb:36-48)

-  Cell-line result tab crashes past page 1. The frontend posts model_name: 'cell_lines'; by_ids.rb does .camelize.constantize → CellLines → NameError (the model is CelllineSample). Any cell-line result tab with more than one page
  crashes on First/Prev/Next. There's also an entity-name casing mismatch (CelllineSampleEntity vs CellLineSampleEntity) that would raise next. (app/usecases/search/by_ids.rb:36)

PR #3025, opened 2026-03-18
feat: reaction scheme UI enhancements and UI for reaction type "interaction"

### LINKED ISSUE #3008: Reaction - small UI and model changes
For being able to implement several other requirements, we need an adaptation of the reaction: 
(1) addition of pH vlaue as a property (with options to select >< =). Knowing that we offer that as a condition (which cannot be listed in the variations....)
(2) Option to edit the name of the reaction (if given, image contains only short label) in the header of the reaction -> removal of reaction name from the scheme tab. 
(3) Rearrangement of the properties given in the scheme tab to bring the vessel and reaction volume closer to the table
(4) please enlarge the field "type (name reaction ontology)"

Image

### LINKED ISSUE #3009: Reaction type
We need the introduction of a reaction type in addition to the already existing reaction scheme changes: 
(1) reaction type should focus on UI adaptations of the tab rather than calculations in the scheme
currently planned: a. standard (per default), b. interactions 
(2) the current scheme options are suboptions of the current reaction type "standard"
(3) would be good the select reaction type as scheme type with dropdown select fields (to avoid confusion with different field types)
(4) we need to shift the current field "status" to the same line as "reaction type" and "scheme type"

Image

### LINKED ISSUE #3010: UI for reaction type "interaction"
The reaction type "interaction" is based on the standard reaction scheme layout (including the changes described in issue #3008 and #3009). There is no subtype just the interaction reaction type. 
The following changes should apply when the reaction type "interaction" is selected: 
- The selection for scheme options (e.g. gas scheme as for the standard reaction type) disappear
- the reaction scheme image is replaced by the image of the product only. 
- "Starting materials" should be renamed to "Guest and host material"
- "Reactants" should be renamed to "Additives" 
-  The button "yield/conv" should be removed if possible
- the row "conditions" in the table should be removed/invisible
- the fields as given in the following screenshot should be removed

Image

TasnimMehzabin:
Closed https://github.com/ComPlat/chemotion_ELN/issues/3010

### DIFF
--- a/app/api/api.rb
+++ b/app/api/api.rb
@@ -129,7 +129,7 @@ class API < Grape::API
     ],
     'reactions' => %w[
       name short_label status conditions rxno content temperature duration
-      role purification tlc_solvents tlc_description rf_value dangerous_products
+      role reaction_type purification tlc_solvents tlc_description rf_value dangerous_products
       plain_text_description plain_text_observation
     ],
     'wellplates' => %w[name short_label readout_titles content plain_text_description],
--- a/app/api/chemotion/reaction_api.rb
+++ b/app/api/chemotion/reaction_api.rb
@@ -152,12 +152,15 @@ module Chemotion
         optional :purification, type: [String]
         optional :dangerous_products, type: [String]
         optional :conditions, type: String
+        optional :ph_operator, type: String
+        optional :ph_value, type: String
         optional :tlc_solvents, type: String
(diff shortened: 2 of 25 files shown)


TasnimMehzabin:
Duplicates https://github.com/ComPlat/chemotion_ELN/pull/3234, so closing this one

PR #3046, opened 2026-03-23
set role to null in copy reaction

headri:


PiTrem:
moved to 
655e0581df - fix: set role to null on copy reaction (#3051)

PR #3189, opened 2026-04-27
feat(sample): enhance melting/boiling point range input with decomposed toggle

### LINKED ISSUE #1151: Option decomposed for melting point
Sometimes you can not measure the melting point because the substance decomposes before. Then it would be good if you could select "Decomposed" instead of a temperature.

### LINKED ISSUE #1725: Melting point range or value
- I am an ELNer (ComPlat)

Currently, if we want to enter a melting range in the melting point tab, we have to write the two points with a space between them (e.g. we have to write 65 68 instead of 65-68) and then the system generates the ´-´ between the numbers itself. This is not user friedly and if someone does not know this and enters a range with the minus (65-68) in the field, this value does not get saved and also no error message appears - meaning that this data gets lost.  
It would be great if you could enable entering melting ranges with a minus directly, as this would prevent data loss and confusion. Furthermore, melting points are sometimes higher than we can measure so you should enable entering special characters into this field (for example, one should be able to enter >300 in the melting point tab)

TasnimMehzabin:
This PR enhances sample melting/boiling point range entry by adding richer parsing, comparison-style labels, and a “Decomposed” toggle for melting points.
- Introduced a new `TextRangeWithAddon` component with styling for alternative checkbox.
- Updated `SampleForm` to include a toggle for the "Decomposed" state of melting point, clearing the numeric range when checked.
- Integrated the new component into the sample form, enhancing user interaction for melting point input.
- Enhance `SampleForm` and `TextRangeWithAddon` for improved range handling

Closes https://github.com/ComPlat/chemotion_ELN/issues/1151 and https://github.com/ComPlat/chemotion_ELN/issues/1725

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -46,6 +46,7 @@
 @import "components/ReorderableList";
 @import "components/SampleName";
 @import "components/SearchInput";
+@import "components/TextRangeWithAddon";
 @import "components/SearchModal";
 @import "components/SearchResult";
 @import "components/SelectionActions";
--- a/app/assets/stylesheets/components/TextRangeWithAddon.scss
+++ b/app/assets/stylesheets/components/TextRangeWithAddon.scss
@@ -1,0 +1,8 @@
+.text-range-alternative-checkbox {
+  display: flex;
+  align-items: center;
+  gap: 0.375rem;
+  font-size: 0.85em;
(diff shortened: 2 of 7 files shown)


adambasha0 (approved):

PiTrem (approved):
CI green; clean melting/boiling point range UX with decomposed toggle. LGTM.

PR #3209, opened 2026-04-30
fix(report): escape HTML special chars in Quill delta converter

### LINKED ISSUE #3201: Report generation fails when the Reaction Description contains "<"
- I am an ELNer
- I am a system/ELN admin

# Expected Behavior

A report is generated when clicking "Generate Report" in a Reaction.

# Current Behavior

It gives an error "docx.htm (Site wasn't available)" (the error message depends on browser) when the Description contains a symbol "<" like "<a". It works when written as "< a".

# Failure Information 

## Steps to Reproduce

1. Write "<a" somewhere in the Description of a Reaction
2. Click Generate Report
3. Check browser download history to see failure

## Context

 - Windows / macOS
 - Chrome Browser / Safari
 - ELN v3.0.0

PiTrem:
## Summary

- `Reporter::Delta#buildDeltaOps` and `#html_with_tags_style` interpolated raw user text (`op["insert"]`) directly into HTML strings without escaping special characters.
- A Reaction Description containing `<a` (or any `<`) produced malformed HTML (e.g. `<span><a</span>`), which Sablon embedded in the `.docx` template, causing the report download to fail with a browser "site not available" error.
- Fix: apply `CGI.escapeHTML` to `op["insert"]` at both injection points so `<`, `>`, `&`, and `"` are safely encoded before insertion into generated markup.

> **Note:** This is a latent defect present since the Delta class was first introduced in Dec 2016 — not a regression from a recent change.

## Test plan

- [ ] Add `<a` (without a space) somewhere in a Reaction Description
- [ ] Click **Generate Report** — report should download successfully
- [ ] Verify the downloaded `.docx` renders `<a` as literal text (not a broken tag)
- [ ] Repeat with `>`, `&`, and `"` in the description to confirm all special chars are handled

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/reporter/delta.rb
+++ b/lib/reporter/delta.rb
@@ -1,3 +1,5 @@
+require 'cgi'
+
 module Reporter
   # Delta class to convert Quill delta format to HTML
   class Delta
@@ -113,7 +115,7 @@ module Reporter
     end
 
     def buildDeltaOps(op)
-      return op["insert"] if (!op["attributes"]) && (!(op && op["insert"].is_a?(Hash) && op["insert"]["image"]))
+      return CGI.escapeHTML(op["insert"]) if (!op["attributes"]) && (!(op && op["insert"].is_a?(Hash) && op["insert"]["image"]))
 
       styles = []
       tags = []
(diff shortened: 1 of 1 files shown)


haditariq (approved):
Tested with:
- Save reaction with HTML tags in description.
- Reload the reaction SVG with HTML tags in the description.

Both cases works well.

PR #3232, opened 2026-05-07
fix: prevent GC from deleting Tempfile during CDXML conversion for reports

PiTrem:
## Summary

- `get_cdxml_from_molfile` in `lib/chemotion/open_babel_service.rb` called `Tempfile.new(...).path` without storing the object reference, making the Tempfile immediately eligible for GC.
- Ruby's Tempfile finalizer deletes the underlying file when the object is collected. Under memory pressure (e.g. in a long-running `delayed_job` process), GC can fire between `c.convert` writing the output and `File.read(output)` consuming it, causing `Errno::ENOENT` and a failed report job.
- The bug has been present since commit `4125a4a` (Dec 15, 2016) when the method was first introduced.

## Fix

Store both `Tempfile` objects in local variables so they stay live through `File.read`, then explicitly `close!` them in an `ensure` block.

## Test plan

- [ ] Generate a DOCX report for a reaction with starting materials, reactants, and products — confirm it completes without error.
- [ ] Verify no temp files are leaked in `/tmp` after report generation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/chemotion/open_babel_service.rb
+++ b/lib/chemotion/open_babel_service.rb
@@ -498,8 +498,11 @@ M  END
     mf = mofile_clear_coord_bonds(molfile)
     mol = mf || molfile
     # `obabel -imol #{file_name} -ocdxml`
-    input = Tempfile.new(["input", ".mol"]).path
-    output = output_path || Tempfile.new(["output", ".mol"]).path
+    # Keep Tempfile references alive until after File.read; GC finalizers delete the files.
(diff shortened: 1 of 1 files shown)


PiTrem:
Superseded by #3253 (merged in 9957411103a0166886a164970844a9eb058fd395), which carries the identical `lib/chemotion/open_babel_service.rb` hunk as its first squashed commit alongside the broader Inkscape-pipeline Tempfile/Pathname fixes.

PR #3266, opened 2026-05-20
fix: ResizeObserver loop on CreatableSelect inputs (CAS, affiliations)

beque:
## Summary

Typing in a `react-select` `CreatableSelect` was triggering the
`ResizeObserver loop completed with undelivered notifications` error.
Root cause: each keystroke grew the input, which resized the parent
container, which fed back into react-select's internal observer.

This PR pins the surrounding layout in the two affected places so the
container no longer reflows on every keystroke.

## Changes

- **CAS field (`SampleDetails`)** — drop `flex-grow-1` from the wrapper
  and apply explicit `styles.input` (`minWidth: 200px; width: 100%`)
  on the `CreatableSelect`.
- **Affiliations settings table** — switch to `tableLayout: fixed` with
  an explicit `<colgroup>` so cell widths stay stable as the
  `CreatableSelect` inside grows.
- **MoleculeNameEntity** — expose `id` so the molecule moderator can
  identify records (previously missing, breaking edit/select flows).
- Minor cleanups in `SampleDetails.js`: `let` → `const`, drop a
  redundant fragment wrapper, destructure `selectedOption.value`, and
  trailing-semicolon fixes.

## Out of scope / to discuss

- `affiliation_api.rb` no longer sets an explicit `status 204` on
  `DELETE /affiliations/:id`. The frontend doesn't read the response,
  but this changes the HTTP contract from `204 No Content` to `200 OK`.
  Should we keep `status 204` to preserve the existing contract?

## Test plan

- [ ] Open a sample → CAS field: type characters, confirm no
      `ResizeObserver` warning in the browser console and the field
      still fills the row on wide viewports.
- [ ] User settings → Affiliations: add/edit a row, type into Country
      and Organization, confirm no console error and column widths
      stay stable.
- [ ] Molecule moderator: load the moderator UI and confirm molecule
      names render with their `id` (no missing-key React warnings).
- [ ] Delete an affiliation and confirm the row disappears (smoke test
      for the API change).

### DIFF
--- a/app/api/chemotion/affiliation_api.rb
+++ b/app/api/chemotion/affiliation_api.rb
@@ -97,9 +97,6 @@ module Chemotion
         u_affiliation = @affiliations.find(params[:id])
         u_affiliation.destroy!
         Affiliation.find_by(id: params[:id])&.destroy! if UserAffiliation.where(affiliation_id: params[:id]).empty?
-        status 204
-        body nil
-        return
       rescue ActiveRecord::RecordInvalid => e
         error!({ error: e.message }, 422)
       end
--- a/app/api/entities/molecule_name_entity.rb
+++ b/app/api/entities/molecule_name_entity.rb
@@ -3,6 +3,7 @@
 module Entities
   class MoleculeNameEntity < ApplicationEntity
     expose(
+      :id,
       :description,
       :name,
     )
--- a/app/javascript/src/apps/mydb/elements/details/samples/SampleDetails.js
+++ b/app/javascript/src/apps/mydb/elements/details/samples/SampleDetails.js
@@ -234,7 +234,7 @@ export default class SampleDetails extends React.Component {
 
   componentDidUpdate(prevProps) {
     const { sample } = this.props;
-    if (sample === prevProps.sample) { return };
+    if (sample === prevProps.sample) { return; }
 
     const smileReadonly = !(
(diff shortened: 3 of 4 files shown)


PiTrem (approved):
Approving — the ResizeObserver loop fixes are correct and well-scoped:

- **CAS field (SampleDetails)** — explicit `styles.input` width pin and dropped `flex-grow-1` stop the keystroke→resize feedback loop.
- **Affiliations table** — `tableLayout: fixed` + explicit `<colgroup>` keeps cell widths stable as the CreatableSelect grows.
- **`MoleculeNameEntity#id`** — exposing the PK is consistent with the rest of the API and fixes record identification in the molecule moderator.

**On the `affiliation_api.rb` `DELETE /affiliations/:id` change** (drops `status 204; body nil; return`, so the contract shifts from 204 No Content to 200 OK): this is intentionally **deferred** rather than treated as a blocker. The affiliation API surface is being reworked in #3272 (Refactor affiliations) and #3294 (Rework fetchers), which is where the contract should be settled. `UserSettingsFetcher.deleteAffiliation` doesn't read the response body, so there's no user-visible regression in the interim.

Remaining review notes (partial-Prettier JSX re-indent in `SampleDetails.js`, CAS-field width visual check) are non-blocking cosmetics. Merging.

PR #3284, opened 2026-05-29
fix: harden sample-task loading, dedup All-collection inserts, and reaction comment N+1

TasnimMehzabin:
- Guard sample task load when the API returns no array; harden SampleTasksFetcher
- Preload comments and use .size for comment_count on the reactions list (N+1 fix)
- Use find_or_create_by / skip duplicate append when the target collection is already “All” (reactions, samples, wellplates, screens, research plans, device descriptions)
- Default SBMM sample purity to 1 in buildEmpty (UI parity with Sample)

Closes https://github.com/ComPlat/chemotion/issues/507


PiTrem (approved):
Re-reviewed at ace0d94: both prior blockers resolved (CI green, rebased/mergeable), dedup refactored to find_or_create_by per suggestion, and the new assignSample param-shape fix verified against the endpoint. LGTM.

PR #3285, opened 2026-05-29
fix(UI): reset dropdown title for solvents selection

### LINKED ISSUE #3220: repetition of solvent in TLC control
- I am an ELNer

In TLC control (properties of a reaction), either writing solvents or choosing a mixture from the dropdown leads to repetition of the solvent mixture:

Image

Should only be mentioned once but the dropdown needs to be available. Could look like this (but allowing to see all content of the white field)

Image

TasnimMehzabin:
Closes https://github.com/ComPlat/chemotion_ELN/issues/3220


haditariq (approved):
Looks good.

PR #3288, opened 2026-05-29
fix(reaction): handle non-finite volume values in ReactionDetailsScheme

TasnimMehzabin:
Refactored volume parsing logic to ensure invalid, null, or empty values are coerced to undefined, preventing erroneous default values and maintaining expected UI behavior.

Closes https://github.com/ComPlat/chemotion/issues/508


alizaib1217 (changes requested):
hi @TasnimMehzabin
  
  Tested it — works as expected. The volume shows correctly now instead of n.d.. Just
  check @Copilot's comment on the wording, otherwise good to go.

alizaib1217 (approved):

PiTrem (approved):
Coerces volume to a finite Number before rendering; shows n.d. for null/empty/invalid. CI green.

PR #3294, opened 2026-06-01
refactor(js): unify fetchers on a single ChemotionApiClient

beque:
- Introduce ChemotionApiClient to unify all API calls
- Rework the Fetchers onto ChemotionApiClient (note: InfoSupportLinksFetcher, pulled in by a later main merge, still needs migrating)
- Standardize element Fetchers for handling Attachments and Annotations
- Add AnnotationsFetcher
- Remove the unused BaseFetcher
- Expand FetcherHelper to remove duplicated key-casing logic and drop the humps dependency

PR #3295, opened 2026-06-02
feat(lcms): on-demand MS-page extraction for the LC/MS spectra layout

Nicolass67:
# Description
This PR aligns the chemotion ELN with the new LC/MS Layout in react-spectra-editor. Here we wire up JCamp handling, on-demand MS “page” extraction, and API behavior so OpenLab and ChemStation–style LCMS data behave correctly in analyses.

## What we changed

- POST …/lcms_page (attachment API)
  - Keeps the endpoint that returns a single ##PAGE= block (mass spectrum) for a given retention time, with optional polarity (positive / negative, or no filter).
  - Small doc/parameter text cleanup; access control unchanged in intent.

- LC/MS JCamp pipeline (AttachmentJcampAasm / AttachmentJcampProcess)
  - Peak generation rules for LCMS: clearer require_peaks_generation? logic (explicit booleans) so UV-Vis “raw” LCMS JDX and queueing cases don’t trigger the wrong path.
  - generate_att: flatter branching for peaks vs edit and LCMS file name patterns (e.g. uvvis peak/edit, TIC/mz/uvvis JDX) so child attachments get the right AASM state (set_edited, set_force_peaked, or set_non_jcamp for raw UV-Vis when appropriate).
  
- On-demand page slice (Lcms::PageExtractor, Lcms::PageIndexer)
  - Safer attachment_path and cache keys; more robust byte-range reads (avoid reading past the intended block when length is not positive).
  - Readability and RuboCop-driven structure only where needed.

## Tests
Specs updated for attachment API, page extractor/indexer


PiTrem (approved):
Re-reviewed at bc229d13: #3294 fetcher re-port landed, combined-save error propagation fixed, CI green. Remaining items (peak-file deletion guard, polarity/predictions robustness) tracked in a follow-up issue.

PR #3296, opened 2026-06-02
fix(reaction UI): allow decimals in coefficients

### LINKED ISSUE #3199: Allow non-integer reaction coefficients
- I am an ELNer / system/ELN admin

# Description
1. I suggest allowing non-integer values in the Coef fields of reactions
2. Non-integer coefficients may sound weird for small molecules, but it is quite common for solid materials. e.g., when you want to keep the important part of the formula as 1 like A1B1.5 or when you have solid solutions like A0.2B0.8C1. Since the coefficients are used in yield calculations, it would be useful to enter non-integer values there.

haditariq:
Reaction coefficients were silently reset to 1 whenever a decimal value was entered. The database already stores coefficients as floats.

-----

- [ ] added code is linted


alizaib1217 (approved):
Everything looks good to me. I want to ask: I was unable to delete all numbers using the backspace key on the keyboard. For the last digit, I have to select and delete to type 0.

PR #3298, opened 2026-06-03
fix: prevent stale research plan list when navigating to shared collections

### LINKED ISSUE #2750: Element not updated in sync collection
- I am a developer

# Expected Behavior

When I switch to a synchronized collection, the item list should be updated accordingly.

# Current Behavior

When I open a synced collection, the lists of items are not updated.

In my case: 
 I opened the list of research plans. Then I switched to a synced collection. However, the list of research plans did not change. Only when I clicked on the Research Plan tab again was the list updated.

## Context

 - OS: All
 - Chrome Browser 
 - ELN version, or branch, or commit: Version 2.1.1 and main Branch

### LINKED ISSUE #3198: Research Plan view not updated when browsing synchronized collections
- I am an ELNer
- I am a system/ELN admin

# Expected Behavior

When opening a synchronized collection, research plans in the collection should be listed, and the count of items should be displayed, just like with reactions and samples.

# Current Behavior

The research plan list and count stay as in the previous collection, and they get updated only after switching to another element tab and going back to the research plan tab. It was reported in #2750 before and persists in v3.0.0.

# Failure Information 

## Steps to Reproduce

1. have a synchronized collection from another user
2. create a research plan
3. open another collection and create a research plan
4. go back to the synchronized collection
5. research plan list and count are not updated

## Context

 - OS: Windows / macOS
 - Chrome Browser 
 - ELN v3.0.0

haditariq:
Closes #3198, related to #2750.

## What was broken

When navigating to a shared collection, `collectionShow` could pass`undefined` to `UIActions.selectCollection` if the collection fetch returned no data (e.g. 404). This caused a silent `TypeError` in `UIStore.handleSelectCollection`, which blocked all element fetches and left the research plan list stale.

Note: the core fetch regression (research plans being skipped entirely for shared collections) was already fixed in #2783 by removing the old `!isSync` guard. This PR closes the remaining edge case.

## Steps to reproduce

1. User A shares a collection (with research plans) with User B
2. User B opens the shared collection → sees plans
3. User B navigates to their own collection
4. User B returns to the shared collection → list was stale (now fixed)

#### ** test case is inconsistent

-------

- [ ] added code is linted

### DIFF
--- a/app/javascript/src/stores/alt/stores/UIStore.js
+++ b/app/javascript/src/stores/alt/stores/UIStore.js
@@ -360,12 +360,17 @@ class UIStore {
   }
 
   handleSelectCollection(collection, hasChanged = false) {
+    if (!collection) return;
+
     const state = this.state;
     const { filterCreatedAt, fromDate, toDate, userLabel, productOnly } = state;
 
     if (!hasChanged) {
       hasChanged = !state.currentCollection;
       hasChanged = hasChanged || state.currentCollection.id != collection.id;
+      const wasShared = !!(state.currentCollection?.collection_share_id);
+      const isShared = !!(collection.collection_share_id);
+      hasChanged = hasChanged || wasShared !== isShared;
       hasChanged = hasChanged || state.currentSearchSelection != null;
       hasChanged = hasChanged || state.currentSearchByID != null;
     }
--- a/app/javascript/src/utilities/routesUtils.js
+++ b/app/javascript/src/utilities/routesUtils.js
@@ -22,6 +22,8 @@ const collectionShow = (e) => {
   const collectionPromise = CollectionsFetcher.fetchByCollectionId(collectionId);
 
   collectionPromise.then((collection) => {
+    if (!collection) return;
+
     if (currentSearchSelection) {
       UIActions.selectCollectionWithoutUpdating(collection);
       ElementActions.fetchBasedOnSearchSelectionAndCollection({


alizaib1217 (commented):
Hi, I am unable to reproduce this on my instance. 
As the main is broken at the moment. I am testing it on my own instance, and it seems to be working there as well. 
https://az-combinesamples.deploy.chemdev.scc.kit.edu/

haditariq:
## Why the test case is inconsistent

The primary root cause of #3198 and #2750 was the `!isSync` guard in `UIStore.handleSelectCollection`, which explicitly blocked research plans (and cell lines, device descriptions) from being fetched for sync/shared collections. Samples and reactions had no such guard — which is exactly why those element types updated correctly but research plans did not.

This guard was removed in #2783, which is **not included in v3.0.0** (the version the reporter tested against). The current `main` branch already has that fix, so the original bug is no longer reproducible on main.

This PR adds two defensive guards on top of that:
1. `if (!collection) return` in both `routesUtils.js` and `UIStore` — prevents a `TypeError` when `fetchByCollectionId` returns `undefined` (e.g. a 404 response), which would have silently blocked all element fetches.
2. `wasShared !== isShared` check — ensures a re-fetch is triggered when transitioning between own and shared collections even if their database IDs match.

Both changes are correct. The difficulty reproducing the issue is because `main` already carries the #2783 fix — not because the guards are unnecessary.

alizaib1217 (approved):

PiTrem:
Closing as superseded, thanks @haditariq.

The stale research-plan list reported in #3198 / #2750 was caused by the `!isSync` guard in `UIStore.handleSelectCollection`, which skipped the research-plan fetch for synced collections. #2783 removed it, so on `main` the fetch is unconditional (shipping in v4.0.0).

What remains here:
- the `wasShared !== isShared` check can't trigger: a collection id is either own or shared-with-me for a given user, so when the id is unchanged the flag can't flip, and when it changes `hasChanged` is already true;
- the `collectionShow` guard covers a real edge case (a 404/403 on `GET /collections/:id` resolves to `undefined`), but returning silently leaves the previous collection's list on screen under the new URL. That's tracked separately, to be handled with a notification and redirect.

PR #3300, opened 2026-06-08
feat(reactions): auto-link -v<id> attachments to reaction variations

### LINKED ISSUE #2845: automated assigment of incoming data to experiments
It would be a great enhancement if the ELN could automatically assign analytical data that is incoming (inbox) to the experiment it belongs to. 
- The assignment should work on the basis of the name of the data file (the short label is usually used as name of the dataset). The name should allow the assignment to a reaction or a sample or any other item as long as the name is unique in the ELN. E.g. NJ-1 initiates the assignment to the sample NJ-1 while a data file with the label NJ-R1 is assigned to the reaction NJ-R1 and so on. 
- For reactions, a special identification and assignment should be enabled: as reactions may have variations, labels such as NJ-R1-v1 should be assigned to the variation 1 of reaction NJ-R1. 
The variations' name space ranges from 1 to any integer or from 1.1 to any number x.x (e.g. giving either reaction labels such as NJ-R1-v15 or NJ-R1-v15.9).
- The user should be able to enable the assignment in the users settings and to disable the automated assignment.

headri:
After uploading an attachment whose filename contains `-v<id>` to a
  reaction analysis, the variation link is now directly written to the DB.

To represent that in the frontend without a page reload:
  The container API now returns the reaction's updated `variations` array
  alongside the container in the PUT /containers/container response.
  The frontend propagates this through ContainerDatasets → ContainerComponent
  → ReactionDetailsContainers, where it is applied to the reaction state
  before triggering a re-render of the variations tab.


StarmanMartin (approved):

JanCBrammer (commented):
The feature does not work as intended. I tried the following steps in the UI.

1. Upload an attachment `attachment-v1.pdf`, using the `Analysis from upload` button in the `Reaction Analysis` tab, which is nested in the `Analyses` tab of the reaction.
2. Once the upload is finished, I navigate to the `Variations` tab of the reaction and expect the attachment to be linked to the variation whose id ends with `1`. This does not happen.

Saving the reaction between steps 1 and 2 does not help either. @headri, am I missing something?

JanCBrammer (changes requested):
According to @headri, the analysis needs to be added via the `Add Analysis` button, not the `Analysis from upload` button. In addition, the reaction needs to be saved between steps 1. and 2.

By following @headri's instructions, I got the feature to work. My suggestion is to document that the feature is intended to be used as per these instructions, or make sure the feature also works via the route I had tried initially (since it's reasonable to assume that users will try that as well).

JanCBrammer (approved):
I got the feature to work now via both, `Add Analysis` and `Analysis from upload`. Consider making the [docs](https://chemotion.net/docs/eln/ui/inbox?_highlight=variation#reaction-variations-link) more specific: what are the explicit steps to make this feature work?

PiTrem (approved):
Both prior Blocking bugs (inbox regex mismatch, nil-variations crash) are fixed with dedicated regression tests; N+1 batching, redundant reload, and duck-typing consistency also resolved. Remaining pronto/rubocop warning investigated separately and not correctness-impacting. Approving.

PR #3302, opened 2026-06-09
feat(mixtures): support multiple components sharing the same molecule

TasnimMehzabin:
- Updated the component creation process to handle multiple components sharing the same molecule, ensuring each is persisted as a separate row. [Demo](https://drive.google.com/file/d/1LsF3AJzyu6J7_GW6cLu7YBbDqh9DtDiL/view?usp=sharing)
- Enhanced the deletion logic to remove components based on their IDs rather than molecule IDs, allowing for better management of newly added components.
- Improved the handling of component updates in the sample's state, ensuring that the API response is utilized to refresh component data accurately.
- Added checks to prevent duplicate component additions during drag-and-drop operations in the UI.
- Fixed an issue where editing an existing component in the structure editor incorrectly created a new component instead of updating the existing one. [Demo](https://drive.google.com/file/d/1WjzjfI0XSo6tx-FTkyoKW-OnVm6QLRKu/view)

This refactor aims to enhance the integrity and performance of component management within samples, particularly for mixtures.

Closes https://github.com/ComPlat/chemotion/issues/515

### DIFF
--- a/app/usecases/components/delete_removed_components.rb
+++ b/app/usecases/components/delete_removed_components.rb
@@ -7,7 +7,7 @@ module Usecases
     class DeleteRemovedComponents
       def initialize(sample_id, components_params)
         @sample_id = sample_id
-        @components_params = components_params
+        @components_params = components_params.map(&:with_indifferent_access)
       end
 
       def execute!
@@ -16,7 +16,7 @@ module Usecases
         # ignored here; they are inserted by Usecases::Components::Create.
         ids_to_keep = @components_params.filter_map do |cp|
           Integer(cp[:id], exception: false)
-        end.compact
+        end
 
         scope = Component.where(sample_id: @sample_id)
         scope = scope.where.not(id: ids_to_keep) if ids_to_keep.any?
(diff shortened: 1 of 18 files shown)


app/usecases/components/delete_removed_components.rb
```diff
@@ -11,15 +11,17 @@ def initialize(sample_id, components_params)
...
-        molecule_ids_to_keep = @components_params.filter_map do |cp|
-          cp.dig(:component_properties, :molecule_id)&.to_i
-        end
+        # Collect existing component ids to keep. Non-integer ids (e.g. client-side
+        # placeholders like 'new_1') correspond to brand-new components and are
+        # ignored here; they are inserted by Usecases::Components::Create.
+        ids_to_keep = @components_params.filter_map do |cp|
+          cp = cp.is_a?(HashWithIndifferentAccess) ? cp : cp.to_h.with_indifferent_access
+          Integer(cp[:id], exception: false)
+        end.compact
```
adambasha0: filter_map already removes any existing nil, compact is redundant here

adambasha0 (changes requested):
I am missing backend validation. Might be suitable for a separate PR

adambasha0 (approved):

PiTrem (approved):
Re-verified both previously-blocking findings against the current head (2359c872): the same-molecule structure-editor dedup is working as designed (pinned by regression tests), and the components nil-vs-[] serialization gap is fixed. No blocking issues remain; merging with the pronto/rubocop warnings accepted per maintainer instruction.

PR #3303, opened 2026-06-09
fix(js): replace removed default Immutable export with named imports

PiTrem:
## Summary

- Immutable v5.0.0 (introduced by #2994) dropped the default export, so `import Immutable from 'immutable'` no longer resolves. Without this fix the bundle compiles via webpack's CJS interop fallback but `Immutable.Map`/`List`/`Set` references are runtime-undefined on every screen that uses them.
- Converted the 20 affected files to named imports (`{ List }`, `{ Map }`, `{ Set }`, `{ List, Map }`, or `{ List, Set }` depending on usage) and rewrote `Immutable.X` callsites accordingly.
- No behavior change; same collection types, just imported the v5-supported way.

## Test plan

- [x] `yarn webpack` (via the dev `webpacker` service) compiles successfully — no module-resolution errors.
- [x] Rails dev server boots and serves the home page.
- [x] Login with a test account works.
- [x] Click through each affected detail tab to confirm runtime behavior: Sample, Reaction, Wellplate, Screen, ResearchPlan, SBM Sample, DeviceDescription, Literature (sample + reaction + research plan tabs), Collection LiteratureModal, ElementsList, GenericElDetails.

refs: #2994

### DIFF
--- a/app/javascript/src/apps/mydb/collections/LiteratureModal.js
+++ b/app/javascript/src/apps/mydb/collections/LiteratureModal.js
@@ -9,7 +9,7 @@ import {
   OverlayTrigger,
   Tooltip
 } from 'react-bootstrap';
-import Immutable from 'immutable';
+import { Map } from 'immutable';
 import { uniqBy } from 'lodash';
 import AppModal from 'src/components/common/AppModal';
 import {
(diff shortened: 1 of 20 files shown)


TasnimMehzabin (approved):
Looks good. Tested for the following:
- [x] Sample
- [x] Reaction
- [x] Wellplate
- [x] Screen
- [x] ResearchPlan
- [x] SBM Sample
- [x] DeviceDescription
- [x] Literature (sample + reaction + research plan tabs)
- [x] Collection LiteratureModal
- [x] ElementsList
- [x] GenericElDetails

PR #3305, opened 2026-06-10
fix(ketcher): guard iframe document access against SecurityError

haditariq:
### Problem
attachClickListeners accessed contentWindow.document at the top level of the function. Optional chaining guards against null but not DOMException(SecurityError), which the browser throws when the iframe is mid-navigation.

### Changes
Moved all contentWindow.document access inside a load event listener (setup) so the document is only read after the iframe has fully loaded
Wrapped every contentWindow.document access in a try/catch that absorbs SecurityError and re-throws anything else
Declared observer and debounceAttach in outer scope so the cleanup closure can still disconnect/clear them
Removed the redundant readyState === 'complete' check inside the timeout — unnecessary inside the load handler
Wrapped cleanup's DOM access in its own try/catch since the iframe may have navigated away by cleanup time

---------------

- [ ] added code is linted

### DIFF
--- a/app/javascript/src/utilities/ketcherSurfaceChemistry/DomHandeling.js
+++ b/app/javascript/src/utilities/ketcherSurfaceChemistry/DomHandeling.js
@@ -311,7 +311,11 @@ const attachClickListeners = (iframeRef, buttonEvents) => {
 
   const iframeEl = iframeRef?.current;
   if (iframeEl) {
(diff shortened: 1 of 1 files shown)


PiTrem (approved):
Well-scoped iframe SecurityError hardening that also fixes several latent listener-cleanup/attachment bugs. CI green; reviewed at 4e7ab513.

PR #3306, opened 2026-06-10
feat(UI): add numeric inputs for moisture and particle size, and dropdown for physical state in SampleForm

### LINKED ISSUE #3242: new property for sample
the property "moisture" is needed as an additional description for the sample in the properties tab.  It should be given in %

TasnimMehzabin:
- Introduced `numericInputWithAddon` method for handling numeric inputs with an addon text.
- Added `physicalStateInput` method to allow selection of physical states (solid, liquid, gas).
- Integrated new input fields for moisture, particle size, and physical state into the SampleForm layout.

Closes https://github.com/ComPlat/chemotion_ELN/issues/3242

### DIFF
--- a/app/javascript/src/apps/mydb/elements/details/samples/SampleDetails.js
+++ b/app/javascript/src/apps/mydb/elements/details/samples/SampleDetails.js
@@ -937,6 +937,7 @@ export default class SampleDetails extends React.Component {
       rfsovents,
       supplier,
       private_notes,
+      color,
       moisture,
       particle_size,
       physical_state,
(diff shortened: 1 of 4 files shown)


alizaib1217 (changes requested):
hi @TasnimMehzabin

  1. Register the new keys for versioning in app/services/versioning/serializers/sample_serializer.rb
  (xref block) — otherwise moisture/particle_size/physical_state won't show in history or be revertible
  like the other xref fields. This is the main one.
  2. (Optional) Add min/max bounds (moisture 0–100, particle size ≥0) to guard against
  negative/out-of-range values.

  Also have a look at the Copilot review — the step/decimal point there isn't actually a problem (12.5
  saves fine in this form), and the NaN case can't happen since type="number" returns '' for invalid
  input, which the existing guard handles.

alizaib1217:
image


 @TasnimMehzabin

Fields show up twice, and design is not correct — the second set (below Description) comes from customizableField() in
  SampleDetails.js.

alizaib1217 (approved):
Hi @TasnimMehzabin, LGTM.

PR #3313, opened 2026-06-12
feat: unified attachments tab across element types (excludes Vessel)                                                        

### LINKED ISSUE #3312: feat: add direct attachments support to Sample, Reaction, Screen, CellLine, and Vessel
- I am an ELNer / developer

# Description
Add a dedicated Attachments tab to the detail views of Sample, Reaction, Screen, Cell Line, and Vessel elements. This tab allows users to directly upload, manage, annotate, and delete files attached to these elements — without going through the Container/Analysis hierarchy.

### WHY
Currently, direct file attachment (outside of analysis containers) is only available on ResearchPlan and Container elements. Researchers frequently need to attach supplementary files directly to a Sample or Reaction without creating a formal analysis entry.

### LINKED ISSUE #3314: Images/media inserted into Quill rich-text editor are silently discarded
- I am an ELNer
- I am a developer

# Expected Behavior

When a user pastes or drops an image into a Quill rich-text editor field, the image should be uploaded as an Attachment linked to the parent element, rendered inline in the editor, and included in exports.

# Current Behavior

Images are silently stripped before the Quill editor initialises.

haditariq:
Helpdesk#[1628 ](https://helpdesk.nfdi4chem.de/#ticket/zoom/1628)

  ## Summary                                            
Introduces a single generic Attachments tab for Sample, Reaction, Wellplate, Screen, ResearchPlan, CellLineSample,
DeviceDescription, and SequenceBasedMacromoleculeSample. Standardises the backend attachable API with a type
whitelist and uniform policy enforcement. Vessel keeps its existing attachment flow (out of scope).                                         
                                                                                                                                
  ## What changed                                                     
                                                                                                                                
  - One unified Attachments tab across Sample, Reaction, Wellplate, Screen, Research Plan, Cell Line, Device Description, and Sequence-Based Macromolecule Sample — [`AttachmentTab.js`](app/javascript/src/apps/mydb/elements/details/attachmentTab/AttachmentTab.js)                            
  - Vessel keeps its existing attachment flow — [`VesselDetails.js`](app/javascript/src/apps/mydb/elements/details/vessels/VesselDetails.js) untouched                        
  - Standardised attachment API so every element type behaves the same way — [`attachable_api.rb#L7`](app/api/chemotion/attachable_api.rb#L7)                                                              
  - Consistent permission checks for uploads and deletes across all element types — [`attachable_api.rb#L27`](app/api/chemotion/attachable_api.rb#L27)                                                            
  - Attachments correctly cleared when a record is removed, instead of leaving orphans — [`wellplate.rb#L81`](app/models/wellplate.rb#L81), [`research_plan.rb#L61`](app/models/research_plan.rb#L61)                  
  - Deleting a file from one record can no longer affect files on another record — [`attachable_api.rb#L66`](app/api/chemotion/attachable_api.rb#L66)                                                            
  - Invalid or unknown element types now rejected up front instead of silent success — [`attachable_api.rb#L29`](app/api/chemotion/attachable_api.rb#L29)                                                            
  - Files preserved correctly when a record's owner is anonymised — [`reaction_entity.rb`](app/api/entities/reaction_entity.rb)
  - Fixed double-save / disappearing / failed-delete bugs across Sample, Cell Line, and Reaction — [`SamplesFetcher.js`](app/javascript/src/fetchers/SamplesFetcher.js), [`CellLinesFetcher.js`](app/javascript/src/fetchers/CellLinesFetcher.js), [`ReactionsFetcher.js`](app/javascript/src/fetchers/ReactionsFetcher.js)                                                      
  - Duplicated per-element attachment code removed — old `ResearchPlanDetailsAttachments`, `GenericDetailsAttachments`, and two `AttachmentForm.js` files deleted; consolidated into [`AttachmentTab.js`](app/javascript/src/apps/mydb/elements/details/attachmentTab/AttachmentTab.js), [`attachmentUtils.js`](app/javascript/src/utilities/attachmentUtils.js)                                                  
                                                                                                                                
  ## Not in this PR (follow-up tickets if required)                                                                                         
  - Vessel migration to the unified endpoint                                                                                                                                                    
  - CellLine MobX / alt-Flux split reconciliation                                                                               

----------------

- [ ] added code is linted


haditariq:
### Code review

Found 2 issues:

1. `del_files` deletion is not scoped to the authorized element — a user authorized to update element A can orphan attachments belonging to element B of the same type by supplying their IDs in `del_files`. The `WHERE` clause filters by `attachable_type` but not `attachable_id`. Fix: add `.where(attachable_id: attachable_id)` to the deletion query.

https://github.com/ComPlat/chemotion_ELN/blob/3fe87d7c470845d63891075244b596b3241f9d3e/app/api/chemotion/attachable_api.rb#L81-L86

2. `handleAttachmentUndoDelete` does not set `changed = true` in any of the five detail components (Sample, Reaction, Screen, CellLine, Vessel), while `handleAttachmentDelete` does. This creates an asymmetry: after undoing a deletion the Save button is never enabled, so the restoration is visually confirmed but never persisted to the backend. Shown here for Sample; the same pattern is present in ReactionDetails, ScreenDetails, CellLineDetails, and VesselDetails.

https://github.com/ComPlat/chemotion_ELN/blob/3fe87d7c470845d63891075244b596b3241f9d3e/app/javascript/src/apps/mydb/elements/details/samples/SampleDetails.js#L505-L514

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

haditariq:
Addressed the Copilot review comments:

**Fixed:**
- Added unique index on `vessels.attachment_ref_id` in the migration — it is used as a surrogate FK and looked up on every auth check, so the index was missing.
- Fixed dirty tracking in `Vessel#refresh_attachment_ref_id` — the plain attribute assignment left the model marked as changed after create, which could trigger spurious UPDATEs. Now uses `write_attribute` + `clear_attribute_changes`.

**Not fixed — reasoning:**

- **schema.rb missing rdkit extension/function**: The project conditionally enables rdkit based on whether the PG cartridge is installed (`WARN PG cartridge extension not available - feature disabled` in app logs). The schema was generated on a dev machine without rdkit, which is expected. The correct long-term fix is either regenerating schema.rb on a CI host with rdkit or switching to `structure.sql`. That is out of scope for this PR and pre-dates it.

- **Unknown `attachable_type` bypasses authorization**: This is intentional. An unknown type cannot match any element policy or database record, so there is no protected data reachable through it. The existing test `"passes through without a 401 (no auth check for unknown types)"` explicitly documents this contract. Blocking unknown types would be a separate change.

- **`Rails/InverseOf` on `vessel.rb`**: `inverse_of: false` is already set on the association. This is the correct value — a polymorphic scope-based association with a surrogate FK cannot have a meaningful inverse, and `inverse_of: false` tells ActiveRecord not to attempt auto-detection. The cop is a false positive here.

haditariq:
### Code review

Found 4 issues:

1. `attachable_id` param type changed from `Integer` to `String` — the `attachments.attachable_id` column is an integer FK; passing a raw string into `Attachment.where(attachable_id: attachable_id)` and `Attachment.new(attachable_id: attachable_id)` may prevent the DB from using the integer index and causes implicit type coercion with no documented reason for the loosening.

https://github.com/ComPlat/chemotion_ELN/blob/0030780aca37bb0b469b2e4dcedfbce8c6f54104/app/api/chemotion/attachable_api.rb#L24-L27

2. `find_by` returning nil passes `nil` into `ElementPolicy#update?` — this works correctly today (policy returns `false` for nil, yielding a 401), but collapses "not found" and "unauthorized" into the same response code across all 9 element types; a nil guard before the policy call would make the intent explicit and protect against policy implementations that raise on nil.

https://github.com/ComPlat/chemotion_ELN/blob/0030780aca37bb0b469b2e4dcedfbce8c6f54104/app/api/chemotion/attachable_api.rb#L30-L33

3. Wellplate `create()` reconstructs the model from the original POST response instead of re-fetching after attachments are uploaded — unlike `ReactionsFetcher` and `SamplesFetcher` which both call `this.fetchById(id)` after `updateAttachables()`, `WellplatesFetcher.create()` calls `wellplateElement(json, id)` with the pre-upload `json`, leaving the UI with an empty attachments array until the user reopens the element.

https://github.com/ComPlat/chemotion_ELN/blob/0030780aca37bb0b469b2e4dcedfbce8c6f54104/app/javascript/src/fetchers/WellplatesFetcher.js#L28-L42

4. `showImportConfirm` is initialized as `[]` (array) in the constructor but used as a plain object keyed by attachment id; newly dropped attachments are never added to the map (no `componentDidUpdate` sync), so `showImportConfirm[newAttachment.id]` is always `undefined` and the import confirm overlay never shows for files dropped after mount.

https://github.com/ComPlat/chemotion_ELN/blob/0030780aca37bb0b469b2e4dcedfbce8c6f54104/app/javascript/src/apps/mydb/elements/details/attachmentTab/AttachmentTab.js#L36-L63

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3340, opened 2026-06-26
feat(import): CAS-only import, per-cell coercion, and a correctable report

adambasha0:
**per-value handling and a correctable import report**

**Problem:** 

- Samples can not be imported using only CAS value, when molfile, smiles checks are missing
- A sample import could lose data in two ways:
  -  One value the database refused took the whole row with it
  - and a value itcould not store was silently replaced instead: a blank or non-numeric purity became 0.0, recording a sample as 0% pure, and a density given without a unit was discarded. 
  - Reported row numbers were computed as index + 2 over a list with unusable rows already filtered out, so they named rows that had imported perfectly well while the broken ones went unmentioned.

**Solution:** 

**1. Per-value handling (Import::ValueCoercion)**
- an unusable cell now costs that field, not the row, and every value dropped or changed is recorded with its original text
- reversed ranges are normalised rather than failing the row, and "120-125" no longer reads its hyphen as a minus sign
- purity accepts a percentage; a blank cell keeps the column default
- density accepts a unit-less number, since the column stores g/mL; units it cannot convert are dropped
- amount units are matched case-insensitively but stored canonically
- shared with the SDF import, so one file imports the same way through either path

**2. Reporting**
- reported row numbers are the row numbers in the user's file
- new import report workbook: a verdict per row, a marking per unusable
  cell, a legend, and sample ids linked to the samples they created on
  this instance
- the report replaces the uploaded file in the Inbox; the notification
  links to it and refreshes the Inbox, the element list and the
  collection tree
- the message is a headline plus one bullet per issue, with long row
  lists truncated
- a file that imported every row but lost values reports as a warning,
  not a success

**3. Correcting and importing again**
- the report carries a sheet holding only the rows that still need
  something, named so that the importer reads it: nothing has to be
  deleted by hand
- a row with a sample id updates that sample instead of creating one,
  and an empty cell means "leave this field alone"
- only samples the importing user may edit can be named (ElementPolicy);
  "not yours" and "does not exist" give the same answer, so the sheet
  cannot be used to probe for samples

4. **Import policy**

- no structure, no CAS and a blank decoupled flag imports as decoupled
  rather than being dropped, which is what an inventory list without
  structures needs
- the same row with decoupled set to no is refused, and the report says
  what to change

**5. Frontend**
- the import dialog resets when an import starts, so the next one does
  not open on the last file, checkboxes and column mapping
- cancelling on the column-mapping step asks before discarding the
  mapping, as the validation step already did
- the template list is no longer clipped by the dialog it opens in
- notification text no longer runs past the edge of the toast
- Force rerender of collection tree sample elements on import
- Enhance notification info
_______________________________________________________________________________________________________________________________

- [x] added code is linted

- [x] testing coverage improvement is improved.


TasnimMehzabin (approved):

PiTrem (approved):
Re-reviewed at e0b482d6c5. All five blocking findings from the last review are fixed, main is merged in (incl. #3486's molfile rstrip fix), and the suite is green: 528 examples, 0 failures. pronto/rubocop stays red at 10 warnings — unchanged from before this round and all pre-existing complexity in create_samples and friends, which #3524 restructures.

PR #3341, opened 2026-06-26
feat(UI): enhance loading value handling and add tests

### LINKED ISSUE #3157: polymer section/elemental composition loading - external estimation
- I am an ELNer

# Expected Behavior

For polymers where I define a loading (mol/g) and check the box "loading according to: external estimation" I expect the estimated loading to be shown.

# Current Behavior

When I entered data for elemental analysis, the loading according to this was calculated and is shown for checking the corresponding box . When I check "external estimation" the loading is set to n.d. When I click into the field I see the originally entered value but cannot save it.

# Failure Information 
In case I choose external estimation I want this to be shown and to be used for further calculation.

## Steps to Reproduce
See video
[2026-04-21 08-30-36.zip](https://github.com/user-attachments/files/26922121/2026-04-21.08-30-36.zip)

## Context

 - ELN version v3.1.0-patch1

TasnimMehzabin:
- Introduced a `parseLoadingValue` function to normalize loading values, ensuring proper handling of null, empty, and non-finite inputs.
- Updated the `handleCustomInfoNumericChanged` method to propagate changes to loading values correctly.
- Modified the rendering logic to display the parsed loading value in the component.
- Added comprehensive tests for the `PolymerSection` component to validate loading value behavior and ensure correct propagation of changes.

Closes https://github.com/ComPlat/chemotion_ELN/issues/3157

### DIFF
--- a/app/javascript/src/apps/mydb/elements/details/samples/propertiesTab/PolymerSection.js
+++ b/app/javascript/src/apps/mydb/elements/details/samples/propertiesTab/PolymerSection.js
@@ -6,6 +6,14 @@ import ElementalCompositionGroup from 'src/apps/mydb/elements/details/samples/pr
 import NotificationActions from 'src/stores/alt/actions/NotificationActions'
 import { Select } from 'src/components/common/Select';
 
+const parseLoadingValue = (rawValue) => {
+  if (rawValue == null) return undefined;
+  if (typeof rawValue === 'string' && rawValue.trim() === '') return undefined;
+
+  const parsed = Number(rawValue);
+  return Number.isFinite(parsed) ? parsed : undefined;
+};
+
 export default class PolymerSection extends React.Component {
   handleCustomInfoNumericChanged(e, name, residue, sample) {
     const { handleSampleChanged, handleAmountChanged } = this.props;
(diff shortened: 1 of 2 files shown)


haditariq:
### Code review

Found 1 issue:

1. `handleSampleChanged(sample)` is added to the `loading` branch of `handleCustomInfoNumericChanged`, creating a second `setState({ sample })` call on top of what `handleAmountChanged` already triggers. In `SampleDetails`, `handleAmountChanged` calls `setState({ sample })` directly — it does not chain to `handleSampleChanged`. Adding an explicit `handleSampleChanged` call immediately after means every keystroke in the loading field now dispatches two `setState` calls with the same mutated sample object. Every historical refactor of this file (going back to 2016) deliberately kept `handleAmountChanged` in the `loading` branch and `handleSampleChanged` in the `else` branch, never both. If the second call is required to fix the blur-reset in issue #3157 (e.g., because `handleAmountChanged`'s `setState` doesn't trigger a re-render of `NumeralInputWithUnitsCompo` when the sample reference is unchanged), a comment explaining why both must be called would help future maintainers avoid reverting it.

https://github.com/ComPlat/chemotion_ELN/blob/1c09b3227665bcf3e493316594a104619f0924dc/app/javascript/src/apps/mydb/elements/details/samples/propertiesTab/PolymerSection.js#L24-L32

For reference, `SampleDetails.handleAmountChanged` (the function passed as the prop):

https://github.com/ComPlat/chemotion_ELN/blob/e2848bef7636249ed6565e952d9c85260683767a/app/javascript/src/apps/mydb/elements/details/samples/SampleDetails.js#L285-L289

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

haditariq (approved):
Looks good. Please confirm if loading values has to be reflected in a small heading under element composition table.

Thanks.

TasnimMehzabin:
superseded by #3350

PR #3351, opened 2026-07-01
fix(sample): show mixture total-volume field as disabled instead of hiding it

TasnimMehzabin:


PiTrem (approved):
Safe, well-scoped simplification: the dropped gas_type/contains_residues checks were dead for the Mixture-only render path, and switching hide→render-disabled matches the sibling mixture fields. CI green.

PR #3360, opened 2026-07-07
fix: creating elements in readonly shared collection

### LINKED ISSUE #3335: Users with "read" permission are able to add new samples to synchronised collection.
- I am an ELNer

# Expected Behavior

I shared a collection of samples with multiple users, alway setting the "Permission level" to "Read". The users should not be able to change or add new samples or reaction to that synchronized collection.

# Current Behavior

Users were not able to change any of the samples I added. They were however able to add now samples and reaction despite having their permission set to read only.

headri:
# Summary
  Element creation endpoints did not enforce permission_level when the target collection was shared. Two distinct bugs existed:
  - Security gap (accessible_for without permission check): POST /samples, POST /wellplates, and POST /device_descriptions used Collection.accessible_for which includes shared collections regardless of permission_level. A user with read-only access (permission_level = 0) could create elements in a collection shared with them.
  - Functional gap (current_user.collections only): POST /reactions, POST /screens, POST /research_plans, POST /vessels, POST /cell_lines, and POST /sequence_based_macromolecule_samples queried only the user's own collections. Creating in a shared collection silently ignored the target and either created the element in the user's "All" collection (burning a short_label counter slot) or returned a misleading 404/401.

  An additional frontend bug caused the "Create" button to remain clickable even in read-only shared collections: CreateElementDropdownToggle used a custom forwardRef component that discarded the disabled prop instead of forwarding it to the underlying <Button>.

#  Changes
  app/models/collection.rb — new Collection.writable_by(user) scope returning own collections and shared collections with permission_level >= 1 (write_elements), mirroring the existing accessible_for pattern.

  app/api/helpers/collection_helpers.rb — new writable_collection_for(collection_id) Grape helper wrapping the scope for use in API endpoints.

  API endpoints (sample, reaction, screen, research_plan, vessel ×2, cell_line) — replaced current_user.collections / accessible_for lookups with writable_collection_for. Added helpers CollectionHelpers to ReactionAPI, VesselAPI, and CellLineAPI which were missing it. Endpoints that previously proceeded silently when the collection was not found now return 403 Forbidden before any model is saved, preventing counter advancement.

  Usecases (wellplates/create, device_descriptions/create, cell_lines/create, sbmm/sample) — replaced accessible_for / current_user.collections lookups with Collection.writable_by. Also fixed a pre-existing bug where elements created in a shared collection were linked to the creator's "All" collection instead of the collection owner's "All" collection, making the element invisible to the owner. This is now consistent with the pattern already used by Usecases::Wellplates::Create.

  app/javascript/src/components/navigation/CreateElementButton.js — CreateElementDropdownToggle now forwards the disabled prop to the underlying <Button>. The disable logic (permission_level < Write) was already correct in onUIStoreChange; the prop was simply never reaching the DOM element.

  # Test plan
  - [ ] Shared collection with permission_level = 0 (Read): "Create" button is disabled; direct API calls return 403; short_label counter is not incremented
  - [ ] Shared collection with permission_level = 1 (Write): "Create" button is enabled; element is created in the shared collection and in the collection owner's "All" collection
  - [ ] Own collection: behaviour unchanged
  - [ ] All element types covered: sample, reaction, wellplate, screen, research plan, vessel (single + bulk), cell line, device description, sequence-based macromolecule sample

### DIFF
--- a/.devcontainer/devcontainer-lock.json
+++ b/.devcontainer/devcontainer-lock.json
@@ -1,0 +1,9 @@
+{
+  "features": {
+    "ghcr.io/devcontainers/features/desktop-lite:1": {
+      "version": "1.2.10",
+      "resolved": "ghcr.io/devcontainers/features/desktop-lite@sha256:3cfd1ed97b027614f3a56d53183d20ebf29e9c36ac58a3d669edf39890e3aec3",
+      "integrity": "sha256:3cfd1ed97b027614f3a56d53183d20ebf29e9c36ac58a3d669edf39890e3aec3"
+    }
+  }
+}
--- a/.devcontainer/devcontainer-lock.json
+++ b/.devcontainer/devcontainer-lock.json
@@ -1,9 +1,0 @@
-{
-  "features": {
-    "ghcr.io/devcontainers/features/desktop-lite:1": {
-      "version": "1.2.10",
-      "resolved": "ghcr.io/devcontainers/features/desktop-lite@sha256:3cfd1ed97b027614f3a56d53183d20ebf29e9c36ac58a3d669edf39890e3aec3",
-      "integrity": "sha256:3cfd1ed97b027614f3a56d53183d20ebf29e9c36ac58a3d669edf39890e3aec3"
-    }
-  }
-}
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,6 +6,13 @@ GIT
     inchi-gem (1.07.5)
 
 GIT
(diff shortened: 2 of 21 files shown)


PiTrem (approved):
Re-reviewed at d85bc1b3: all prior blocking findings (orphaned rows on screen/research_plan, unrelated schema/Gemfile.lock drift) and suggestions (reaction_api missing 403, inconsistent HTTP status codes) are fixed, with new RSpec coverage for the read-only-403 / writable-201 paths across the affected endpoints. CI green.

PR #3361, opened 2026-07-07
fix(collections): use PUT for bulk own-collection update

headri:
Bug was introduces here: https://github.com/ComPlat/chemotion_ELN/pull/3294
There, a PUT request was inadvertently changed to a POST.


PiTrem (approved):
Correct regression fix: route is PUT; restores the verb flipped to POST in #3294. CI green.

PR #3362, opened 2026-07-07
build(dev): install asdf via /tmp to avoid cross-device tar error

beque:
install asdf with `--one-top-level=${ASDF_DIR}/bin/ ` causes this errors
```
tar: /home/ubuntu/.asdf/bin/asdf: Cannot open: Invalid cross-device link
tar: Exiting with failure status due to previous errors
```

Fix this with adding a tmp folder for installation

### DIFF
--- a/prepare-asdf.sh
+++ b/prepare-asdf.sh
@@ -21,7 +21,10 @@ if command -v asdf ; then
     fi
 else
     echo '>>> Missing asdf. Installing...'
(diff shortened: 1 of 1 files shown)


PiTrem (approved):
Correct fix for the cross-device tar failure on volume-mounted ${ASDF_DIR}; extract to /tmp then cp. CI green.

PR #3364, opened 2026-07-07
fix(tab-layout): guard null collection labels and stabilize All tab layout

phuang26:


cllde8 (approved):

PiTrem (approved):
Two correct defensive fixes (null-label guard + All-collection tab-layout exclusion) plus a clean consolidation onto isAllCollection; CI green, reviewed at dcd5a99df.

PR #3368, opened 2026-07-09
feat(ui): enable conditional loading of UI components and add weighing tasks feature

TasnimMehzabin:
Closes https://github.com/ComPlat/chemotion/issues/534


PiTrem (approved):
Fail-closed feature-flag mechanism is sound end-to-end (backend gate, frontend hook, camelCase round-trip all verified); no blockers. CI green.

Issue #3297, opened 2026-06-02
Option to also synchronize subcollections of a collection when the collection is shared


medemander2:
- I am an ELNer

# Description
It would be nice if there's an option to also synchronize Subcollections of a collection when it is shared with other users.
E.g. in a practical course, every group of students is allocated a synchronized collection with subcollections for each experiment. Ideally, the supervisor would only have to share the group collection once and the folder like structure is retained for the students in the "Synchronized with me" section.

PiTrem:
Resolved by #3379, which adds the `apply_to_subcollections` option on collection-share creation — the sync-subcollections-when-sharing feature requested here — plus a shared-with-me tree fix so cascaded sub-collections display nested instead of flattened. Extended by #3436, which lets re-editing an existing share (`include_new_subcollections`) also pick up sub-collections created after the original share went out.

PiTrem:
NB: in  v4.0.0. 
- the collection-share setting can be cascaded to all sub-collections. It is a one time action.
- new sub-collection created after this do not inherit automatically the collection-share setting from the parent collection.

Issue #2772, opened 2025-11-10
Subfolders in Synchronized Folders are not synchronized automatically


JTS-ioc:
- I am an ELNer

# Expected Behavior
If a student synchronizes a folder eg "Bachelor thesis" with you, he cannot use subfolders in his ELN, otherwise they are not automatically synchronized with you or need to be individually selected and synchronized. Depending on the structure of the student, the amount of reactions/subfolders can be larger. Ideally the subfolders would automatically be synchronized.

# Current Behavior
Only the head folder is being synchronized, in case the student wants to use their own ordering system with folders, these will not be visible for the supervisor except if manually selected for each.
Automatical synchronization of the subfolders would be desired, when the head-folder is chosen to be synchronized with someone.

PiTrem:
#3379 and #3436 (which resolved the related #3297) substantially improve this: sharing a collection can now cascade to its sub-collections on create, and re-editing an existing share can pick up sub-collections created since. That said, it's still opt-in and requires a manual re-trigger rather than fully automatic — a new subfolder isn't picked up with zero action, which is the specific gap this issue describes. Leaving this open, scoped to that remaining automatic-inheritance behavior.

PR #3379, opened 2026-07-13
feat(collections): shared-with-me tree UX and sharee/delegate capabilities

### LINKED ISSUE #1305: Hierarchical structure for synchronised collections
- I am an ELN admin

# Description
Suggested enhancement:
I suggest that collections and sub-collections can be synchronised with their hierarchical structure. Or that the users are able to structure collections that are synchronised with them, themself. See attached video. On the left side, the original collections is shown (with the hierarchical structure) on the right, how it looks like for the user with whom the collection was synchronised.

Why it is needed:
It is just very useful to be able to synchronise a whole collection structure. This is specially true, at least for us, when larger groups work together in one project. 

# Image

https://user-images.githubusercontent.com/107389350/236212982-2f552e05-dcde-4e96-a35f-9c38074bf557.mp4

### LINKED ISSUE #3275: Shared items
It would be great if the title of the shared collection as given in the elements lists could provide information on the rights that the recipient has. As it's a bit difficult to assess one's options to edit etc if the rights are only visible in the collection management. Enable that information as hover over would be great

PiTrem:
## Summary

Follow-on to the collection-share permission-ladder work (#3374, now on `main`). This
batch builds out the **shared-with-me** experience and the capabilities a sharee /
`manage_shares` delegate has, plus a data migration that backfills owner "All"
memberships. No changes to the permission model itself — everything here rides on the
ladder and guards introduced in #3374.

## What's in it

1. **Dual-owned element badge** — when an element sits in both a collection you own and
   one shared to you (owned by someone else), the collection-label badge highlights the
   shared count and tooltips it, so you can tell the element also lives in a not-owned
   collection.

2. **Name & parent a shared collection created from a selection** — the "share as new
   collection" flow no longer hardcodes the label and top-level placement; the share modal
   (`shareType: 'new'`) gains optional new-collection name and parent inputs, threaded
   through to `addElementsToCollectionAndShare` (falling back to the previous
   label/top-level when left empty).

3. **Export collections shared with you** — export was owner-only; the backend gate now
   also accepts collections you hold any share on (`read_elements`+), and the export modal
   lists shared-with-me collections.

4. **Nest & truncate shared sub-collections in the shared-with-me tree** — the tree used to
   flatten every shared collection under its owner root. It now builds the real hierarchy:
   a shared sub-collection nests under its parent when that parent is also shared, and is
   promoted to the top level under its owner when the parent is not (truncating the branch).

5. **Backfill missing owner "All"-collection memberships** — data migration inserting the
   missing `(element, owner's-All)` join rows for any element present in an owned collection
   but absent from that owner's locked "All". Set-based per join table (8 tables; vessels
   excluded — uuid-keyed, no "All" in the create paths), `NOT EXISTS` + `ON CONFLICT DO
   UPDATE SET deleted_at = NULL` so it is idempotent and resurrects soft-deleted rows.
   Irreversible `down` (documented).

6. **Cascade a share to sub-collections** — `POST /collection_shares` gains
   `apply_to_subcollections`: when set, the same share (level + detail levels, per user) is
   upserted onto every descendant collection the caller may administer. The share modal
   exposes an "Apply these share settings to all sub-collections" checkbox on the create
   flow. The per-descendant guards (privilege escalation, share-with-owner, ownership-offer)
   still apply to each row.

7. **Share management on the shared-with-me tree for `manage_shares` delegates** — the
   backend already authorizes a `manage_shares` (4) delegate to administer a collection's
   shares, but the controls lived only in MyCollections. Adds "Add share" / "Manage shares"
   buttons (and the modal wiring) to shared-with-me nodes where the user's
   `permission_level >= manage_shares`, next to the existing reject-share control.

## Testing

- New/extended specs: `collection_api_spec` (sharee export gate), `collection_share_api_spec`
  (cascade-to-subcollections), and the backfill-migration spec.
- Targeted suites green in the app container (`collection_share_api_spec`,
  `collection_api_spec`, backfill migration — 44 examples, 0 failures); rebased cleanly onto
  the merged #3374.

### DIFF
--- a/app/api/chemotion/collection_api.rb
+++ b/app/api/chemotion/collection_api.rb
@@ -121,13 +121,10 @@ module Chemotion
         # Own collections may always be exported. A sharee may export a collection shared to them
         # only when they hold FULL detail access on it (every element detail level at max): the
         # export path serializes full element content and ignores detail levels, so admitting a
-        # partial-detail share would leak data the share deliberately withheld. Both queries are
+        # partial-detail share would leak data the share deliberately withheld. Both lookups are
         # scoped to the requested ids — this only needs to know the requested set is authorised.
         owned_ids = Collection.own_collections_for(current_user).where(id: collection_ids).pluck(:id)
-        shared = Collection.shared_with_minimum_permission_level(
-          current_user, CollectionShare.permission_level(:read_elements)
-        ).where(id: collection_ids - owned_ids)
-        shared_ids = shared.select { |collection| collection.full_detail_access_for?(current_user) }.map(&:id)
+        shared_ids = Collection.full_detail_access_ids(current_user, collection_ids - owned_ids)
         allowed_ids = owned_ids + shared_ids
 
         error!('401 Unauthorized', 401) if (collection_ids - allowed_ids).any?
--- a/app/api/chemotion/collection_api.rb
+++ b/app/api/chemotion/collection_api.rb
@@ -116,10 +116,15 @@ module Chemotion
       end
       post '/export' do
         collection_ids = params[:collection_ids].uniq
-        all_collection_ids = Collection.own_collections_for(current_user).pluck(:id)
+        # A sharee may export a collection shared to them at any level (read_elements+), in addition
+        # to their own collections.
+        allowed_ids = Collection.own_collections_for(current_user).pluck(:id) +
+                      Collection.shared_with_minimum_permission_level(
+                        current_user, CollectionShare.permission_level(:read_elements)
+                      ).pluck(:id)
 
         error!('Select the collections you want to export.', 403) if collection_ids.empty?
-        error!('401 Unauthorized', 401) if all_collection_ids & collection_ids != collection_ids
+        error!('401 Unauthorized', 401) if (collection_ids - allowed_ids).any?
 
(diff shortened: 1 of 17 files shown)


PiTrem:
Thanks @copilot — addressed in 43fb2b6f:

- **`collection_api.rb` export gate (N+1):** replaced the per-collection `full_detail_access_for?` predicate with `Collection.full_detail_access_ids`, a single grouped `MAX(col) >= 10` query. Since `read_elements` is level 0 the previous permission floor was always satisfied, so this preserves the "any share + full detail" semantics while dropping the N+1.
- **`SelectionShareModal.js` typos:** "Transfering" → "Transferring", "sub collections" → "sub-collections".
- **`SelectionShareModal.js` ladder hint:** reworded to "Each level also grants all the levels below it" for clarity.

The export-modal owner-grouping point (`ModalExportCollection.js`) is a fair UX observation; deferring it as a follow-up since the shared-with-me owner roots are synthetic nodes and rendering them as headings needs more than the one-line change.

PiTrem:
Noting for traceability: item 6 here (`apply_to_subcollections` on share creation) resolves #3297, and item 4 (nested shared-with-me tree) plus item 6 together substantially improve #2772.

PR #3399, opened 2026-07-17
fix(collections): enforce write permission for Split, Copy and My-labels actions

TasnimMehzabin:


PiTrem (approved):
Sound authorization fix: closes real gaps where split/copy endpoints had no or source-only auth; target-collection gating is add_elements- and group-aware, well-tested, CI green.

PR #3403, opened 2026-07-21
fix(spectra): route same-type multi-curve archives through bagit grouping

headri:
spc_type carries the archive's specific spectrum type (e.g. "CYCLIC VOLTAMMETRY"), only falling back to "bagit" for mixed-type archives, so checking `spc_type == 'bagit'` alone missed same-type multi-curve archives (CV, multi-run NMR). Those need read_bagit_data's "N_bagit" filenames, which ContainerDatasetModalContent#classifyAttachments groups into one series; otherwise they landed as ungrouped standalone attachments.


PiTrem (approved):
Reviewed via /code-review (see pr-3403-review.md): sound, narrowly-scoped fix, CI green. Two non-blocking suggestions noted for follow-up (undiscussed single-curve lcms routing change, missing create_process test coverage).

PR #3404, opened 2026-07-21
fix(collections): notify users about locked samples during removal actions

### LINKED ISSUE #462: Sharing reaction and then trying to unshare a sample. - Please make a Popup: "You cannot unshare a sample connected to a shared reaction"
- [ ] I am an ELNer
- [ ] I am a developer
- [x] I am a system/ELN admin

# Expected Behavior

After sharing a reaction, if someone tries to remove a sample then a pop will a great option to show whether it fails to remove the sample or if there is any other issue.

# Current Behavior

After sharing samples and reactions with the other user, if I try to unshare the single sample then it's not getting removed from the shared collection. It is getting removed only when I removed the reaction or all the samples. But when I try to remove the sample from the shared collection, nothing happened on the screen and it is hard to say if it fails or there is some other issue.

## Steps to Reproduce

1. Share the reaction containing samples with the other user.
2. Try to unshare the one sample by removing it from the shared collection.
3. Nothing happens on the screen. It leaves the user in a confused state if something happened or not.

## Context

 - [x] Chrome Browser 
 - [ ] Firefox Browser
 - ELN version, or branch, or commit

## Failure Logs / Screenshots

Include any relevant log snippets or files here.
image
image (1)

### LINKED ISSUE #1033: Remove from collection issue
Current behavior: 
- A collection has 1 reaction, 2 samples that are part of a reaction
- User wants to remove sample from the collection: nothing happens (unexpected)
- User needs to remove reaction from the collection: samples are removed with the reaction (expected)

Desired behavior: 
- User wants to remove sample from the collection: user gets notification that the reaction needs to be unassigned
- User removes/unassigns reaction from the collection: samples are removed with the reaction

### LINKED ISSUE #2442: error notification "remove sample from collection"
If one tries to remove a sample from a collection and the sample is part of a reaction, one should receive an error notification/info why the sample cannot be removed. 
Current behavior: no notification, the sample is just not removed

TasnimMehzabin:
Resolves the following,

- https://github.com/ComPlat/chemotion_ELN/issues/462
- https://github.com/ComPlat/chemotion_ELN/issues/2442
- https://github.com/ComPlat/chemotion_ELN/issues/1033

### DIFF
--- a/app/api/chemotion/collection_elements_api.rb
+++ b/app/api/chemotion/collection_elements_api.rb
@@ -37,8 +37,8 @@ module Chemotion
         )
 
         # Some samples cannot be unshared on their own because they belong to a
-        # reaction that is still in the collection. Report them so the UI can
-        # explain why nothing happened; otherwise reply 204 No Content.
+        # reaction or wellplate that is still in the collection. Report them so the
+        # UI can explain why nothing happened; otherwise reply 204 No Content.
         if result[:locked_sample_ids].present?
           status 200
           { locked_sample_ids: result[:locked_sample_ids] }
--- a/app/api/chemotion/collection_elements_api.rb
+++ b/app/api/chemotion/collection_elements_api.rb
@@ -31,12 +31,21 @@ module Chemotion
 
       desc 'Removes elements from a collection'
       delete '/:collection_id' do
-        Usecases::Collections::RemoveElements.new(current_user).perform!(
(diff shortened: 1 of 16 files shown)


PiTrem (approved):
Blocking finding from prior review resolved (locked-set reconciliation now runs after cascades); CI green; remaining items are non-blocking suggestions/nits per pr-3404-review.md.

PR #3408, opened 2026-07-22
fix(converter): lazy-load converter-client to stop global scrollbar loss

headri:
@complat/chemotion-converter-client injects an unscoped `@media (min-width: 990px){body{overflow:hidden}}` rule as a side effect of import. ConverterAdmin.js imported it statically, and since application.js requires ConverterAdmin unconditionally on every page, this disabled body scrolling app-wide on desktop widths once the JS bundle finished loading. Load it via React.lazy so it's only fetched when the converter admin page actually mounts.

### DIFF
--- a/app/javascript/src/apps/converter/ConverterAdmin.js
+++ b/app/javascript/src/apps/converter/ConverterAdmin.js
@@ -1,9 +1,16 @@
-import React, { useEffect, useState } from 'react';
+import React, {
+  useEffect, useState, Suspense, lazy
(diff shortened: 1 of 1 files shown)


PiTrem:
Superseded by #3412, which bumps the ELN dependency pin to converter-client 0.16.2 (containing ComPlat/chemotion-converter-client#173's root-cause fix for the scroll-lock CSS leak). The lazy-load approach here is no longer needed to resolve #542.

PR #3411, opened 2026-07-22
fix(collections): honor can_update for read-only shared research plans, screens, and wellplates

### LINKED ISSUE #1369: Modifying read only collections
- I am an ELNer

# Expected Behavior

If I create a shared collection with "Permission level" read I expect that the person with who I shared the collection cannot change anything.

# Current Behavior

If the collection contains a Research Plan the person with only read rights can modify the ResearchPlan (which is at least odd). But when pressing "save" it is saving forever. No error is thrown and no message is shown. However, it does not stop saving. I stopped it after ca. 3 min.

## Steps to Reproduce

1. Create a collection
2. Create a research Plan within the collection
3. Share the collection with someone and only assign the Permission level: read.
4. The person with who you shared the collection has to modify now the research plan

## Context

 - Windows/Linux 
 - Firefox Browser

TasnimMehzabin:
## Summary
- Expose `can_update` from research plan, screen, and wellplate detail APIs via `ElementPolicy`, and default models to editable when the flag is omitted from list payloads.
- Gate Save and close-overlay behavior, as well as fields, analyses, attachments, drop zones, designer/list editing, metadata, and related actions when `can_update` is false.
- Add API/entity/model specs covering owner, read-only share, and writable share permission cases.

## Test plan
- [x] Open a research plan / screen / wellplate shared read-only: Save disabled; fields and mutating controls not editable
- [x] Confirm Attachments tabs hide upload/edit/delete; download still works
- [x] Confirm Screen wellplates/research-plan drop zones and “Add” actions are hidden when empty/read-only
- [x] Confirm Wellplate designer and list cells cannot be edited when read-only
- [x] Confirm Research Plan metadata and wellplates tabs are read-only when shared that way
- [x] With edit-share permissions, editing and Save still work
- [x] Attempting Save on a read-only element does not offer “Save and Close”; Discard path works if local dirty state exists
- [x] Run relevant API/entity specs (`screen_api`, `wellplate_api`, `research_plan_api`, entity specs) and JS model specs

Closes https://github.com/ComPlat/chemotion_ELN/issues/1369

### DIFF
--- a/app/api/chemotion/research_plan_api.rb
+++ b/app/api/chemotion/research_plan_api.rb
@@ -100,7 +100,8 @@ module Chemotion
         research_plan.collections << all_coll if all_coll && research_plan.collections.exclude?(all_coll)
 
         update_element_labels(research_plan, params[:user_labels], current_user.id)
-        present research_plan.reload, with: Entities::ResearchPlanEntity, root: :research_plan
+        @element_policy = ElementPolicy.new(current_user, research_plan)
+        present research_plan.reload, with: Entities::ResearchPlanEntity, root: :research_plan, policy: @element_policy
       end
 
       namespace :table_schemas do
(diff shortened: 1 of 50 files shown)


PiTrem (approved):
Reviewed via /code-review (7/8 finder angles) across two rounds plus a follow-up perf fix: all four Blocking findings from the first round are fixed and verified (server-side auth on research_plan_metadata_api, fail-open/fail-closed can_update default, wrong-entity permission gating, element_policy scope-creep into SampleEntity); the fail-open list-endpoint N+1 this PR introduced (per-record ElementPolicy with no eager-load) was root-caused to an unnecessary unconditional can_update exposure and fixed by guarding it the same way reaction/sample already are — confirmed via benchmark that the +110% shared-collection list latency regression is fully gone. CI green, 184 examples covering owner/read-only-share/writable-share cases across API/entity/model layers.

PR #3419, opened 2026-07-28
fix(export): render research-plan structure images at correct aspect ratio and size

### LINKED ISSUE #3179: report of research plan to docx - structure
- I am an ELNer

Currently a structure that is used in a research plan is displayed as kind of a picture - different from the way we do it for reactions and samples:

Image

Should be shown in the same way we do it for reports of normal samples.

v3.1.0-patch1

TasnimMehzabin:
## Problem

Chemical structures embedded in a research plan (a linked sample, a
linked reaction, or a freehand Ketcher drawing) rendered in the exported **docx**
as low-quality, oddly-shaped pictures, unlike the crisp images in a plain
sample/reaction report.

## Root cause

`ExportResearchPlan#to_png` rasterized each structure's raw SVG through
`Reporter::Img::Conv.by_inkscape`, whose defaults (`1550x440`) are shaped for a
wide **reaction scheme**. Combined with `--export-area-page`, a square sample SVG
(`200x200`) was stretched ~3.5x into that wide canvas, producing the distortion.
The composer-driven sample/reaction *report* path never hits this because it
composes structures into a canvas that already matches `1550x440`; the
research-plan path feeds raw editor SVGs straight in.

## Changes

**Rasterize at the SVG's own aspect ratio**
- New shared helpers in `Reporter::Img::Conv`: `svg_page_dimensions` (reads the
  root `<svg>` width/height, viewBox fallback, ignoring nested layout viewBoxes)
  and `export_size_for` (fits the page into a max box preserving aspect ratio).
- `to_png` now sizes the export canvas per-SVG, so samples stay square, reactions
  stay wide, and Ketcher drawings keep their proportions.

**Control on-page size (Pandoc embeds a bare `width` as px @ 96 dpi)**
- Roughly-square structures (samples, Ketcher) use `STRUCTURE_DISPLAY_WIDTH = 300`
  (~3.1 in) so a high-res PNG no longer dominates the page.
- Wide reaction schemes use `REACTION_DISPLAY_WIDTH = 600`, filling the page text
  column at a legible height (~1.6 in) instead of ~0.9 in at 300 px.
- The `width` attribute is emitted only when the image `src` is present, so a
  missing structure no longer produces a width-only broken ``.

**Consolidation & robustness**
- `ExportExcel` now delegates to the shared `Conv.export_size_for`, replacing its
  private, divergent SVG-dimension helpers (it read viewBox-first via regex and
  could pick up a nested `<svg>`); xlsx and docx now size SVGs identically.
- viewBox parsing handles comma-separated and leading-whitespace values.
- `parse_svg_length` handles scientific notation (`1e3` becomes `1000`) and
  preserves a leading minus sign, so degenerate/negative lengths fall through to
  the viewBox instead of being silently flipped positive.
- Retained PNG `Tempfile`s are explicitly closed/unlinked after Pandoc reads them,
  fixing a premature-GC risk and a temp-file leak.

## Testing

- Visually verified the Inkscape output: a squashed benzene ring became a crisp,
  correctly-proportioned one.
- Measured the embedded docx image sizes via Pandoc across width candidates to
  pick the reaction width (Word clamps anything wider than the text column).
- Added specs in `spec/lib/reporter/img/conv_spec.rb` and
  `spec/lib/export/export_research_plans_spec.rb` covering aspect-ratio sizing,
  viewBox edge cases, `parse_svg_length` (units, scientific notation, sign,
  percentages), the sample/reaction/Ketcher width hints, the missing-image case,
  and the fallback path.

## Notes

- Deliberately keeps the research-plan pipeline on HTML to Pandoc rather than
  switching to Sablon OLE embedding (true report-parity); that is a larger,
  separate change if pixel-perfect parity is ever required.
- `ExportExcel` now derives a reaction scheme's aspect ratio from the outer
  `<svg>` page rather than a nested viewBox, worth a glance during QA of xlsx
  export, though square molecule images (the common case) are unaffected.


PiTrem (approved):
Re-reviewed at 80eec3dd: the premature-tempfile-cleanup blocking issue from the earlier review is fixed and covered by a new regression spec. CI green, remaining items are non-blocking follow-ups.

PR #3421, opened 2026-07-28
fix(spectra): resolve 2D source data via NMRium's sources[]/selector.root registry

headri:
## Summary

  Fixes a crash reopening a saved 2D NMR spectrum (HSQC/HMBC/COSY/NOESY etc.) in NMRium after saving it from the NMR
  viewer. Root cause: `SpectraHelper.js`'s `cleaningNMRiumData` stripped a 2D spectrum's `info`/`data` before saving
  whenever it referenced a source file (jcamp/zip), on the assumption NMRium would re-derive them by re-parsing the
  source on reload. It doesn't — traced into NMRium's actual `@zakodium/nmrium-core` source and confirmed its reader
  only re-fetches a spectrum when `spectrum.selector.root` matches an id in a top-level `sources[]` registry; it never
  uses this repo's ad hoc `source`/`sourceSelector` fields for that. Every previously-saved 2D spectrum therefore fell
  into NMRium's inline render path with neither embedded data nor a resolved fetch, crashing in various ways depending
  on what was missing.

  ## What changed

  - `cleaningNMRiumData` now builds a real `sources[]` + `spectrum.selector.root` reference (via new
  `buildSourceId`/`ensureSource` helpers, keyed off the spectrum's own stable file label) for 2D spectra backed by a
  source, and only then drops the embedded `data` matrix — closing the crash at its root instead of trading it for one
  of the earlier `f16f79b06`/`1f15fe887`/`90bef9bf5` workarounds that just always kept everything embedded (safe, but
  risks pushing large 2D matrices — 2048×1024+ points — over Shrine's 100MB upload cap on save).
  - Deliberately does **not** wrap the result in `{version, data: {...}}` or declare an explicit schema version:
  verified against a real, working NMRium-captured `.nmrium` file that a source-backed spectrum reloads fine with
  neither, and declaring a version explicitly seems to skip whatever normalization an unversioned document gets on load.
  - Widened `prepareNMRiumDataAttachment`'s `hasAnySource` check so spectra using the new `selector.root`/`sources[]`
  shape are still recognized (avoids an incorrect legacy `{version: 7, ...}` wrapper firing for them).
  - 1D spectra and the jdx-only first-load fallback (no `.nmrium` attachment yet) are untouched — no bug was ever
  confirmed there, and an earlier attempt to move the fallback onto the same mechanism broke first-time uploads
  outright.
  - `isSourceOnly2D` renamed to `is2DWithSource` and the repeated `info?.dimension === 2 || originalInfo?.dimension ===
  2 || meta?.dimension === 2` check extracted into a shared, exported `isSpectrum2D` (was duplicated 3x across two
  files).

  ## Test plan

  - [x] `yarn test` — full JS suite passes (1457 tests), including new cases in `SpectraHelper.spec.js` covering:
  building `sources[]`/`selector.root` and dropping `data` for 2D-with-source spectra; falling back to keeping `data`
  when no source URL resolves; collapsing same-named spectra sharing one zip into one `sources[]` entry with
  per-spectrum disambiguation; 1D spectra unaffected.
  - [x] `yarn eslint` on both touched files — no new errors vs. `main`.
  - [x] Manually verified end-to-end against the live NMRium instance: save a 2D spectrum with a jcamp source → reopen →
  renders correctly (previously crashed). Save/reopen a 1D spectrum → unaffected. First-time `.jdx` upload (no existing
  `.nmrium`) → unaffected.

  ## Known limitation

  Large 2D datasets *without* a resolvable source URL still embed the full matrix (unchanged, pre-existing behavior) and
  can still approach the 100MB Shrine cap in that case — accepted tradeoff, discussed and confirmed with the reporter.

---

## Review follow-ups (`6b88fde4`)

A review pass found that the `sources[]` reference, once written, did not always resolve back to
the file it came from — so `data` was dropped in exchange for a reference that could not be
followed. Fixed here:

- **Stale token on reopen.** `resolveSpectrumSourceUrl` preferred the `sources[]` entry persisted
  by an earlier save over the freshly re-minted `root.source`. Attachment URLs carry a 48h token,
  so that entry is expected to be dead by the next reopen — the spectrum was re-pointed at an
  expired URL with no embedded data to fall back on. This reproduced the exact failure the PR
  fixes, just delayed by one save cycle. The fresh URL now wins.
- **The registry addressed a member inside the archive, not the archive.** A new `splitArchiveRef`
  separates the zip URL from the member path across all three shapes that reach it (absolute URL,
  `patchZipAndJcampReference`'s rewritten server path, bare member), so `sources[]` holds the
  archive and `selector.files` holds paths that can match its contents. Server paths previously
  passed the "not absolute" filter and landed in `selector.files` verbatim, matching nothing.
- **Two different files could collapse onto one source id.** `buildSourceId` keys off the display
  name, and `patchZipAndJcampReference` forces one label onto every spectrum when a zip is present
  — so identical names are the normal case, and the second spectrum's `ensureSource` overwrote the
  first's URL while both had `data` already deleted. `ensureSource` now takes a per-pass URL → id
  map: same file shares one entry, different files get distinct ids.
- **Multi-entry sources** are no longer reduced to a single element.
- **Rename before cleaning** in the zip fallback, so `selector.root` is derived from the same name
  the next save will derive its id from (previously `patchZipName` ran after `cleaningNMRiumData`
  and every rename left a dead `sources[]` entry).
- **`info` backfill** lifts only `dimension`/`isFid` out of `meta` rather than spreading the whole
  raw JCAMP header dictionary into `info`.
- **Robustness**: the wrapper copy is guarded (`data` may legitimately be a non-object) and
  `new URL()` is wrapped, so a prefix-matching but malformed URL can no longer take a save down.
- Orphaned `nmrium-src-*` entries are pruned; `buildSourceId`/`ensureSource` dropped from the
  export list (imported nowhere); two verbose forms simplified.

Adds regression specs for the reopen, member-path and collision cases.

### Correction to the summary above

The "the jdx-only first-load fallback … untouched" bullet holds for the **jdx** path, but the
**zip** first-load path does go through `cleaningNMRiumData` and was affected — that is what the
`splitArchiveRef` and rename-ordering items above address.

### Still open

1D source-backed spectra keep their `data` on this branch, where `main` dropped it unconditionally
for any source-backed spectrum. No 1D reload bug was ever reported against `main`'s behaviour, so
the extra payload may be buying nothing — and it works against the Shrine 100 MB rationale cited
above. Left as-is because a spec here asserts the new behaviour; worth settling deliberately
before merge.

### Verification

- `yarn test` (full suite, in the app container) — **1464 passing, 0 failing, 19 pending** at
  `6b88fde4`.
- Rebases onto current `main` cleanly; the full suite there is **1712 passing, 0 failing**.
- `yarn eslint` **not** re-run: it fails to load the `storybook` plugin in this container on every
  branch, including untouched `main`. Left to CI's `linting` job.

### DIFF
--- a/app/javascript/src/components/nmriumWrapper/NMRiumDisplayer.js
+++ b/app/javascript/src/components/nmriumWrapper/NMRiumDisplayer.js
@@ -252,9 +252,13 @@ export default class NMRiumDisplayer extends React.Component {
         return;
       }
 
+      // Rename before cleaning, not after: cleaningNMRiumData derives each spectrum's sources[] id
+      // from its name, so renaming afterwards would leave selector.root pointing at an id the next
+      // save no longer mints.
+      if (zip?.label) this.patchZipName(nmriumState, zip?.label);
+
       const cleaned = cleaningNMRiumData(nmriumState);
 
-      if (zip?.label) this.patchZipName(cleaned, zip?.label);
       if (molfile) { cleaned.molecules = [{ molfile }]; }
 
       this.iframeRef.current?.contentWindow.postMessage({ type: 'nmr-wrapper:load', data: { type: 'nmrium', data: cleaned } }, '*');
--- a/app/javascript/src/components/nmriumWrapper/NMRiumDisplayer.js
+++ b/app/javascript/src/components/nmriumWrapper/NMRiumDisplayer.js
@@ -8,7 +8,7 @@ import UIFetcher from 'src/fetchers/UIFetcher';
 import Attachment from 'src/models/Attachment';
 import { SpectraOps } from 'src/utilities/quillToolbarSymbol';
 import { FN } from '@complat/react-spectra-editor';
-import { cleaningNMRiumData } from 'src/utilities/SpectraHelper';
+import { cleaningNMRiumData, isSpectrum2D } from 'src/utilities/SpectraHelper';
 
 export default class NMRiumDisplayer extends React.Component {
   constructor(props) {
@@ -151,11 +151,7 @@ export default class NMRiumDisplayer extends React.Component {
       if (!rawState) return;
 
       const spectra = rawState?.spectra || rawState?.data?.spectra || [];
-      const is2D = this.state.is2D || spectra.some((spc) => (
-        spc?.info?.dimension === 2
-        || spc?.originalInfo?.dimension === 2
-        || spc?.meta?.dimension === 2
-      ));
+      const is2D = this.state.is2D || spectra.some(isSpectrum2D);
       const version = rawState?.version ?? rawState?.data?.version ?? 1;
       const nmriumData = version > 3 && rawState.data ? rawState.data : rawState;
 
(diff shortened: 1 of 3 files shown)


PiTrem (approved):
Registry approach verified against a real next-release .nmrium; both blocking findings fixed and covered by regression specs.

PR #3427, opened 2026-07-31
perf(samples): eliminate N+1 queries and duplicate counts in sample list endpoint

headri:


PiTrem (approved):
Performance fix is sound: eliminates the sample-list N+1 (comment_count, reactions_samples, molecule computed_props) and a duplicate pagination-total count query, backed by a regression test and green CI. One outstanding nit (a self-contradicted code comment in sample_entity.rb) is non-blocking.

PR #3429, opened 2026-08-03
fix(SelectionShareModal): prevent invalid ownership transfer to groups or multiple users

TasnimMehzabin:


PiTrem (approved):
Small, correctly-scoped frontend guard mirroring the existing backend invariant in prevent_invalid_ownership_offer!. Verified against all call sites (create vs. edit flows) and the type-field data path. No blocking issues; full review recorded in cgit as pr-3429-review.md.

PR #3434, opened 2026-08-04
fix(research-plan): resolve linked sample and reaction ids on export/import

headri:
Research plans with linked samples or reactions now export and import correctly. On import, a linked sample/reaction now points at the newly created element instead of a stale id from the source system.

When the linked element isn't part of the export (e.g. it lives in a collection that wasn't selected), the link is preserved as static chemistry instead of silently breaking, provided the exporting user has permission to see that level of detail: a sample becomes a Ketcher structure field, a reaction becomes an image of its report-style scheme. If the exporting user can't see it, the link is dropped instead — and the drop is now recorded in the import log so it's visible, not silent.

Also fixed: an unfilled sample/reaction placeholder field (added but never linked to anything) survived the round trip untouched instead of being dropped, and the RADAR export path now carries the acting user through so this fallback works there too.

### DIFF
--- a/app/controllers/oauth/radar_controller.rb
+++ b/app/controllers/oauth/radar_controller.rb
@@ -63,7 +63,7 @@ class Oauth::RadarController < ApplicationController
           collection.metadata.reset_radar_ids
 
           # enqueue the job to create the files and upload them to radar
-          ExportCollectionToRadarJob.perform_later(access_token, collection_id, response['id'])
+          ExportCollectionToRadarJob.perform_later(access_token, collection_id, response['id'], current_user.id)
 
           return redirect_to action: 'export'
         end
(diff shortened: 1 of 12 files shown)


PiTrem (approved):
Re-reviewed via /code-review xhigh at 530fa2e — both blocking findings from the prior pass (orphan attachment leak, uuid?-as-membership-test bug) are fixed and verified against the code, all previously-flagged test-coverage gaps are now closed with real regression tests. CI green, mergeable. A few non-blocking follow-ups noted in pr-3434-review.md (latent json/udm attachment-cleanup gap, missing reaction-image dedup, an incidental error-swallowing change in Reaction#update_svg_file!) — none merge-blocking.

PR #3445, opened 2026-08-06
refactor(api): make detail-level handling consistent across list, write, report and export paths

TasnimMehzabin:
Extends the collection detail-level enforcement so every server-side path — list/search,
writes, report generation and exports — applies the same per-collection detail levels and
access checks the single-element endpoints already use, which they were previously
inconsistent with.

## Changes

**List / search serialization** — pass the collection's per-class detail levels to the
entities in the bulk paths that did not compute them (`search_api.rb`
`serialization_by_elements_and_page`, `by_ids.rb`, `samples/ui_state`, and
`shared_methods.rb` for advanced/structure search). Levels are resolved once per page via
`ElementDetailLevelCalculator.for_collection`; samples keep their per-element calculation.

**Writes** — `ElementPolicy#update?` requires a full detail level to edit a shared element
(alongside edit permission), consistent with `#copy?`. `can_update` makes the UI forms
read-only for shares below full detail.

**Reports** — require element read access before generating a report (`GET /reports/docx`,
`POST /reports`); the report worker skips objects the author cannot read. The report numeric
formatter no longer errors on placeholder values, so reports for below-full-detail shares
render normally.

**Exports** — correct a collection-share scope join in the reaction-sample export, add the
detail-level row guard the analyses sheet already has to the components sheet, and apply
`is_top_secret` masking on the XLSX/SDF export.

## Verification

- RuboCop clean on changed lines; the calculator + entity specs, the six element PUT APIs,
  the policy specs, and the report/export specs pass.
- `ElementDetailLevelCalculator` produces correct per-class levels and the entities redact
  `anonymize_below` fields at a below-threshold level.

## Follow-ups (not in this PR)

- Consider making `ApplicationEntity#detail_levels` fail-closed.
- `DeviceDescriptionEntity` / `CellLineSampleEntity` expose fields without anonymization, so
  their detail-level threading is currently inert.
- Minor query-reuse efficiency in the list paths.

### DIFF
--- a/app/api/chemotion/report_api.rb
+++ b/app/api/chemotion/report_api.rb
@@ -19,6 +19,23 @@ module Chemotion
       def time_now
         Time.zone.now.strftime('%Y-%m-%dT%H-%M-%S')
       end
+
+      # Rejects the request unless the caller may read every element referenced in +objTags+.
+      # Without this a report is an IDOR: the generator does a bare +Model.find(id)+ with no
+      # ownership/collection check, so any authenticated user could pull any element's data.
+      # Only Sample/Reaction are accepted — the types {Reporter::Worker#extract} renders; any
+      # other type is rejected rather than silently ignored (the worker would raise on it anyway).
+      #
+      # @param obj_tags [Array<Hash>] the +objTags+ param ({ 'id' =>, 'type' => } entries)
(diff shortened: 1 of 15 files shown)


PiTrem (approved):
Detail-level handling made consistent across list, write, report and export paths; CI green (pronto/rubocop warnings are pre-existing/waived).

Issue #599, opened 2021-12-06
Generic elements: formula field: calculate units too


NRayya:
- [ ] I am an ELNer
- [ ] I am a developer
- [ ] I am a system/ELN admin

# Description
Currently in the generic elements in the formula fields, only the number gets calculated, but the corresponding unit is not shown along with it. I think it would be better to show the unit as well.

Issue #597, opened 2021-12-04
Generic elements: Enhance Checkboxes by adding new fields


NRayya:
- [ ] I am an ELNer
- [ ] I am a developer
- [ ] I am a system/ELN admin

# Description
Currently, check-boxes can only indicate if a certain information is applicable or not. It would be great to be able to add optional fields that appear only when the checkbox is checked. For example, one could have a checkbox "Frozen" to indicate if an enzyme is frozen or not, when checked, a new text field appears to input the freezing method.

NRayya:
Closing issue: Already possible with restrictions for the layers

Issue #533, opened 2021-09-08
Generic elements/segments: select function in tables


nicolejung:
The table includes already a lot of functionality. It would be great to have the select function included as well.

nicolejung:
was done and is added, I will close the issue

PR #3450, opened 2026-08-10
fix(ui): sidebar, notifications, and detail-header UX papercuts

headri:
Started as a fix for unbounded notification pagination and grew to cover a batch of related minor UX/UI issues found along the way, all in the same low-risk "layout/CSS + small behavior fix" category:

- **Notification pagination** — the notification list's pagination is now windowed to a fixed number of visible pages (with First/Prev/…/Next/Last), instead of rendering one page button per page.
- **Reaction → Analyses tab misclick** — a product's sub-tab header (icon + title) is now only clickable to open the Sample view when that tab is already active; clicking a non-active tab just switches tabs instead of navigating away.
- **Long collection names overflow the sidebar** — labels with a very long word no longer overflow the sidebar into the element list panel; they wrap (up to 2 lines) and stay readable instead of being clipped or blowing out the layout.
- **Sidebar collapse/expand button clipping** — fixed a regression introduced while fixing the above: the sidebar's own collapse/expand toggle was being clipped by the new overflow handling.
- **Element-detail header alert layout** — the "Structure Alert" and "you are viewing the original sample" notices no longer distort the panel header's height or get squeezed between the header buttons; they render as a full-width banner below the header row.
- **Element-list action-button wrapping** — in the narrow width range between icon-only and icon+text display, action buttons (Share, References, My labels, …) no longer wrap their label onto a second line and blow up out of alignment with the rest of the row; a button that doesn't fit now drops to a full second row instead.

---

Ruby testscript to generate test notifications (for testing the pagination fix):
```ruby
user_email = "complat.user1@eln.edu"
count = 40

receiver = User.find_by!(email: user_email)
sender = receiver

channel = Channel.find_by!(subject: Channel::SEND_INDIVIDUAL_USERS)

count.times do |i|
  Message.create_msg_notification(
    channel_id: channel.id,
    message_content: { data: "Test notification ##{i + 1}" },
    message_from: sender.id,
    message_to: [receiver.id],
  )
end

puts "Created #{count} notifications for #{receiver.email}"
```

### DIFF
--- a/app/assets/stylesheets/components/CollectionTree.scss
+++ b/app/assets/stylesheets/components/CollectionTree.scss
@@ -53,9 +53,17 @@
   &__title {
     @extend .flex-grow-1;
     min-width: 0;
-    overflow-x: hidden;
-    white-space: nowrap;
-    text-overflow: ellipsis;
+    // Wrap normally (so a multi-word long label still shows in full across a
+    // couple of lines, as it did before long-label overflow was clipped) but
+    // allow breaking mid-word so a single unbroken word can't force the row
+    // wider than the sidebar. Capped at 2 lines so one extreme label can't
+    // stretch the row indefinitely; anything beyond that gets an ellipsis on
+    // the second line.
+    overflow-wrap: anywhere;
+    display: -webkit-box;
+    -webkit-line-clamp: 2;
+    -webkit-box-orient: vertical;
+    overflow: hidden;
   }
 }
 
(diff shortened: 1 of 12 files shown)


PiTrem (approved):
Reviewed and re-reviewed across all commits (see pr-3450-review.md); CI fully green, all three underlying chemotion issues (#533, #597, #599) verified resolved and cross-linked.

PR #3453, opened 2026-08-10
fix(collections): keep the repository subtree nested, and make collection ownership personal

PiTrem:
> **Note — this PR now also changes what "owns a collection" means.** The last two commits are a
> breaking semantic change, not part of the original bug fix. See *Ownership is personal* below.

## Problem

The `transferred` collection — created by a gate transfer under the user's locked
`chemotion-repository.net` root — is shown at the **top level of "My Collections"** even though its
`ancestry` in the database correctly points at that root. Saving the collection tree then makes the
display true: `ancestry` is rewritten to `/` and the collection has permanently left the repository
subtree.

## Why a correct ancestry does not produce nesting

The frontend never derives the parent link from `ancestry`. `addCollectionToTree` derives it by
searching *the array it is inserting into* for a node claiming to be an ancestor:

```js
const parentIndex = collectionTree.findIndex(element => element.isAncestorOf(collection))
```

So a collection nests only if its parent is already an element of that same array — and the
repository root never is. It is assigned to the scalar `chemotion_repository_collection` field and
never pushed into `own_collections`, nor is it inside its own `children` array.

What hides this is the routing line below it: anything whose `ancestorIds` include the repository id
is sent into `chemotion_repository_collection.children`, and the sidebar renders that array under
the "chemotion-repo" group header. **The nesting a user sees there is the group header, not a
parent–child link.** So the fallback in `addCollectionToTree` fires on every load and flattens the
collection to `'/'` — invisible, because the header supplies the apparent nesting — and if the
destination array is chosen wrongly the collection lands in `own_collections` instead, where there
is no header to fake it and it renders as a root.

The destination is chosen by that same scalar field, which was assigned on the **label alone, for
every row carrying it**, so the last one won. A member of a group whose account still carried a
legacy `chemotion-repository.net` collection therefore lost the field to the group's.

## Changes

**1. Remove the obsolete repository collections of group accounts.** A repository collection belongs
to a person; `User#create_chemotion_public_collection` has been Person-only for years, so these are
legacy rows that cannot recur. Soft-deleted, only rows holding nothing, anything with content
reported and left alone.

**2. Keep the repository subtree out of the own-collections tree.** `addCollectionToTree` still
places a parentless collection at the top of the array it is given but no longer rewrites its
`ancestry` — that field is the persisted parent pointer, not display state. The repository node is
matched on `is_locked` as well as the label (an imported archive can carry an unlocked collection of
that name), the first match is kept, and the field is rebuilt per pass.

**3. Lock `transferred`, and keep it reachable.** Every guard that should keep it in place keys off
`is_locked`, yet it was created unlocked. Now created locked, with the existing rows locked by
migration — only those under a locked repository root, never on the label alone. A root-level
`transferred` is reported, not moved: nothing distinguishes one that escaped from one a user made.
`CollectionSubtree.handleClick` now navigates for a locked node as well as expanding it, since
`transferred` has no sub-collections and would otherwise be unclickable.

**4. Keep locked containers closed to tree moves and to creates.** `UpdateTree` excludes a locked
container's whole subtree, not just its direct children — a grandchild listed at the top level
escapes identically. `Create` refuses a locked parent anywhere on the path, which would otherwise
make such a collection immovable for good.

**5. Ownership is personal, group reach is through shares.** `Collection.own_collections_for` and
`#owned_by?` counted a group's collections as owned by every member. Pre-#2783 the code kept three
notions apart — the tree was strictly `user_id = current_user.id`, group reach was an opt-in
argument that additionally required `is_locked = false`, and only element access ever passed
`group_ids`. #2783 merged them, and every consumer inherited group inclusion whether it wanted it or
not: any member could rename, delete, re-parent, share onward or offer pass-ownership of a group's
collection, and because `UpdateTree` writes `user_id: current_user.id` into every payload entry, a
member merely saving their tree reassigned the group's collections to themselves. Dropping the
`is_locked = false` guard is also what pulled a group's locked rows into each member's payload —
the cause of the bug above.

Group membership is meaningful as a **sharee**, and that channel already exists:
`CollectionShare.shared_with` resolves `shared_with_id IN (self, *group_ids)`. So ownership now keys
on `user_id == user.id`, and `accessible_for` / `writable_by` / `belongs_to_or_shared_by` keep group
ids on their share half only — "mine, or shared to me or to one of my groups".

The same conflation was repeated in four other places, each now split: `ElementsPolicy` used one id
list for both the ownership test and the share recipients (so a member holding only `:read_elements`
passed bulk update and destroy as an owner); `ElementDetailLevelCalculator` returned owner-level 10
where the model returned zeros; `CollectionShare.shared_by` resolved the sharer as the member; and
`WithdrawElements` carried its own group-widened query.

**6. Stop giving group accounts an "All" collection.** It backs the MyDB element paths, and
`config/routes.rb` routes a `Group` session to the command-and-control view for `/mydb` — a group
can never use one. Guarded on `Group` specifically rather than on `Person`: `Admin` subclasses
`User`, and admins do hold and use an "All" collection. Existing group rows retired by migration,
and the paths that attach an "All" now nil-check it, since an owner without one is a reachable
state.

## Testing

- Ruby (`spec/api spec/models spec/usecases spec/policies spec/db/migrate spec/jobs spec/services`)
  — 2197 examples, 4 failures, all pre-existing on `main` (attachment path, two RADAR metadata, one
  SFTP admin-device).
- JS — 1502 passing, 0 failing.
- RuboCop clean on every touched file; ESLint at or below baseline.

Six specs asserted the old ownership rule and now assert the new one, across the model, the policy,
the withdraw use case, the element API and group/user creation.

### Staged against a production copy

Both cleanup migrations and the lock migration were run end to end against a copy of a real
database: every `transferred` row locked and nested under its owner's locked repository root, none
at root level, no group-owned collection left alive, 860 person "All" collections intact, and the
shares addressed to groups still resolving (390 rows, 31 collections, 128 members reached).

## Notes

Two known-latent items left alone deliberately. The `transferred` lock cannot distinguish a row a
user created under the repository root via the API from the job's — documented in the migration,
and `Create` now refuses that parent, so the population is bounded. And a collection nested under
the locked "All" would be rejected wholesale by `UpdateTree`; there are none, and `Create` now
prevents making one.

### DIFF
--- a/app/api/chemotion/collection_api.rb
+++ b/app/api/chemotion/collection_api.rb
@@ -10,6 +10,10 @@ module Chemotion
       error!(e.message, 403)
     end
 
+    rescue_from Usecases::Collections::Errors::CreateForbidden do |e|
+      error!(e.message, 403)
+    end
+
     resource :collections do
       get '/' do
         own_collections =
--- a/app/api/chemotion/reaction_api.rb
+++ b/app/api/chemotion/reaction_api.rb
@@ -290,7 +290,8 @@ module Chemotion
         # (reaction_id, collection_id) index when the chosen collection is the
         # user's "All" collection (both inserts would otherwise be identical).
         CollectionsReaction.find_or_create_by(reaction: reaction, collection: collection) if collection
-        CollectionsReaction.find_or_create_by(reaction: reaction, collection: all_collection)
+        # nil when the owning account has no "All" collection — a Group never gets one.
+        CollectionsReaction.find_or_create_by(reaction: reaction, collection: all_collection) if all_collection
         CollectionsReaction.update_tag_by_element_ids(reaction.id)
 
         if reaction
--- a/app/javascript/src/apps/mydb/collections/CollectionSubtree.js
+++ b/app/javascript/src/apps/mydb/collections/CollectionSubtree.js
@@ -92,12 +92,24 @@ function CollectionSubtree({
       UIActions.toggleCollectionManagement();
     }
 
+    // A locked node still toggles, since the locked containers are the ones worth folding away,
+    // but it must navigate too: "transferred" is locked and has no sub-collections, so toggling
+    // alone would leave the elements a gate transfer moved there unreachable. Expand on the way in
+    // and never collapse — folding the subtree shut on the node being selected reads as a failed
+    // click, and the chevron already has its own handler.
     if (node.is_locked) {
-      toggleExpansion(e, node);
+      if (!visible) toggleExpansion(e, node);
     } else {
       setVisible(visible || isVisible(node, uiState.currentCollection));
-      aviatorNavigationWithCollectionId(node.id, element?.type, (element?.isNew ? 'new' : element?.id), true, true);
     }
+
+    // The shared-with-me tree's owner rows are synthetic grouping nodes with no collection behind
+    // them (CollectionsStore.setSharedWithMeCollections gives them id 0). Keyed on the id rather
+    // than on is_locked, which is only incidentally true of them: navigating one would request
+    // /collections/0, 404, and clear the current collection.
+    if (!node.id) return;
+
+    aviatorNavigationWithCollectionId(node.id, element?.type, (element?.isNew ? 'new' : element?.id), true, true);
   };
 
   const handleTakeOwnershipKeyDown = (e) => {
--- a/app/javascript/src/apps/mydb/collections/ModalExportCollection.js
+++ b/app/javascript/src/apps/mydb/collections/ModalExportCollection.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import AppModal from 'src/components/common/AppModal';
 
 import { StoreContext } from 'src/stores/mobx/RootStore';
+import { SYSTEM_LABELS } from 'src/stores/mobx/CollectionsStore';
 import UserStore from 'src/stores/alt/stores/UserStore';
 
 function ModalExportCollection({ onHide }) {
@@ -61,7 +62,7 @@ function ModalExportCollection({ onHide }) {
     const nodes = collections.map((collection) => {
       const nodeKey = `${collection.id}-${collection.label}`;
 
-      if (collection.is_locked && !['All', 'chemotion-repository.net', 'transferred'].includes(collection.label)) {
+      if (collection.is_locked && !SYSTEM_LABELS.includes(collection.label)) {
         return (
           <li key={nodeKey}>
             <h6>{collection.label}</h6>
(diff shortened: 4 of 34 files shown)


PiTrem:
Force-pushed: history rewritten from 8 commits to 4, each a standalone change rather than a correction of an earlier one on the branch. The resulting tree is identical to the previous head (`git diff 794799a2 eb2896c4` is empty).

Two substantive changes since the first version of this branch, both worth a second look:

- The repository root is now identified by **ownership** (`OwnCollectionEntity` exposes `user_id`) instead of by label + `is_locked`. A legacy group-owned root is locked too, and both are roots with no position, so the previous tie-break was effectively database row order.
- **Every** repository subtree is kept out of `own_collections`, not just the one being rendered. Dropping a second root while still routing its children by ancestry sent them into the collection-management payload, which `UpdateTree` now refuses to reparent — that combination would have failed every tree save with a 403.

PiTrem:
Force-pushed again: dropped the self-healing `tr_col.update_column(:is_locked, true)` from `MoveToCollectionJob` (and its `rubocop:disable`).

It was added in response to the first review round, on the reasoning that rows the backfill missed would stay unlocked forever. That reasoning no longer holds — the migration's lock step was since changed to `is_locked IS NOT TRUE` (so NULL rows are covered) and never constrained the repository root's own ancestry (that predicate belonged to the repair statement, which was removed). What was left for the healing line was a rolling-deploy window the migration closes minutes later, and cases like an unlocked repository root that are a different bug and better left visible than silently repaired.

The invariant now has a single owner: the migration for existing rows, `find_or_create_by`'s block for new ones.

PiTrem:
Force-pushed. Two new commits and a round of review fixes; still six commits, each standalone.

**New scope, flagged because it is a breaking semantic change:** `own_collections_for` / `owned_by?` no longer count a group's collections as its members'. That conflation is what pulled a group's locked rows into each member's payload in the first place, so the two halves belong together — but it also means a group member can no longer rename, delete, re-parent, share onward or withdraw from a group's collection except through a share. Group membership now grants access exclusively as a sharee, which is what `CollectionShare.shared_with` already implemented.

Worth a careful look, since a review round found each of these:

- `ElementsPolicy` used **one** id list for both the ownership test and the share recipients. Narrowing it naively broke group-derived shares; it is now split into `own_user_ids` and `shared_with_ids`. Before the split, a member holding only `:read_elements` passed bulk update and destroy as an owner.
- The "All" guard is on `Group`, not `Person`: `Admin < User`, and admins do hold and use an "All" collection — a Person-only guard would have left new admins without one and crashed sample import on `collections << nil`.
- "Owner has no All" is now reachable, so `reaction_api`, the SBMM sample and device-description use cases and wellplate create nil-check it (the other four call sites already did).
- Both cleanup migrations were missing `collections_vessels` from their emptiness check, and all three `down` methods reported success while dropping the `schema_migrations` row — they now raise `IrreversibleMigration`.

The repository multi-root machinery from the previous revision is gone: narrowing ownership made it unreachable, and its comments described a payload contract this PR removes.

PR #3458, opened 2026-08-12
feat(reaction-variations): autofill sample values from a linked analysis dataset

StarmanMartin:
### Summary

- Adds a **"Populate samples from data file"** action to the analysis-linking modal on the
  Reaction Variations tab. When a linked analysis dataset carries a `reaction_variation.json`
  attachment, the button (or a dropdown, if a reaction has several qualifying datasets) parses
  it and proposes autofilled values for the variation row.
- Samples in the JSON are matched to reaction materials by identifier (short label, external
  label, name, molecule formula, IUPAC name, or sum formula), and applied to the first
  matching column with a compatible unit.
- Proposed values are shown in a new confirmation modal (`AutofillVariationSamplesModal`) —
  listing each material/value/unit pair — before anything is written. This modal is rendered
  as a sibling of the ag-grid instance rather than inside the analysis-linking cell editor,
  because applying the values bumps `gridVersion` and remounts the grid, which would otherwise
  tear the modal down before the user could confirm.
- Adds `AttachmentFetcher.loadAttachmentContent`, exposing the raw attachment response so the
  frontend can parse the JSON payload itself.

### Incidental fixes bundled into this branch

- Null-safe `hasComment` checks (wrap in `Boolean(...)`) across the Sample / Reaction / Screen
  / Wellplate / CellLine / ResearchPlan analyses-tab containers.
- Guards against a missing `cellData` in `MaterialFormatter` and `cellIsEditable`
  (`ReactionVariationsComponents.js`, `ReactionVariationsMaterials.js`).
- `shallowCamelizeKeys` now returns non-object input as-is instead of throwing, so a failed
  fetch (which resolves to `undefined`) doesn't blow up in a fetcher's `.then()`.
- `ColumnSelection` re-syncs its local state when `selectedColumns` changes externally.
- `TextTemplateActions#fetchTextTemplates` defaults a missing dispatch result to `{}`.
- `AnalysesCellRenderer`/`AnalysisOverlay`/`AnalysisVariationLink`/`AnalysesCellEditor`
  converted from function declarations to arrow-function consts (style-only).

### Unrelated / generated

- Rubocop `annotate`-style schema-comment refresh (column/index reordering only, no schema
  change) across `api_token.rb`, `attachment.rb`, `measurement.rb`, the `*_text_template.rb`
  hierarchy, and their specs, plus `reaction_spec.rb` / `molecule_name_spec.rb`. These read as
  drive-by `bundle exec annotate` output rather than part of this feature and could be split
  out of the PR.
- `attachment.rb`'s schema comment now has a duplicated index block (three
  `index_attachments_on_*` lines repeated) — looks like a stale `annotate` artifact, not the
  JSON-autofill feature.

### Test plan

- New JS coverage for `AutofillVariationSamplesModal` (empty state, sample listing,
  confirm/cancel wiring) in `ReactionVariationsAnalyses.spec.js`.
- `spec/javascripts/setup.js` now stubs `global.fetch` to a default empty `200` response so
  unrelated `componentDidMount` fetches in other specs don't flood output with stack traces.
- No new Ruby spec coverage for the feature itself (the touched model specs are
  schema-comment-only diffs).


headri (approved):

app/javascript/src/apps/mydb/elements/details/reactions/variationsTab/ReactionVariations.js
```diff
@@ -442,6 +449,158 @@ const ReactionVariations = ({ reaction, onReactionChange }) => {
     }
   }, []);
 
+  /*
+  Resolve the material a `reaction_variation.json` sample identifier refers to.
+  Labels are tried in `SAMPLE_LABELS_WITH_SUM` order rather than material order, so an
+  identifier that is one material's `short_label` and another's `molecule_formula`
+  resolves to the former instead of to whichever material happens to come first.
+  */
```
JanCBrammer, line 457: I feel like this approach is brittle and prone to misidentifying samples. Wouldn't it be preferable to identify the samples uniquely and explicitly by their IDs?

app/javascript/src/apps/mydb/elements/details/reactions/variationsTab/ReactionVariationsAnalyses.js
```diff
@@ -11,12 +11,21 @@ import {
...
 } from 'src/apps/mydb/elements/details/reactions/variationsTab/ReactionVariationsUtils';
+import AttachmentFetcher from 'src/fetchers/AttachmentFetcher';
+import { rootStore } from 'src/stores/mobx/RootStore';
 
 function getReactionAnalyses(reaction) {
   const analyses = reaction.analysisContainers?.() ?? [];
   return cloneDeep(analyses).filter((analysis) => !analysis.is_new);
 }
 
+const SINGLE_HEADER = {
```
JanCBrammer, line 22: Inline `SINGLE_HEADER` or declare it closer to where it's used.

JanCBrammer (commented):
Please provide ocumentation (can be in the PR description) on how to manually test this feature in the UI.
	
Please provide a Bagit archive containing a reaction_variations.json for manual testing. The archive could be used as a fixture for units tests as well.
	
Is there a specification of reaction_variation.json? From the PR it's not evident how reaction_variation.json is supposed to look like.
	
Note that this feature has to be re-implemented when #3530 is merged.

PiTrem:
@JanCBrammer — thanks, all three asks are fair. Here is the specification, plus the part that
was missing from the PR description: **this file is not meant to be written by hand — it is a
derivative produced by `chemotion-converter-app`.**

---

## 1. Where the file comes from

`chemotion-converter-app` ≥ **1.9.1** already emits it. When a converter *profile* has its
**"Reaction Variations values"** configured, `JcampZipWriter` adds one entry to the BagIt
archive it returns:

```
<archive>.zip
├── bagit.txt
├── manifest-sha256.txt / manifest-sha512.txt
├── data/table_01.jdx …
└── metadata/
    ├── converter.json
    ├── rdf.ttl
    └── reaction_variation.json   ← written only when the profile yields ≥ 1 triplet
```

(`converter_app/writers/jcampzip.py`, payload built by
`Converter.get_reaction_variation_matches` in `converter_app/converters.py`.)

The profile side is configured in the converter admin UI (ELN → *Converter* admin → edit
profile → **"Reaction Variations values"** tab). It is a three-column table —
**Sample / Amount / Unit** — where each row is one triplet and each cell is bound to its own
*identifier*, resolved per uploaded file from **file metadata**, **table metadata**, or a
**table-header regex**.

So the intended user workflow is: *upload the instrument file → the converter recognises it →
the JSON is generated automatically*. The end user never authors it.

---

## 2. Specification of `reaction_variation.json`

A JSON object with a top-level `samples` array. Each entry is a **three-element array**
`[identifier, value, unit]`:

```json
{
  "samples": [
    ["S-101",        1.5, "g"],
    ["Benzaldehyde", 2.0, "mmol"],
    ["Toluene",      5,   "ml"],
    ["C7H6O",        0.5, "mol/l"],
    ["H2",           486, "ppm"]
  ]
}
```

- **`identifier`** (string, required) — matched against the reaction's materials, trying these
  sample fields **in this order**, first hit wins: `short_label` → `external_label` → `name` →
  `molecule_formula` → `molecule_iupac_name` → `sum_formula`. The order is by *field*, not by
  material. An identifier matching nothing is dropped from the proposal.
  (Re: your inline comment on matching by ID — agreed in principle, but the producer here is an
  instrument file that knows compound *names*, not ELN sample ids, so some name-based resolution
  is unavoidable at this boundary. Happy to discuss narrowing the field list.)
- **`value`** (number or numeric string) — written through the column's normal value parser, so
  it goes through the same validation and derived-quantity recomputation as a hand-typed edit.
- **`unit`** (string, required) — this is what selects **which column** the value lands in. A
  `null`/missing unit is not a wildcard; the entry is skipped.

**Unit → column mapping** (first column of the matched material whose unit list contains the
given unit):

| Unit | Column | Applies to |
|---|---|---|
| `g`, `mg`, `μg` | Mass | starting materials, reactants, products |
| `mol`, `mmol` | Amount | starting materials, reactants |
| `l`, `ml`, `μl` | Volume | starting materials, reactants, solvents, products |
| `mol/l` | Concentration | starting materials, reactants |
| `ppm` | Concentration | gas materials (gas mode on) |
| `%` | **Equivalent** (special-cased) | starting materials, reactants — *not* the reference material |

Temperature and duration are *row properties*, not material columns, so they cannot be set from
this file.

⚠️ The filename is **singular**: `reaction_variation.json`. `reaction_variations.json` (as in
your comment) is never matched — the converter writes the singular name too, so the two sides
agree.

---

## 3. Worked example — GC-BID (the primary use case)

The shipped profile *"GC BID extended"* already exposes one metadata group per detected peak on
**table 1**:

| Metadata key | Example | Meaning |
|---|---|---|
| `Ch1.Id 1.Name` | `H2` | compound name of peak 1 |
| `Ch1.Id 1.R.Time` | `2.477` | retention time |
| `Ch1.Id 1.Conc.` | `486.444` | concentration, **ppm** |
| `Ch1.Id 1.Conc. %` | `41.515` | concentration, area-% |

One *Reaction Variations* row per peak `N`:

- **Sample** → table metadata, table 1, key `Ch1.Id N.Name`, match *any*
- **Amount** → table metadata, table 1, key `Ch1.Id N.Conc.`, match *any*
- **Unit** → an identifier resolving to the literal string `ppm`

which lines up with the consumer side, where `ppm` routes to the **Concentration** column of
**gas materials**. GC-BID + gas mode is the feature's primary use case; `g` / `mol` / `ml` / `%`
cover other instrument classes.

---

---

## 4. Short how-to (user-facing)

**Prerequisites:** a reaction you can edit, with the materials present in the **Scheme** tab, and
the Reaction Variations tab enabled. For GC-BID data, switch the reaction to **gas mode** and make
sure the gas materials' names/short labels match the GC peak names (`H2`, `CO`, …).

1. **Get the file.** Normally the converter produces it: upload the instrument file to a dataset
   and let the converter run. *Today* (see gap 2) you have to download the resulting `<name>.zip`
   from the dataset and take `metadata/reaction_variation.json` out of it — or write the JSON by
   hand following the spec above.
2. **Attach it to an analysis dataset.** Reaction → **Analyses** tab → add an analysis → add a
   dataset → open it → **drag the bare `reaction_variation.json` onto the dataset's attachment
   drop area**. Do *not* use the bulk "Upload files" dropzone on the Analyses tab: all three of
   its modes re-zip their input, so the attachment ends up named `something.zip` and the button
   never appears. If you start from a BagIt bag, unpack it first.
3. **Save the reaction.** Required twice over: unsaved analyses are filtered out of the linking
   modal, and the attachment needs a server-side id before its content can be fetched.
4. **Open the Reaction Variations tab,** add a variation row if there is none, and make sure the
   **Analyses** column is visible (column selection → *Metadata → Analyses*).
5. **Double-click the Analyses cell** of that row. The linking modal opens, listing every saved
   analysis with a checkbox.
6. **Tick the checkbox** for the analysis holding your dataset. A **"Populate samples from data
   file"** button appears next to it (a dropdown, if several datasets qualify). It stays disabled
   until the checkbox is ticked. *No button?* → check the filename, that the JSON is a direct
   attachment of a **dataset** (not of the analysis, not inside a zip), and that the reaction was
   saved after the upload.
7. **Click it.** A confirmation modal lists each proposal (`Set Starting material: S-101 to 1.5 g`)
   before anything is written. Confirm, and the matching columns are un-hidden, switched to the
   unit from the file, and filled; derived quantities recompute as for a manual edit.
8. **Save and reload** to confirm the values and the link survive the round-trip.

**Known rough edges** — the confirmation modal currently lists these, then applies nothing and
says nothing: a **product** with `mol`/`mmol` (Amount is read-only for products), a **product**
with `%` (products have no Equivalent column), `%` on the **reference material** (Equivalent is
read-only), and a **solvent** with a non-volume unit. Cheapest fix is to filter them in
`handleAutofill` rather than bail at write time — the eligibility predicate in
`ReactionVariations.js` is already exactly what is needed. I would fix this before merge.

---

Re: your remaining two asks — the BagIt fixture is a code change on the branch (the JSON above is
ready to be used as its payload), and yes, this will need porting once #3530 lands; worth deciding
explicitly whether this merges first and pays the port, or waits.

PiTrem (approved):
Reviewed at b0eca30ddb: CI fully green, xhigh review findings fixed in place, remaining items recorded as follow-ups. Merging with the known caveat that the converter-fed path needs the pending labimotion bump.

PR #3459, opened 2026-08-12
fix(api): eliminate N+1 comment count query on list endpoints

headri:


PiTrem (approved):
Reviewed via /code-review xhigh — N+1 fix confirmed correct against a live DB, all list endpoints preload :comments before comment_count reads it. Two non-blocking style nits noted (duplicated comment_count impl, non-YARD comment shape); unrelated pre-existing sbmm list_order bug filed separately as ComPlat/chemotion#606. CI green, mergeable.

PR #3461, opened 2026-08-13
fix(collections): propagate share rename, permission, and revoke changes to sharees

headri:
All shared collection changes produce server-side notifications.
Fixed security issue in message creation.


PiTrem (approved):
CI green; net change (server-side share/rename notifications + sharee tree propagation fixes) matches the reviewed intent, later commits are incremental fixes on the same feature.

PR #3462, opened 2026-08-13
fix(collections): make the element collection-label badge legible and its warning owner-only

PiTrem:
The collection badge on an element had flattened to two bare counts separated by a pipe, with "this element is also in someone else's collection" signalled only by a colour change and a hover title — unreadable on a touch device, invisible to a colour-blind reader, and with no way to tell *also shared with me* apart from *also somewhere I can't see*.

####  Single ownership as owner
image

####  Single ownership as sharee
image

#### Multiple ownerships - reachable collection
image

#### Multiple ownerships - unreachable collection
image

### The badge

Each part now renders only when it applies:

- **`▤ 3 ↗ 1`** — in 3 of your own collections, 1 of which you have shared out. The sub-count is read off `Collection.shared`, which the store already fetches, so no backend change.
- **`↘ 2`** — also in 2 collections shared with you. Directional glyphs, so the two share directions can't be mistaken for each other.
- **`⚠`** — the element is also in a collection you cannot open.

Every chip carries a `title` spelling its count out in words, and the triangle is drawn persistently rather than by colour alone.

### Who gets the warning

Gated on ownership: own a collection holding the element and you are told it also sits somewhere you cannot see; own none of them and you are not. That is the same line `ElementPolicy` draws everywhere else — `record_is_in_own_collection?` is what every capability resolves to — so the badge stops inventing a second notion of ownership.

Who holds the hidden collection deliberately does not narrow it further. An intermediate version suppressed the warning whenever the viewer held *any* share for the element; on production-shaped data an unrelated colleague's share then masked a genuine third party holding the element, which is the one thing an owner cannot discover another way. Telling those apart would need the hidden collection's owner, and `collection_labels` carries `{id}` and nothing else.

Co-presence in a collection the viewer *can* open is no longer a warning at all — the shared count already says so. It keeps a plain line in the popover, because removing the element from your own collection still won't remove it from theirs.

### The popover

Grouped by owner (keyed on the abbreviation, so two people with the same plain name stay apart), sorted by owner then label, retitled **Shared with me by**. Rows carry the sidebar's own icon and its tooltips — who an own collection is shared with, and how a shared one reaches you, group included.

That last part retires a constraint this file used to document: the group behind *(via group)* was left unnamed because resolving it needed a fetch per badge. An overlay isn't mounted while hidden and `Dropdown.Menu` doesn't render rows until first opened, so a table of badges fetches nothing until someone opens one popover and hovers one row. A spec pins that, and fails if the menu ever gains `renderOnMount` or the fetch is hoisted.

### Alongside

- `ownCollectionIds` omitted the repository subtree, so `isOwnCollection` contradicted `Collection#owned_by?` for exactly those collections and the badge called the viewer's own `transferred` one they can't access. Folded back in. `ElementDetailSortTab`, the only other caller, gains the tab-layout editor inside the repository section, which follows from the same definition.
- The detail header rendered the badge at `btn-sm btn-secondary` beside `btn-light btn-xxsm` labels, and in a different order from the list row. `ElementDetailCard` gains a `titleAppendixLeading` slot so a header can reproduce its list order without every caller having to render the badge itself — six of the nine detail components pass no appendix at all and would otherwise lose it.
- **Separate commit:** deleting a user is a soft delete and nothing purges the shares they were on, so `belongs_to :shared_with` resolved to `nil` through the paranoia scope and `CollectionShareEntity#shared_with` dereferenced it — a 500 on every read of that collection's shares, hit by the sidebar hover, the Manage-shares modal (leaving the orphan unrevokable), and the element list's prefetch. Scoped `with_deleted`.

### Testing

1536 JS specs pass; the badge gains its first spec file (18 cases) plus two `ownCollectionIds` cases on the store spec and two on `collection_share_api_spec`. Behaviour was checked against a production-sized dataset throughout — the ownership gate, the repository-subtree fix, and the dual-ownership cases each have real accounts behind them.

### Known gaps

- If another user's only link to the element is their locked **All**, nothing warns: `collection_tag` drops every locked All, so no entry for them exists to reason about. Closing it needs a tag rebuild.
- `collection_labels` is not filtered per viewer, so the ids ship regardless of what the badge renders. Filtering them server-side would close that and subsume the rule here.
- Sample and SBMM detail headers omit labels their list rows show (wellplate/generic, and reaction respectively). Follow-up.

### DIFF
--- a/app/api/entities/collection_share_entity.rb
+++ b/app/api/entities/collection_share_entity.rb
@@ -18,6 +18,11 @@ module Entities
     expose! :sequencebasedmacromoleculesample_detail_level
     expose! :wellplate_detail_level
 
+    # Both dereference the association unguarded, and may: collection_shares.shared_with_id has a
+    # FK to users, so the row cannot outlive the account, and the association is declared
+    # +with_deleted+ so a *soft*-deleted sharee still resolves. Before that, a deleted sharee made
+    # this nil and every reader of the share 500ed. Keep them non-nil strings regardless — the
+    # frontend's mobx model declares both as required types.string.
     def shared_with
       "#{object.shared_with.name} (#{object.shared_with.name_abbreviation})"
     end
--- a/app/javascript/src/apps/mydb/elements/details/ElementDetailCard.js
+++ b/app/javascript/src/apps/mydb/elements/details/ElementDetailCard.js
@@ -24,6 +24,7 @@ export default function ElementDetailCard({
   isPendingToSave,
   title,
   titleTooltip,
+  titleAppendixLeading,
   titleAppendix,
   headerToolbar,
   headerBanner,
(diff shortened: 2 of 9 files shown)


TasnimMehzabin (approved):
Looks good to me

PR #3463, opened 2026-08-13
fix(collections): restrict kebab-menu and metadata/RADAR actions on shared collections to owners

headri:


PiTrem:
Pushed three commits onto this branch — sorry for landing on your PR without asking first. All three build on what's already here; nothing existing was rewritten or reverted. Happy to move any of it out if you'd rather.

**1. `fix(collections): stop reporting a rejected metadata save as success`** — the one that matters.

Once the metadata endpoints started refusing people, the refusal needed to be visible, and it wasn't. `MetadataContainer.handleSave` fired `storeMetadata` and then called `metadata.updateChecksum()` on the next line, synchronously, before the request had left the browser. That clears the dirty flag, the dirty flag drives the Save button, so the button vanished whatever the server answered — a 200 and a 403 looked identical and the edit was lost silently. Three layers each swallowed it: `MetadataFetcher.store` passed no options so the client defaults applied (no `response.ok` check, rejections turned into `undefined`), `ElementActions.storeMetadata` ended in `.catch(console.log)`, and `handleSave` never looked at the result. All three now surface it, with an error toast through the notifications store the component already had in context.

The same commit drops the `rescue_from RecordNotFound -> 401` that sat between the two metadata endpoints. It was dead before this branch (the old bodies used `Metadata.where`, which returns nil rather than raising) and the new `.find` calls resurrected it. Grape scopes a `rescue_from` to the endpoints declared *after* it, so the same "not yours or not there" condition answered **404 on GET and 401 on POST** — and anything appended below would have inherited the 401. Both now fall to the class-level 404, which your GET spec already pinned; I updated the POST spec to match.

**2. `fix(literature): report a failed collection-reference request`** — the client half of the same acceptance criterion your `rescue_from` addressed. With the JSON 404 the API now returns, `ApiClient` *resolves* rather than rejects, so `fetchReferencesByCollection` destructures four `undefined`s and throws on `collectionRefs.map` — the same TypeError, one line further down, and the Reference Report modal still opens blank with nothing shown.

**3. `chore(literature): drop two dead guards and cover an inaccessible collection`** — `error!(404) unless @c` can't run (`set_var` uses `.find`, which raises) and would have answered 500 anyway, since Grape's signature is `error!(message, status)`. Your `rescue_from` makes both permanently unreachable. Also added the case your literature spec was missing: a collection that *exists but isn't accessible*, which is what a real user hits — `accessible_for(...).find` raises for it exactly as for a missing id.

PiTrem (approved):
Blocking finding from the prior review pass (rejected metadata save silently reporting success) is fixed in 64e9ab4d2c, and both suggestions S2/S3 (dead literature guards, inaccessible-collection spec) are addressed in b8e37ec75b. Verified by diff, not just commit message. CI green.

PR #3470, opened 2026-08-17
fix(reports): prevent Preview/Label and docx crashes, restore SI-Synthesis analyses

### LINKED ISSUE #3169: Report generation breaks for samples
Select sample from sample list via checkbox ->
Click "Report" in reporting dropdown -> 
Click on the "Label" or "Preview" tab in the "Report" element -> 
Application crashes with:

```
TypeError: Cannot read properties of undefined (reading 'map')
at analysesParagraph (SectionSample.js:48:29)
at AnalysesContent (SectionSample.js:58:30)
```

The error occurs on 18fec19813fb46e0e9db52bc44107da6958a0d30.

### LINKED ISSUE #3170: Report generation breaks for reactions
Select reaction from reaction list via checkbox ->
Click "Generate" in "Report" element ->
Report generation fails with various errors:

e.g., ```Report could not be generated: undefined method `/' for "":String```
or ```Report could not be generated: invalid value for BigDecimal(): "n.d"```

The error is caught in 

https://github.com/ComPlat/chemotion_ELN/blob/18fec19813fb46e0e9db52bc44107da6958a0d30/app/api/chemotion/report_api.rb#L250

The error occurs on 18fec19813fb46e0e9db52bc44107da6958a0d30.

TasnimMehzabin:
## Summary

Fixes two independent crashes in report generation, one on the sample side
(front end) and one on the reaction side (back end).

Closes ComPlat/chemotion_ELN#3169
Closes ComPlat/chemotion_ELN#3170

## 1. Sample report crash on Label/Preview tab (#3169)

Opening the **Label** or **Preview** tab of the Report element for a sample
crashed with:

TypeError: Cannot read properties of undefined (reading 'map')
at analysesParagraph (SectionSample.js:48)

**Root cause.** In `AnalysesContent.analysesParagraph`, the reduce seed
`init` is `reactionDescription.ops` when the "reaction description" setting is
on and `reaction_description` is an object. When that object lacks a usable
`.ops` array (a malformed / non–Quill-Delta shape), `init` is `undefined`. With
zero analyses, `analyses.reduce(fn, init)` never calls `fn` and returns `init`
unchanged, so `dataMerged` is `undefined` and `.map()` throws. An empty
`analyses` array is truthy, so the paragraph builder runs even with no analyses.

**Fix.** Guard both ends in `SectionSample.js`:
- `reactionDescription.ops || []` — closes the reported crash directly.
- `(dataMerged || []).map(...)` — insurance at the consumption point.

## 2. Reaction docx report crash (#3170)

Generating a docx report for a reaction failed (silently, via the generic
rescue in `report_api.rb`) with:

undefined method `/' for "":String
Float can't be coerced ... / invalid value for BigDecimal(): "n.d"

**Root cause.** Mixture-component and product fields can arrive as strings (or
blank) from stored JSON, but two spots ran arithmetic on the raw, un-coerced
value:
- `calculate_mixture_amount_mol` guarded the branch with
  `rel_mol_weight&.to_f&.positive?` but then divided by the **raw** value
  (`mass.to_f / rel_mol_weight`), so a present numeric-string relative
  molecular weight raised `TypeError`.
- The product `equiv` computation ran `(s.equivalent * 100).nan?`, which raises
  when `equivalent` is a non-numeric placeholder.

(The `BigDecimal(): "n.d"` variant is already neutralized by the
`exception: false` hardening of `Chemotion::Calculations.valid_digit` in #3445;
`valid_digit` is left unchanged so placeholders still render verbatim.)

**Fix.** Coerce operands to `Float` before arithmetic in
`lib/reporter/docx/detail_reaction.rb`, so numeric strings compute correctly and
blank / nil / non-numeric values fall through to the safe branch.

## Testing

- Verified both patched Ruby expressions against numeric, numeric-string,
  blank, `nil`, `"n.d."`, and `NaN` inputs — none raise.
- `ruby -c` passes on the modified file.

## Backport

Both call sites are byte-for-byte identical to the v3.x channel, so this
backports cleanly.

### DIFF
--- a/app/javascript/src/apps/mydb/elements/details/reports/SectionSample.js
+++ b/app/javascript/src/apps/mydb/elements/details/reports/SectionSample.js
@@ -43,12 +43,17 @@ const AnalysesContent = ({ show, showRecDes, analyses, reactionDescription }) =>
     const dataMerged = analyses.reduce((sum, a) => {
       const defaultContent = "{\"ops\":[{\"insert\":\"\"}]}";
 
-      const contentJSON = a.extended_metadata.content || JSON.parse(defaultContent);
+      // extended_metadata can be null for legacy rows (hstore column is nullable), so guard with `?.`.
+      const contentJSON = a?.extended_metadata?.content || JSON.parse(defaultContent);
       const ops = Array.isArray(contentJSON.ops) ? contentJSON.ops : [];
       return [...sum, ...ops];
     }, init);
     const data = dataMerged.map(d => {
-      d.insert = d.insert.replace(/\n/g, ' ');
+      // A Quill embed op (e.g. an image) has a non-string `insert` object; only strings
+      // support .replace. QuillViewer strips embeds downstream, so just leave them intact here.
+      if (typeof d.insert === 'string') {
+        d.insert = d.insert.replace(/\n/g, ' ');
+      }
       return d;
     });
     return { ops: data };
--- a/app/javascript/src/apps/mydb/elements/details/reports/SectionSample.js
+++ b/app/javascript/src/apps/mydb/elements/details/reports/SectionSample.js
@@ -36,17 +36,18 @@ const SVGContent = ({ show, svgPath }) => {
 }
 
 const AnalysesContent = ({ show, showRecDes, analyses, reactionDescription }) => {
-  const isReDesObj = typeof reactionDescription === "object";
+  const isReDesObj = typeof reactionDescription === 'object';
   const reDesOps = showRecDes && isReDesObj ? reactionDescription.ops : null;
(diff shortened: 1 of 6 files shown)


PiTrem (approved):
Re-reviewed at 6c726e79. All blocking findings from the previous round are fixed and verified: the extended_metadata.report String/Boolean comparison (which silently emptied the SI-Synthesis analyses paragraph), the unguarded data/extended_metadata dereferences, non-string Quill embed inserts, the in-place ops reversal, and the render-time prop mutation. Ruby and JS suites green in CI and verified locally (RSpec 17/0, Mocha 8/8). Remaining items are non-blocking follow-ups, tracked in the review artifact.

PR #3473, opened 2026-08-17
fix(components): coerce relative_molecular_weight to Float at the write boundary

TasnimMehzabin:
`relative_molecular_weight` was the only numeric component field not declared in the
components params block, so a client-sent string was persisted uncoerced. That string
then crashed both the mixture DOCX report (`Float / "150.0"` → TypeError) and the
component write path (`"150.0" > 0` → `ArgumentError: comparison of String with 0
failed`). This fixes the root cause at the API boundary and hardens the shared usecase.

### Changes
- **`component_api.rb`:** declare `optional :relative_molecular_weight, type: Float`
  alongside its numeric siblings, so Grape coerces numeric strings and rejects
  non-numeric input on the standalone component API.
- **`Components::Create`:** compare with `existing_relative_mw.to_f.positive?` instead
  of `existing_relative_mw > 0`. The reaction material-update flow reaches this usecase
  via `Reconcile` *without* Grape coercion, so the API declaration alone doesn't cover
  it; this also stays safe for legacy rows that already hold string values.

### Tests
- `component_api_spec`: a numeric-string `relative_molecular_weight` is coerced to and
  stored as a Float.
- `create_spec`: numeric-string and blank-string values no longer raise and behave
  correctly (preserve vs. recalculate).

### DIFF
--- a/app/api/chemotion/component_api.rb
+++ b/app/api/chemotion/component_api.rb
@@ -42,6 +42,7 @@ module Chemotion
             optional :amount_mol, type: Float, desc: 'Component moles'
             optional :amount_l, type: Float, desc: 'Component volume'
             optional :amount_g, type: Float, desc: 'Component mass'
+            optional :relative_molecular_weight, type: Float, desc: 'Relative molecular weight in g/mol'
             optional :density, type: Float, desc: 'Density in g/ml'
             optional :molarity_unit, type: String, desc: 'Molarity unit'
             optional :molarity_value, type: Float, desc: 'Molarity value'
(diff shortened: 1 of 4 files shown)


PiTrem (approved):
Both prior blocking findings are genuinely fixed at `f90bd8d9`: the usecase now writes the coerced Float back instead of preserving the raw String, and the spec asserts the correct contract (`250.0`, not `'250.0'`). CI green including `testing`. Remaining suggestions are pre-existing scope that applies equally to `main` today.

PR #3476, opened 2026-08-18
fix(net): allow preferring one IP address family for outbound requests

PiTrem:
## Problem

On a deployment whose host and containers have addresses in only one family — typically
IPv6-only, with no IPv4 address and no IPv4 route — every backend-initiated outbound
request can fail, while the same hosts are reachable over IPv6 from inside the container
(`curl -6` succeeds). The frontend is unaffected; only server-side requests break.

It hits every integration that goes through `Net::HTTP` (directly, or via `httparty` /
`faraday`): PubChem, CAS lookup, SciFinder-n, DataCite, NMRShiftDB, RADAR, UniProt, and
the repository transfer jobs.

## Cause

Ruby calls `getaddrinfo` **without `AI_ADDRCONFIG`**, so glibc does not filter out
addresses of a family the host cannot use. Measured on an IPv4-only host against a
dual-stack name:

```
flags=0 (Ruby's default) -> {"AF_INET"=>8, "AF_INET6"=>8}
AI_ADDRCONFIG explicit   -> {"AF_INET"=>8}
```

The unusable family is offered regardless of connectivity. Symmetrically, on an IPv6-only
host Ruby is handed the A records for every external service despite having no way to
route them.

## Change

Setting `ELN_PREFER_ADDRESS_FAMILY` to `ipv6` or `ipv4` moves name resolution off glibc
and onto Ruby's `Resolv`, then selects the first address of the preferred family — falling
back to stock resolution when the host publishes none.

```
ELN_PREFER_ADDRESS_FAMILY=ipv6
```

- `lib/chemotion/address_family_preference.rb` — configuration and address selection
- `config/initializers/address_family_preference.rb` — gated installer
- `.env.production.example` — documents the variable and its caveat

The initializer requires `resolv-replace` explicitly. That stdlib shim is what routes
`TCPSocket` (and therefore `Net::HTTP`) through `Resolv`; without it, patching `Resolv`
has no effect on outbound HTTP at all.

## Safety

**Disabled unless the variable is set**, and deliberately not enabled by default. Enabling
it removes glibc's RFC 6724 destination sorting, so on a dual-stack host a service whose
preferred-family address is published but unreachable would no longer fall back to the
other family. `/etc/hosts` is still honoured (`Resolv::Hosts`).

An unrecognised value disables the feature rather than guessing at it.

The change is also self-limiting: `resolv.rb` only queries AAAA when the host has an IPv6
address of its own (`Resolv::DNS#use_ipv6?`), so enabling `ipv6` on a host without one is
a no-op rather than a breakage.

## Testing

`spec/lib/chemotion/address_family_preference_spec.rb` — 13 examples covering
configuration parsing, family selection, the single-family fallback, `Resolv` address
objects, unparseable input, and the resolver override itself. Rubocop clean.

Verified end to end in the dev container:

| `ELN_PREFER_ADDRESS_FAMILY` | `enabled?` | `resolv-replace` loaded | `TCPSocket#initialize` |
|---|---|---|---|
| unset | `false` | `false` | stock |
| `ipv6` | `true` | `true` | `resolv-replace.rb` |
| `inet6` (invalid) | `false` | `false` | stock |

### DIFF
--- a/.env.production.example
+++ b/.env.production.example
@@ -83,3 +83,11 @@ SENTRY_FRONTEND_SAMPLE_RATE=1.0
 # DELAYED_JOB_ARGS=""
 # PG_CARTRIDGE=rdkit
 # PG_CARTRIDGE_VERSION=4.4.0
+
+# Prefer one IP address family for outbound connections (PubChem, DataCite, RADAR, ...).
+# Needed when the host and containers have addresses in only one family, typically
+# IPv6-only: Ruby resolves without AI_ADDRCONFIG, so unroutable IPv4 addresses are still
+# handed to the connect path and outbound requests fail.
+# @note leave unset on dual-stack hosts — it disables glibc's RFC 6724 fallback ordering
+# @example ELN_PREFER_ADDRESS_FAMILY=ipv6
+# ELN_PREFER_ADDRESS_FAMILY=
--- a/config/initializers/address_family_preference.rb
+++ b/config/initializers/address_family_preference.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require_relative '../../lib/chemotion/address_family_preference'
+
+# Makes outbound connections prefer one IP address family.
+#
+# Needed on deployments whose host and containers have addresses in only one family —
+# typically IPv6-only. Ruby omits AI_ADDRCONFIG when resolving, so glibc happily returns
+# A records for dual-stack services on a host with no IPv4 address and no IPv4 route, and
+# outbound calls to PubChem, DataCite, RADAR and friends fail.
+#
+# Disabled unless ELN_PREFER_ADDRESS_FAMILY is set to "ipv6" or "ipv4":
+#
+#   ELN_PREFER_ADDRESS_FAMILY=ipv6
+#
+# Do NOT enable this on a dual-stack host. It removes glibc's RFC 6724 destination
+# sorting, so a service whose preferred-family address is published but unreachable will
+# no longer fall back to the other family.
+return unless Chemotion::AddressFamilyPreference.enabled?
+
+# Routes TCPSocket (and therefore Net::HTTP) through Resolv instead of glibc. Without
+# this shim, patching Resolv has no effect on outbound HTTP at all.
+require 'resolv-replace'
+
+Resolv.singleton_class.prepend(
+  Module.new do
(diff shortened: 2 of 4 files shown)


PiTrem:
Marking draft pending confirmation of the root cause.

The `AI_ADDRCONFIG` behaviour this PR is built on is measured and real — Ruby resolves without the flag, so glibc returns addresses of a family the host cannot route:

```
flags=0 (Ruby's default) -> {"AF_INET"=>8, "AF_INET6"=>8}
AI_ADDRCONFIG explicit   -> {"AF_INET"=>8}
```

What is *not* yet established is that this is what breaks the affected IPv6-only deployment. The workaround that deployment runs patches `Resolv.getaddress` without loading `resolv-replace`, and `TCPSocket#initialize` there is the stock C implementation — so that patch never reaches the outbound HTTP path, and cannot be evidence that address selection is the problem.

The deciding observation, from the affected host:

```ruby
require 'socket'
p Socket.getaddrinfo('pubchem.ncbi.nlm.nih.gov', 443, nil, :STREAM).map { |a| [a[0], a[3]] }
```

A records returned and sorted first -> this PR addresses the real mechanism. Only AAAA returned -> glibc is filtering correctly there and the failure lies elsewhere, so this PR should not be merged as a fix for it.

The code stands on its own either way (it requires `resolv-replace` explicitly, so unlike the field workaround it actually takes effect), but it should land as a deliberate capability, not as a fix for an unconfirmed cause.

PiTrem:
Closing. The case that motivated this turned out to be a container networking configuration issue — the Docker network the application runs on had no IPv6 enabled, so the container had no working route out by either address family. Enabling IPv6 on that network resolved it, with no application change required.

The behaviour documented here is still real: Ruby resolves without `AI_ADDRCONFIG`, so glibc returns addresses of a family the host may not be able to route.

```
flags=0 (Ruby's default) -> {"AF_INET"=>8, "AF_INET6"=>8}
AI_ADDRCONFIG explicit   -> {"AF_INET"=>8}
```

But no confirmed deployment needs the knob, and a correctly configured single-family host does not hit the problem — `TCPSocket` already iterates across address families on its own, provided the unusable family fails fast rather than black-holing.

Rather than carry an opt-in setting with no demonstrated user, closing this. The branch remains if a genuine case turns up later.

PR #3481, opened 2026-08-20
fix(clipboard): repair dead citation copy button and add feedback to all copy buttons

TasnimMehzabin:
## What & why

The per-reference copy button in the References tab did nothing — it still
carried the `clipboardBtn` / `data-clipboard-text` markup from the removed
`clipboard` npm package, but no handler was ever wired after the package was
dropped. This fixes it and, along the way, hardens copy-to-clipboard behavior
app-wide.

## Changes

- **Fix the dead button** — wire the citation copy button to the shared helper.
- **Harden `copyToClipboard`**
  - Fall back to a legacy `execCommand` path so copy works on plain `http://`
    origins (where `navigator.clipboard` is undefined), not just https/localhost.
  - Return `true`/`false` and **toast on failure** so a silent no-op can never
    look like a success. Every failure resolves `false` (never rejects), even if
    the DOM setup throws.
- **Reusable `CopyButton`** (`components/common/CopyButton.js`) — icon button
  that confirms a successful copy by briefly swapping the clipboard icon for a
  check. Adopted at every icon copy button: SampleDetails (CAS/InChI/SMILES/
  molfile), Required-volume input, reaction duration, device path, literature
  modal, and citations.
- **Route the search-results copy** through the shared helper (it was calling
  `navigator.clipboard` directly, bypassing the fallback + toast).
- Removed the leftover `clipboardBtn` class and now-dead imports/helpers.

## Tests

New specs for the helper (secure path, `execCommand` fallback, throw handling,
failure toast) and for `CopyButton` (icon swap on success, no swap on failure).
Full JS suite green (1617 passing).

## Not included (out of scope)

`LiteratureDetails` (labeled dropdown item), `AuthToken`, and the ketcher
template copy still copy directly; they get the failure toast where they route
through the helper but not the check-icon confirmation.

### DIFF
--- a/app/javascript/src/apps/admin/devices/DeviceDataCollectorTab.js
+++ b/app/javascript/src/apps/admin/devices/DeviceDataCollectorTab.js
@@ -1,7 +1,7 @@
 /* eslint-disable react/prop-types */
 import React, { useState, useEffect, useContext } from 'react';
 import {
-  Form, InputGroup, Tooltip, OverlayTrigger, Button
+  Form, InputGroup
 } from 'react-bootstrap';
 import { Select } from 'src/components/common/Select';
 import { startsWith, endsWith } from 'lodash';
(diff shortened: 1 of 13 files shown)


PiTrem (approved):
no blocking issues, CI green. Two non-blocking suggestions noted (CopyButton unmount guard on in-flight copy; pre-existing dead data-clipboard-text gate in SearchResultTabContent) — fine to land as-is, worth a follow-up.

PR #3486, opened 2026-08-21
fix(import): ignore empty PolymersList tags and stop corrupting molfile headers

phuang26:
Ketcher writes an empty "> <PolymersList>" block for ordinary, non-polymer structures.
Those molfiles were treated as polymers throughout import and rendering, giving every
such row a synthetic POLYMER_<sha> inchikey — collapsing them onto one dummy molecule —
and an Indigo-rendered SVG with polymer-specific options.

Polymer detection now keys on the block's payload rather than the tag's presence, with
the block-scanning rules owned by Chemotion::MolfilePolymerSupport and shared by the
importers and SvgRenderer:

- A data block ends at the next "> <Tag>" header, at "$$$$", or at "M  END", so a
  following "> <TextNode>" is never read as polymer payload and the legacy in-CTAB
  layout ketcher-rails wrote no longer swallows the end-of-CTAB marker into the payload.
- Legacy PolymersList grammars ("0s", "0s/1.00-2.00") parse their atom index instead of
  being dropped, so the shape is still injected.
- Non-polymer rows store CTAB-only molfiles, keeping SDF data blocks away from SvgRenderer.

Molfile headers are no longer corrupted on import: molfiles are rstripped rather than
stripped and the Open Babel padding only appends, so a titled molfile is not shifted
down a line into a blank inchikey.

Molfiles carrying non-UTF-8 bytes are scrubbed rather than raising ArgumentError out of
ActiveSupport's #blank?, which previously aborted the whole import.

### DIFF
--- a/app/api/chemotion/molecule_api.rb
+++ b/app/api/chemotion/molecule_api.rb
@@ -250,8 +250,11 @@ module Chemotion
           svg = KetcherService::SVGProcessor.clean_and_trim_svg(svg) || svg
           svg_process = SVG::Processor.new.structure_svg('ketcher_epam', svg, svg_digest, true)
         else
-          # Molfile has PolymersList tag -> use Indigo first; else Ketcher first; fallback to OpenBabel.
-          svg_service = Chemotion::SvgRenderer.has_polymers_list_tag?(molfile) ? 'indigo' : 'ketcher'
+          # Molfile has an actual PolymersList payload -> use Indigo first; else Ketcher first; fallback to OpenBabel.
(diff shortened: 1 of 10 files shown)


PiTrem (approved):
Polymer detection now keys on PolymersList payload rather than tag presence, with block-scanning centralised in MolfilePolymerSupport and covered by new specs; molfile header handling (rstrip + append-only padding) and non-UTF-8 scrubbing are correct. Remaining pronto warnings are pre-existing metrics offences on touched lines.

PR #3488, opened 2026-08-24
fix(nmr): one nmr zip yields two bagit datasets and only the first curve opens in nmrium

headri:
### Summary

A multi-curve NMR ZIP upload (e.g. two runs sharing one archive) was broken end-to-end: only the first curve ever
reached NMRium, the dataset modal's UI made two legitimate curves look like a duplicated dataset, and (found via
further testing below) a whole separate cluster of AASM/state and duplicate-attachment bugs meant a bagit archive's
curves could get stuck out of the viewer entirely, 500 on Reprocess, silently lose their peak table, or get corrupted/
deleted by a race with the archive's own reprocessing. Fixed in sequence:

- NMRiumDisplayer.js: `trySendUrlsToNMRium` picked the jcamp source with `.find`, so only the first fetched curve was
  ever included in the `spectra[]` payload sent to the NMRium iframe. Switched to `.filter` and build one `spectra[]`
  entry per fetched jcamp file.
- NMRiumDisplayer.js + attachment_jcamp_aasm.rb: `prepareNMRiumDataAttachment` derived the `.nmrium`/`.svg` basename
  via `specInfo.label?.split('.')[0]`, collapsing distinct curves (`x.1_bagit...` and `x.2_bagit...`) onto the same
  basename when a curve was edited and saved back through NMRium. `delete_related_nmrium` scoped its cleanup by
  `filename_parts[0]`, which also dropped the curve token and could prune a sibling curve's `.nmrium` instead of only
  the current curve's stale one. Preserved the `N_bagit` token in the basename, and rescoped `delete_related_nmrium`'s
  match to the same curve. (An initial attempt at this also threaded `arr_nmrium` into `read_bagit_data` to generate a
  `.nmrium` for every curve at upload time — turned out chem-spectra's zip-archive response never contains nmrium data
  for any curve, so that code path was dead on arrival and was reverted; `edit_process`'s existing per-attachment
  `arr_nmrium` handling already covers the real "save an edited curve back as `.nmrium`" case once the basename fix
  above stops it colliding across curves.)
- ContainerDatasetModalContent.js: Processed groups are keyed by `<archive>.<curveIndex>` but were titled
  `Processed: <archive>.N` — two near-identical titles differing only by a trailing digit, reading as a duplicated
  dataset. Added `getProcessedGroupTitle` to label each series `Experiment N / total of <archive>`.

### Note on curve naming (`N_bagit` vs. old `processed_1`)

A reviewer flagged that the old UI (`740.peak.jdx` + `740.processed_1.peak.jdx`) looked like it distinguished a
processed curve from an unprocessed one, and that the new uniform `Experiment 1 / 2` / `Experiment 2 / 2` naming
(`N_bagit`) loses that distinction. The naming change itself isn't a regression, but the underlying concern turned
out to be pointing at something real, so it's worth spelling out precisely:

- Archive 740 is a Bruker experiment folder containing a raw `fid` at the top level plus an already-processed
  `pdata/1/` subfolder (TopSpin's own FFT'd/peak-picked output). Chem-spectra (`bruker.py`,
  `FidHasBruckerProcessed.__read()`) treats these two curves differently at the signal-processing level: curve 0 is
  the raw FID that chem-spectra itself fully processes (remove digital filter, FFT, autophase, baseline correction);
  curve 1 is read from `pdata/1`'s already-FFT'd binary and only gets a light baseline pass, since Bruker/TopSpin
  already processed it before upload. That's a real, meaningful distinction.
- Neither naming scheme ever surfaced it, though. The old `processed_1` suffix came from `original_stem`'s
  blank-filename fallback in the (buggy, LCMS-misclassified) `read_processed_data` path — unrelated to which curve
  chem-spectra fully processed vs. passed through. The new `N_bagit` naming in `read_bagit_data` is purely positional
  (`"#{idx + 1}_bagit"`). So this PR doesn't remove information that used to be visible to the user; that
  distinction has never been exposed anywhere in the UI, old or new.
- It's also not fixable purely on the Rails side. Chem-spectra computes the raw-vs-pdata distinction internally but
  discards it before responding: for this archive shape, the response zip's jcamp/img entries get anonymous
  tempfile names (not the meaningful names the true BagIt/CV/LCMS path already produces), and the
  `x-extra-info-json` header carries only a single `spc_type` scalar (overwritten per curve) and a single
  `invalid_molfile` bool — no per-curve provenance field. The only thing that survives to Rails is ordering (raw
  curve first, then `pdata/N` subfolders in `os.walk` order), which happens to hold for this specific Bruker
  fid+pdata shape but isn't a signal `read_bagit_data` can safely generalize to its other callers (true BagIt
  archives, independently-uploaded multi-experiment zips, etc. also route through it).
- Logging this as a known limitation rather than fixing it here: a real fix needs a chem-spectra-side change (e.g.
  a per-curve provenance field in `x-extra-info-json`, or meaningful entry names on this path like the BagIt path
  already has) plus a Rails follow-up to consume it — out of scope for this PR.

### Additional bugs found while manually testing the fixes above

Retesting upload/Reprocess/peak-editing on real multi-curve archives (both bagit-shaped and plain LCMS-shaped)
surfaced a second, independent cluster of bugs — none of them specific to the naming change above, all reproducible
on `main`:

- **attachment_jcamp_aasm.rb (`jcamp_peak_addon?`)**: a bagit curve's own filename addon is `"<n>_bagit"` — it never
  gains a `.peak.`/`.edit.` addon. That addon matched neither `jcamp_peak_addon?` nor `jcamp_edit_addon?`, so
  `generate_att`'s dispatch fired no AASM event and the row stayed in `:queueing` forever, which
  `extractJcampFiles.js` filters out of the viewer — disabling the spectra editor for the whole dataset. Extended the
  predicate to also match `\A\d+_bagit\z`.
- **attachment_jcamp_aasm.rb (`set_regenerating`)**: because of the addon shape above, `SpectraHelper.js#JcampIds`
  buckets a bagit curve as `"orig"` by filename shape alone, unaware its `aasm_state` is already `:peaked` —
  and `regenerate_spectrum` (`attachment_api.rb`) calls `set_regenerating` on every `"orig"` id. That raised
  `AASM::InvalidTransition` and 500'd the whole Reprocess request. Added `:peaked` to the event's `from:` list.
- **attachment_jcamp_aasm.rb (`read_bagit_data`)**: unlike `create_process`'s plain (non-bagit) branch,
  `read_bagit_data` never transitioned `self` (the archive's own `.zip` attachment) out of `:queueing`/
  `:regenerating` — so it stayed stuck there after every upload or Reprocess. `:regenerating` isn't excluded by
  `extractJcampFiles.js`'s `isApp` check, so the stuck `.zip` (a jcamp-like extension) got fed to the spectra editor
  as if it were JCAMP content instead of being filtered out, producing a broken/blank graph even though the actual
  per-curve attachments it had generated were entirely correct. Added `set_done` at the end of `read_bagit_data`.
- **SpectraHelper.js (`JcampIds`)**: same root cause as the two bullets above — a bagit curve's filename shape alone
  made it indistinguishable from a genuine original upload, so it was classified as `"orig"`. `regenerate_spectrum`
  then resubmitted it independently in the same request that was also reprocessing the archive that generated it,
  racing the archive's own regeneration against the curve's own redundant one — both touching the same rows.
  Confirmed via DB inspection: this spawned duplicate generations and, when the race's last writer was a cleanup/
  destroy call, could delete the curve's attachment rows outright. Classified `N_bagit`-suffixed files as generated
  (`gene`), not original.
- **lib/chemotion/jcamp.rb (`stub_http`)**: HTTParty's multipart encoding sends a `nil`-valued form field and an
  explicit `""` identically on the wire. chem-spectra-app relies on that distinction — `peaks_str` present-but-empty
  means "explicitly cleared", absent means "not specified" (`request.form.get('peaks_str', default=None)`). Without
  compacting `nil`s out of the request body first, every unset param (including `peaks_str` on a plain reprocess)
  still arrived as an empty field, making a plain Reprocess indistinguishable from the user having cleared every
  peak — wiping the EDIT_PEAK table on every reprocess. Added `.compact` before posting.
- **attachment_jcamp_aasm.rb (`generate_att` reuse, `reconnect_dot_style_name`, cleanup task)**: `generate_att`
  already reuses an existing derived-attachment row (by lineage + target filename) instead of minting a new one on
  every save/regenerate, but two related gaps in that logic still let duplicates through:
  - `read_processed_data` (the LCMS multi-curve path) immediately renames a freshly-resolved row from its dot-style
    name to an underscore-style stem, so `generate_att`'s dot-style lookup couldn't find it again on the next
    save/regenerate and minted a duplicate instead of reusing it. Added `reconnect_dot_style_name` to rename an
    existing row back to its dot-style name first so the lookup finds and reuses it.
  - The lookup-then-create had no DB-level uniqueness to fall back on, so two concurrent saves deriving the same
    target filename could both pass the lookup before either committed its insert, minting a duplicate under
    concurrency. Wrapped `generate_att` in a transaction holding a `pg_advisory_xact_lock` keyed by
    `(attachable_id, meta_filename)` for its duration.
  - Added `lib/tasks/cleanup_duplicate_derived_attachments.rake` (`CleanupDuplicateDerivedAttachmentsTask`, dry-run
    by default) to soft-delete stray duplicate rows left over from before this fix, keeping the highest-id row per
    `(attachable_id, filename, lineage)` group — the same identity `generate_att` itself now resolves to.
- **package.json**: bumped `@complat/react-spectra-editor` 1.8.0 → 1.8.2, pulling in fixes filed against
  [ComPlat/react-spectra-editor#329](https://github.com/ComPlat/react-spectra-editor/issues/329) found during the same
  manual testing: a crash reaching into `payload.operation` on an undefined value that cancelled every other
  redux-saga watcher, a self-sustaining reset loop from an unmemoized `extractParams()` recomputing on every render,
  and a `jcampconverter` positional-lookup fix so repeated `##$CSCATEGORY=` LDRs in a peak table aren't collapsed to
  the last value. Companion fixes on the chem-spectra-app side (honoring an absent vs. empty `peaks_str`, and adding
  `.OBSERVE FREQUENCY`/`.SOLVENT NAME` headers to non-SPECTRUM-ORIG peak-table blocks) aren't part of this repo's
  diff but are necessary for the `.compact` fix above and for NMRium to render a curve's frequency/solvent correctly.

### Test plan

- [x] `bundle exec rubocop app/models/concerns/attachment_jcamp_aasm.rb lib/chemotion/jcamp.rb` — no new offenses vs.
  pre-edit baseline (verified by linting the merge-base version of each file for comparison)
- [x] `bundle exec rspec spec/models/attachment_spec.rb spec/models/concerns/attachment_jcamp_aasm_spec.rb spec/lib/cleanup_duplicate_derived_attachments_task_spec.rb`
  — 160 examples, 0 failures, 3 pending (pre-existing "not yet implemented" stubs, unrelated to this PR)
- [x] `yarn eslint` on each changed JS file (`SpectraHelper.js`, `NMRiumDisplayer.js`, `ContainerDatasetModalContent.js`)
  — no new offenses vs. pre-edit baseline (same before/after comparison as above; this repo's `yarn test`/mocha
  harness currently fails on unmodified `main` too with a `$RefreshReg$`/Enzyme adapter error, so JS unit tests were
  verified via standalone Node scripts exercising the changed logic instead — see the added `SpectraHelper.spec.js`
  case, which documents the expected behavior even though the harness can't currently run it end-to-end)
- [x] Manual verification on real archives (`740.zip` — Bruker fid+pdata bagit shape; a plain multi-experiment LCMS
  zip): upload, Reprocess, and repeated peak add/remove/clear-all all confirmed working, including re-opening a
  dataset after each operation
- [ ] Manual verification: upload a multi-curve NMR ZIP and confirm both curves open in NMRium, both get `.nmrium`
  attachments with distinct filenames when edited and saved back, and the dataset modal shows them as separate
  labeled experiments


PiTrem (approved):
Multi-curve NMR zip uploads now work end to end in NMRium, reprocessing and derived-attachment handling; CI green on the final head.

PR #3494, opened 2026-08-24
fix(spectra): reuse the canonical derived attachment instead of minting a duplicate per save

headri:
Look up the canonical row for this filename across the whole dataset (not just
self's own children): re-editing an already-edited file, or editing a curve whose
dataset still holds both a .peak. and .edit. lineage, must reuse that single row -
scoping to children_of(self) would miss it (self isn't its own child) and mint a
duplicate .edit.jdx attachment on every save.

Linted module AttachmentJcampAasm completely and moved all "rubocop:disable" to the whole file because of a rubocop version difference between github and ELN

### DIFF
--- a/app/models/concerns/attachment_jcamp_aasm.rb
+++ b/app/models/concerns/attachment_jcamp_aasm.rb
@@ -211,6 +211,12 @@ module AttachmentJcampProcess
                     .where(filename: meta_filename)
                     .order(id: :desc)
                     .detect { |candidate| (candidate.root_id || candidate.id) == lineage_root }
+    # Re-parent a reused row onto self: the lookup above is identity by (lineage, filename),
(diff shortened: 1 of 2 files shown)


PiTrem (approved):
All three blocking findings from the last review are resolved at 915522ae: the reparent is guarded with ancestor_of? (verified present in the pinned ancestry 4.3.3), the four related-file cleanups are scoped to the source's ancestry root, and failure is now a valid from: state on the terminal events so a stuck row recovers instead of staying invisible to the viewer. CI green on all 7 checks. Two follow-ups, non-blocking: delete_related_edited_jcamp and delete_related_jsons still match container-wide without the lineage guard the other four gained, and the two comments about AASM non-bang persistence contradict each other (the persist: false one is correct).

PR #3519, opened 2026-08-25
perf(permissions): make permissions/status top-secret check set-based

TasnimMehzabin:
## What

Rewrites the `POST /api/v1/permissions/status` handler to stop loading and
iterating full element rows in Ruby. Two changes:

1. **Top-secret detection is now set-based.** The old code materialised every
   selected row and walked associations in Ruby:

   ```ruby
   is_top_secret ||= selected_elements[Sample].any?(&:is_top_secret?)
   is_top_secret ||= selected_elements[Reaction].lazy.flat_map(&:samples).any?(&:is_top_secret?)
   # ...wellplates, and screens two levels deep
   ```

   `any?` with a block cannot be pushed into SQL, so it loaded the whole
   relation; `flat_map(&:samples)` fired one `SELECT samples.*` per reaction and
   per wellplate (per wellplate *per screen* for the screen branch) — an N+1
   that grew with the size of the selection. `is_top_secret` is a plain boolean
   column, so each element type now resolves to a single `EXISTS` query
   (`SELECT 1 ... LIMIT 1`), joining through the container types where needed.
   The `||` chain still short-circuits on the first hit.

2. **`ElementsPolicy` is built once and reused.** The handler previously
   constructed a fresh `ElementsPolicy` per element type in each of two passes
   (destroy/remove/update, then share). Since `ElementsPolicy` memoises its
   own/shared record-id lookups per instance, the second pass re-ran those
   queries. The instances are now built once and reused across both passes.

## Why

Observed as an N+1 in production traces: whole sample rows loaded to test one
boolean, plus an unbatched association walk per selected container. The
sample-branch regression (single-column `pluck` replaced by block-form `any?`)
was introduced by #2783; the association walks are original 2015 behaviour.

## Behaviour

No functional change. Association coverage is identical (verified against the
generated SQL, including the soft-delete `deleted_at IS NULL` scopes carried
into every JOIN), and the policy result is unchanged (memoised lookups are a
deterministic function of the unchanged user/scope pair).

## Tests

- Correctness: top-secret detected through a selected reaction and through a
  selected screen (the deepest `screen → wellplate → well → sample` path).
- Regression guard: the top-secret probe stays a single query whether 1 or 5
  reactions/screens are selected — reverting to the association walk fails it.

All specs pass; RuboCop clean.

### DIFF
--- a/app/api/chemotion/permission_api.rb
+++ b/app/api/chemotion/permission_api.rb
@@ -28,15 +28,20 @@ module Chemotion
             end
           end
 
-          # checking if the selected elements include any one element that is top secret
-          is_top_secret = false
-          is_top_secret ||= selected_elements[Sample].any?(&:is_top_secret?)
-          is_top_secret ||= selected_elements[Reaction].lazy.flat_map(&:samples).any?(&:is_top_secret?)
-          is_top_secret ||= selected_elements[Wellplate].lazy.flat_map(&:samples).any?(&:is_top_secret?)
-          is_top_secret ||= selected_elements[Screen].lazy
-                                                     .flat_map(&:wellplates)
-                                                     .flat_map(&:samples)
-                                                     .any?(&:is_top_secret?)
+          # Checking if the selection includes any top-secret sample. is_top_secret is a plain
+          # boolean column, so each element type is checked with a single EXISTS query
+          # (SELECT 1 ... LIMIT 1) instead of materialising every selected row and walking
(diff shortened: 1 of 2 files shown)


PiTrem (approved):
Reviewed (see internal pr-3519-review.md): behavior-preservation independently verified (Rails console + full spec run), no correctness issues. Two minor nits noted (repetitive EXISTS clauses, duplicate rationale comments) — non-blocking.

PR #3523, opened 2026-08-26
fix(pubchem): stop LCSS misses re-queueing forever, and guard Fault responses

adambasha0:
# PubChem Fault handling and the LCSS backfill

## The problem

`Molecule#pubchem_lcss` stored the client's `nil` verbatim. A JSON null in
`taggable_data['pubchem_lcss']` matches `->>'pubchem_lcss' is null` — the predicate
`PubchemLookupJob#pending_scope` selects on — exactly as a missing key does, so those molecules
never left the pending scope and were re-asked on every sweep, indefinitely. Measured on a
production-data copy: **12,658 molecules, 23% of all molecule tags**, permanently stuck.

Around that, the client had no guard against PubChem's `Fault` bodies:

- `get_lcss_from_cid` checked only `success?`. Every Fault observed in practice arrives with a
  **404 or 400**, so `success?` does catch them today — but the endpoint's error contract is not
  guaranteed, and nothing between the status check and the store would stop a non-record body
  from being written as a molecule's GHS classification.
- Four other raw-body methods (`get_record_from_molfile`, `get_cids_from_inchikeys`,
  `get_molfile_by_inchikey`, `get_xref_by_inchikey`) checked **neither** `success?` nor `Fault`,
  so an error page could be parsed as a record.
- `interpret_record` assumed the shape of a `PC_Compounds` record and raised on a malformed one,
  failing the whole enrichment job rather than that molecule.
- `URI.encode` was still in use; it was removed in Ruby 3.

## The fix

- One `safe_response` gate: a response is usable only if `success?` **and** not a `Fault`. Every
  raw-body method returns `nil` instead of a Fault body. `get_lcss_from_cid` checks for a Fault
  after parsing, so a Fault can never be stored regardless of the status it arrives with.
- `interpret_record` reads every nested value through a `dig_hash` helper that walks only while
  each level is a Hash, so a malformed record — a non-Hash entry, a wrongly typed `id`, missing
  `props`, a props entry with no `urn` — degrades to the same nil-filled result a Fault produces
  instead of raising `NoMethodError`/`TypeError` up through `PubchemLookupJob`. The per-property
  branch moves to `apply_prop` and the duplicated literal to `empty_result`.
- `fault_response?` logs the expected not-found codes at `info` and everything else at `warn`:
  "not found" is PubChem's routine answer, and with five call sites now routed through this
  method a sweep would otherwise log one warning per molecule.
- `URI.encode` becomes `URI::DEFAULT_PARSER.escape` (verified character-for-character
  equivalent). `get_molfiles_by_inchikeys` is deleted: it had a broken URL and no callers.
- `ver_20170215133510_molecule.rake` skips a slice when `get_cids_from_inchikeys` returns nil,
  rather than passing nil to `JSON.parse`.

## The backfill (`20260825120000`)

The code fix stops producing bad values but cannot repair the stored ones, so the migration does:

- **A JSON null** becomes `false`, the "asked, PubChem has none" sentinel, with no network call.
  Note that a null also records a timeout or a 5xx, not only a genuine miss; collapsing both is
  deliberate, because `lcss || false` in `Molecule#pubchem_lcss` already does the same at
  runtime.
- **A stored Fault body** is truthy, so such a molecule would never be in the pending scope and
  would keep serving a Fault as safety data. The key is deleted so it gets asked once more with
  the corrected client, rate-limited by the existing `pubchem_checked_at` TTL. No such rows have
  been observed in production data — this branch is a safety net for an instance that has them,
  and a no-op otherwise.

Both statements are idempotent and the migration is not reversible: the previous values were not
valid data.

### Validated against a production-data copy

Run inside a rolled-back transaction on a database carrying production data:

| check | result |
|---|---|
| JSON-null rows rewritten | 12,658 |
| Fault-body rows cleared | 0 |
| real `Record` classifications altered | 0 |
| rows where any other `taggable_data` key changed | 0 |
| rows with no `pubchem_lcss` key touched | 0 |
| second run of both statements | 0, 0 (idempotent) |
| pending scope | 44,404 → 31,746 rows |

~0.3 s per statement over 54k rows.

## Tests

`spec/lib/chemotion/pubchem_service_spec.rb` (new) covers the nil-on-Fault contract and the
malformed-record shapes; `spec/lib/pub_chem_spec.rb` the Fault detection, the `safe_response`
gate on the raw-body methods, and both log levels; `spec/db/migrate/20260825120000_..._spec.rb`
the backfill, including what it must not touch — asserted against the real Postgres `->>`
predicate rather than a Ruby-side read of the column. The LCSS model spec previously asserted
against a molecule with no cid at all, so it passed regardless of what PubChem answered; it now
stubs the call and asserts the real `false`-not-nil storage convention. A broken heredoc in
`spec/spec_helper.rb` (a quoted string containing the literal `<<~BODY`) is repaired.

167 examples, 0 failures across the affected spec files; rubocop clean on every touched file.

### DIFF
--- a/db/migrate/20260825120000_backfill_pubchem_lcss_sentinels.rb
+++ b/db/migrate/20260825120000_backfill_pubchem_lcss_sentinels.rb
@@ -10,8 +10,17 @@ class BackfillPubchemLcssSentinels < ActiveRecord::Migration[6.1]
   #    selects on -- matches a JSON null exactly as it matches a missing key. So the molecule never
   #    left the pending scope and was re-asked on every sweep, indefinitely. The convention the code
   #    uses now is `false`, meaning "asked, PubChem has none": it reads back through `->>` as the
-  #    string 'false' and clears the scope. Rewriting these needs no network call, because the
-  #    answer is already known.
+  #    string 'false' and clears the scope. Rewriting these needs no network call.
+  #
+  #    A JSON null does not only mean "PubChem has none", though: get_lcss_from_cid also returns
+  #    nil on a timeout, a 5xx or any rescued exception, and Molecule#pubchem_lcss stored that
+  #    verbatim too. Those molecules are marked answered here without ever having been answered.
+  #    That is deliberate, because it is what the fixed code already does at runtime -- `lcss ||
+  #    false` collapses a transient failure to the same sentinel -- so the migration matches
+  #    application behaviour rather than introducing a new loss. Distinguishing the two would
+  #    need the :not_found/:unavailable outcome split that fetch_record_from_inchikey already
+  #    carries for the inchikey path, extended to LCSS; until then a molecule whose lookup
+  #    merely failed is recoverable by clearing its key, exactly as branch 2 below does.
   #
   # 2. A Fault body stored as GHS data. PubChem answers an unknown cid on the pug_view endpoint with
   #    HTTP 200 and a `{"Fault": ...}` body, and get_lcss_from_cid stored that body as if it were a
--- a/db/migrate/20260825120000_backfill_pubchem_lcss_sentinels.rb
+++ b/db/migrate/20260825120000_backfill_pubchem_lcss_sentinels.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+class BackfillPubchemLcssSentinels < ActiveRecord::Migration[6.1]
+  # Repairs the two bad shapes of `element_tags.taggable_data->'pubchem_lcss'` that the PubChem
+  # client fix shipping alongside this migration stops producing. Neither shape self-heals, so the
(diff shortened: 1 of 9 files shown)


PiTrem (approved):
Approving at cfd405395b. The sentinel fix is the substance and it is well evidenced: validated against a production-data copy, 12,658 molecules (23% of molecule tags) were stuck in the enrichment pending scope, and the migration retires them with every invariant holding — real classifications untouched, no other taggable_data key altered, never-asked rows untouched, idempotent on a second run. Review findings S1-S4 are fixed on the branch. The Fault guard stays as defence-in-depth; the PR description no longer claims a 200+Fault incident that the data does not support. All checks green.

PR #3529, opened 2026-08-27
fix(fetchers): restore pagination defaults in preparedCollectionParams

TasnimMehzabin:
## What

Re-inject the `page` / `per_page` defaults inside `preparedCollectionParams`, so
collection-list fetchers that are called without explicit pagination fall back to
the configured page size (`number_of_results`, default 15) instead of the
endpoints' small defaults (samples 7, vessels 5).

## Why

PR #3294 (rework fetchers) deleted `BaseFetcher.fetchByCollectionId`, which had
always injected `page = page || 1` and `per_page = per_page || number_of_results`.
The replacement `preparedCollectionParams` injected neither. Callers that pass
`{}` therefore silently fell through to the endpoints' small defaults and their
lists shrank. Normal collection click/refresh still passed `per_page`, so the
regression was limited to these flows:

- **Split as subsamples** (`ElementStore.handleSplitAsSubsamples`) - passes `{}`,
  sample list capped at 7.
- **Split as subwellplates** (`ElementStore.handleSplitAsSubwellplates`) - passes
  `{}` / nothing, so both the wellplates list (capped at 5) and the samples list
  (capped at 7) shrank.
- **Vessel-template mutations** (`VesselTemplateDetails.syncTemplateAndInstances`,
  called from create / bulk-create / update / delete instance and update template)
  - vessel instance list capped at 5.

## How

Centralized fix inside `preparedCollectionParams`, mirroring the deleted
`BaseFetcher` behavior, rather than patching each call site:

- `page` defaults to `1`, `per_page` defaults to `UIStore.number_of_results`.
- Both are only filled in when the caller omits them (`??`), so explicit values
  and the normal collection flow are unchanged. Handles both `perPage` and
  `per_page` input.
- `UIStore` is loaded at call-time via `require('...').default` rather than a
  static import. A static import closed a module-init cycle
  (`UserStore → UserActions → UsersFetcher → FetcherHelper → UIStore`) and crashed
  the app with `Cannot access '__WEBPACK_DEFAULT_EXPORT__' before initialization`.
  Same call-time pattern already used in `VesselUtilities.js`.
- Adds a `preparedCollectionParams` regression suite (spec/javascripts/utils/FetcherHelper.spec.js)
  covering default injection and explicit-value precedence.

## Manual QA

For each flow, watch the `/api/v1/...` request in DevTools:
- **Before:** no `per_page` param, list capped at 7 (samples/wellplates) or 5 (vessels).
- **After:** `per_page=15`, full set shown.

1. Split as subsamples -> `/api/v1/samples` carries `per_page=15`.
2. Split as subwellplates -> both `/api/v1/wellplates` and `/api/v1/samples` carry `per_page=15`.
3. Vessel-template create/update/delete instance + update template -> `/api/v1/vessels` carries `per_page=15`.
4. Control: normal collection click still respects an explicit page-size setting (e.g. set to 25 -> `per_page=25`).

Refs: N3 from the PR ELN#3294 rework-fetchers regression tracker.

### DIFF
--- a/app/javascript/src/utilities/FetcherHelper.js
+++ b/app/javascript/src/utilities/FetcherHelper.js
@@ -67,8 +67,19 @@ const classifyString = (str) => upperFirst(camelCase(str));
 const filteredSearchParams = (params) => new URLSearchParams(omitBy(params, isNil));
 
 const preparedCollectionParams = (id, params) => {
+  // Re-inject the pagination defaults the removed BaseFetcher.fetchByCollectionId used to
+  // supply, so callers passing {} (e.g. split-as-subsamples/subwellplates, vessel-template
+  // mutations) still get the configured page size instead of the endpoints' small defaults.
+  // Loaded at call-time (not a static import) to avoid a module-init import
+  // cycle: FetcherHelper is pulled in by every fetcher, and UIStore transitively
+  // imports those fetchers. Same pattern as VesselUtilities.js.
+  const UIStore = require('src/stores/alt/stores/UIStore').default;
+  const page = params?.page ?? 1;
+  const perPage = params?.perPage ?? params?.per_page ?? UIStore.getState().number_of_results;
   const collectionParams = {
     ...params,
+    page,
+    per_page: perPage,
     collection_id: id,
     from_date: (params?.fromDate ? dateToUnixTimestamp(params.fromDate) : null),
     to_date: (params?.toDate ? dateToUnixTimestamp(params.toDate) : null)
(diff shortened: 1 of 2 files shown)


PiTrem (approved):
Correct fix at the right layer: preparedCollectionParams re-injects the page/per_page defaults that BaseFetcher.fetchByCollectionId supplied before #3294 removed it, and explicit values still win so the normal collection flow is untouched. I checked all ten endpoints behind this shared helper — every one is Kaminari-paginated with a small default (5 or 7), none unbounded, so the change strictly raises the cap and cannot truncate a list that previously returned everything. Full JS suite green at this head (1691 passing, 0 failing). No blocking findings. Two non-blocking follow-ups worth a later pass: the spec's require.cache stub for UIStore is never restored and is safe only because 'utils/' sorts last in mocha's glob, and the comment above the lazy require names an import cycle through UsersFetcher that does not exist (that file does not import FetcherHelper) — the workaround is right, the documented path is not.

PR #3536, opened 2026-08-28
fix(ui): rework image preview modal layout and fix PDF sizing

headri:
## Summary
- Move the analysis preview modal's thumbnail carousel to a vertical rail beside the main preview instead of a horizontal strip below it, so the preview keeps the full body height and the rail scales to analyses with many attachments.
- Fold the filename caption into the modal title bar and move "Set as preferred" above the rail, reclaiming vertical space in the body.
- Skip re-fetching the large image when the already-selected thumbnail is clicked again (unnecessary reload).
- Fix PDF previews rendering at a tiny, squashed height: the PDF `<iframe>` has no intrinsic-size fallback (unlike ``, which falls back to its aspect ratio) when a percentage height can't resolve against the modal's content-sized, non-stretched ancestors, so it now gets an explicit height.

## Test plan
- [ ] Open an analysis with multiple image/PDF attachments and confirm the vertical thumbnail rail renders beside a full-height preview.
- [ ] Click the already-selected thumbnail and confirm no reload/spinner flash occurs.
- [ ] Click a different thumbnail and confirm the preview updates and "Set as preferred" enables/disables correctly.
- [ ] Open a PDF attachment and confirm it renders at a usable height, not collapsed.
- [ ] Confirm no unexpected scrollbar appears on the modal (header/footer stay pinned) at typical window sizes.

image

### DIFF
--- a/app/javascript/src/components/common/ImageModal.js
+++ b/app/javascript/src/components/common/ImageModal.js
@@ -197,6 +197,7 @@ export default class ImageModal extends Component {
 
   // click a thumbnail = preview large only; never persists
   handleSelectThumbnail = (thumb) => {
+    if (thumb.id === this.state.selectedId) return; // already showing this image
     this.setState({ selectedId: thumb.id });
     this.fetchLargeImage(thumb.id);
   };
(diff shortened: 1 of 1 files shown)


PiTrem (approved):
Self-contained layout rework of ImageModal. Verified AppModal forwards both new props (bodyClassName lands on Modal.Body; scrollable reaches react-bootstrap via {...rest}), and that the 'Set as preferred' gate is preserved structurally now that the button lives inside renderCarousel, which early-returns when there are no thumbnails. The iframe's explicit height is justified in-comment:  falls back to its intrinsic aspect ratio when a percentage height can't resolve, an iframe doesn't.

PR #3545, opened 2026-08-31
fix(ui): clarify reaction volume Conc checkbox label

### LINKED ISSUE #3541: “Calculate Conc” is confusing because it selects the calculation method
- I am an ELNer / ELN admin

# Description
The label “Calculate Conc” is confusing because it sounds like the option determines whether Conc is calculated or not.
However, switching this option only changes how Conc is calculated; it does not enable or disable concentration calculation.
I understand that the tooltip already explains this switching behavior. The issue is only that the label itself does not appropriately describe the function.
For example, “Use for Conc” may describe the function more clearly.

# Image
Image

TasnimMehzabin:
## What

Renames the reaction volume checkbox label from "Calculate Conc" to "Use for Conc".

## Why

"Calculate Conc" reads as if the checkbox turns concentration calculation on or off. It doesn't: concentration is always calculated. The checkbox only switches *which volume* the calculation is based on:

- **checked:** the reaction volume entered above
- **unchecked:** the summed volume of all reaction materials (solvents, starting materials, reactants)

"Use for Conc" describes that the entered volume is used for the Conc calculation. The existing tooltip already explains the full switching behavior and is unchanged.

## Changes

- Label text in `ReactionDetailsScheme.js` (`use_reaction_volume` checkbox). No behavior, tooltip, or logic changes.


JanCBrammer (approved):
Looks good. I'd suggest to write "Concentration" instead of "Conc" for clarity.

PiTrem (approved):
Label-only change on the use_reaction_volume checkbox; no behaviour, tooltip or logic touched. "Calculate Conc" wrongly implied the checkbox toggles concentration calculation on and off, when it only selects which volume feeds it. Re-approving the current head (13978dae), which changes the label to "Use for concentration" — the earlier approval was against 610da616 ("Use for Conc") and did not cover this string.

PR #3549, opened 2026-09-02
fix(test): resolve order-dependent flake in sample molecule_sort spec

### LINKED ISSUE #613: UI: harmonize fonts
# Description
I found that we do currently not have a consistent font scheme throughout the UI. Sometimes we default to sans-serif, otherwise we prefer Arial over Helvetica etc. In order to harmonize the look and feel for Chemotion, I suggest to have system default fonts followed by a single, uniform (sans) font for all the UI on all platforms, followed by some system-specific, "catch-most" fonts to support CJK and Emoji Characters.

Considering different licenses and looks, the Roboto font families might be a good fit (Apache 2.0, they can be used freely in products & projects - print or digital, commercial or otherwise). In addition, Roboto has a plain & simple appearance and is widely used having people perceive it as familiar. It comes in many styles and several variants, such as a thin to bold black and narrow or monospace. Considering technical aspects, Roboto is available in several formats, including `.woff` and `.woff2` and could thus easily be included via the Google Fonts CDN (to make increase the chance of cache hits) or shipped locally (for installations not having access to the web). The fonts do have a installation package for the Ubuntu platform and could easily be used on the backend for rendering static assets and graphics. In addition a wide range of latin and non-lating glyphs are supported which might help with other languages such as cyrillic or greek.

For the reasons above I suggest to use this in our `application.scss.erb`:

```scss
body, html {
  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
}
```
preferably in conjunction with loading a default font in the main application similar to this:
```html
<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto|Roboto+Condensed|WhatEverWeNeedCouldGoHere">
```

To increase consistency even further, I would suggest to carry this over to Chemotion Repo as well.

headri:
## Summary
- The `molecule_sort` spec's fixture setup set attributes in a `create_list` post-save block, which never persisted to the DB (FactoryBot's block runs after the INSERT without calling `save!`). Both molecules ended up sharing the default `sum_formular: 'H2O'`, making the two compared rows exact ties whose relative order Postgres doesn't guarantee — this is what flaked. Fixed the fixture to actually persist distinct values via `force_attributes`/factory overrides.
- Added deterministic secondary sort keys (`id`) to both `molecule_sort` branches in `SampleAPI` so the endpoint itself no longer depends on scan-order luck when ties occur, independent of the spec fix.
- Logged the exception in `database_cleaner.rb`'s swallowed-retry rescue branch so a future cross-example state bleed (the other suspected cause in the issue) can be pinpointed with evidence instead of silently retried.

## Test plan
- [x] `bundle exec rspec spec/api/chemotion/sample_api_spec.rb -e "with molecule_sort"` passes across 5 different `--seed` values
- [x] `bundle exec rspec spec/api/chemotion/sample_api_spec.rb` — full file, 69 examples, 0 failures
- [x] `bundle exec rubocop` on changed files shows only pre-existing, unrelated offenses

### DIFF
--- a/app/api/chemotion/sample_api.rb
+++ b/app/api/chemotion/sample_api.rb
@@ -214,10 +214,11 @@ module Chemotion
                            .where(id: (sample_scope.pluck :molecule_id))
                            .order(Arel.sql("LENGTH(SUBSTRING(sum_formular, 'C\\d+'))"))
                            .order(:sum_formular)
+                           .order(:id)
           reset_pagination_page(molecule_scope)
           paginate(molecule_scope).each do |molecule|
             samples_group = sample_scope.select { |v| v.molecule_id == molecule.id }
-            samples_group = samples_group.sort { |x, y| y.updated_at <=> x.updated_at }
+            samples_group = samples_group.sort { |x, y| [y.updated_at, y.id] <=> [x.updated_at, x.id] }
             samples_group.each do |sample|
               sample_list.push(
                 Entities::SampleEntity.represent(sample, detail_levels: detail_levels, displayed_in_list: true),
@@ -227,7 +228,7 @@ module Chemotion
           sample_count = sample_scope.size
         else
           sample_count = reset_pagination_page(sample_scope)
-          sample_scope = sample_scope.order('samples.updated_at DESC')
+          sample_scope = sample_scope.order('samples.updated_at DESC, samples.id DESC')
           paginate(sample_scope).each do |sample|
             sample_list.push(
               Entities::SampleEntity.represent(sample, detail_levels: detail_levels, displayed_in_list: true),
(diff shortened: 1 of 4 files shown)


PiTrem:
Dropped the last commit (search API tie-breakers) from this branch to keep the PR scoped to the flaky spec; the search ordering issues it touched are being tracked separately. Retitled as `test(fix):`. If you have local work on this branch, please reset it to the new head.

PiTrem (approved):
Fixture now actually persists distinct formulas and the SampleAPI ordering is deterministic on ties; scoped to the flaky spec.

PR #1930, opened 2024-05-21
Update react-bootstrap library

JoachimTillessen:
Draft: update to latest version of bootstrap-react
the rest of the code is untouched

PR #3551, opened 2026-09-03
test: fill in missing API specs, fix five authorization/exposure bugs found along the way

headri:
## Summary

Works through the old `pending 'TODO: Add missing spec'` / `'not yet implemented'` placeholders across six API spec files. For each, checked whether the endpoint under test is still live before writing coverage; several placeholders tested endpoints that had actually been removed or renamed years ago, so those were dropped in favor of a short note instead of faked coverage.

Writing real request specs against the actual behavior surfaced five pre-existing authorization/exposure bugs, fixed here alongside their regression specs, plus two small unrelated correctness bugs found and fixed along the way:

### Authorization / exposure fixes

- **`user_label_api.rb`** — `PUT /api/v1/user_labels/save_label` looked up the target label with `UserLabel.find(params[:id])`, with no check that it belonged to `current_user`. Any authenticated user could rename/recolor another user's label by id, and the update would even reassign it to themselves (`attr` always set `user_id: current_user.id`). Now scoped to `UserLabel.find_by(id:, user_id: current_user.id)`, returning 404 on mismatch.
- **`user_api.rb`** — `GET /api/v1/users/omniauth_providers` serialized the raw `current_user` AR object instead of going through `Entities::UserEntity` like every other user-facing endpoint. This leaked `encrypted_otp_secret`, the *decrypted* `otp_secret` (devise-two-factor's `attr_encrypted` virtual attribute), and `otp_backup_codes` to any authenticated user. The frontend (`OmniauthCredential.js`) only ever reads `current_user.providers`, so the response is now scoped to that.
- **`attachable_api.rb`** — `POST /api/v1/attachable/update_attachments_attachable` only authorized `attachable_type == 'ResearchPlan'`. Every other type the frontend actually sends (`Wellplate`, `DeviceDescription`, `SequenceBasedMacromoleculeSample`, `SequenceBasedMacromolecule`) had **no authorization check at all** — any authenticated user could attach a file to, or detach an existing attachment from, any other user's element by id. Replaced the single-branch check with a lookup table covering every real `attachable_type`, resolved uniformly through `ElementPolicy#update?`; an unrecognized type now fails closed.
- **`attachment_api.rb`** — three POST endpoints ran with no authorization at all, since the shared `before` block only covers GET/DELETE:
  - `POST /attachments/infer` — any authenticated user could run spectrum inference against any attachment by id, and the response leaks the base64-encoded file content (a read bypass, not just write).
  - `POST /attachments/save_spectrum` — same gap, letting any user regenerate/overwrite spectrum data on an attachment they don't own.
  - `POST /attachments/:attachment_id/annotation` — any authenticated user could overwrite the annotation SVG of any attachment by id.
  All three now use `writable?(@attachment)` (`AttachmentPolicy`), the same check already used by `regenerate_spectrum`/`regenerate_edited_spectrum` in this file.

None of these were previously reported in the `ComPlat/chemotion_ELN` or `ComPlat/chemotion` issue trackers (searched both before fixing).

### Small unrelated bugs fixed along the way

- **`attachment.rb`** — `Attachment#reload` returned `set_key`'s value (`nil`) instead of `self`, breaking the standard `record.reload.attribute` chain for any `Attachment` instance. Found while writing a detach regression spec.
- **`download_analyses_job.rb`** — `DownloadAnalysesJob#perform` checked `if rt == false` instead of `if @rt == false`, so it always raised `NameError` (silently swallowed by a surrounding `rescue`) and never wrote the zip to disk for the large-file async export path. Found while writing the `sample_analyses` spec.

### Per-file breakdown

- `spec/api/chemotion/user_label_api_spec.rb` — filled `save_label` (create/update/default access_level + the IDOR regression case).
- `spec/api/chemotion/wellplate_api_spec.rb` — filled `subwellplates` (split behavior, well/sample duplication, read-only-collection 401). Dropped the `well_label`/`well_color_code` stubs — both routes were removed in #1930; label/color_code are now set via `PUT /wellplates/:id`.
- `spec/api/chemotion/search_api_spec.rb` — filled `reactions`/`wellplates`/`screens` (substring search + cross-referenced elements). Dropped the `elements` stub — the route was renamed to `/search/all` back in early 2024 and is already covered under that describe block.
- `spec/api/chemotion/user_api_spec.rb` — filled `list_editors`, `omniauth_providers` (incl. the leak regression test), `update_counter`, `scifinder`, `sign_out`, `devices/novnc`, `devices/current_connection`.
- `spec/api/chemotion/attachable_api_spec.rb` — filled `update_attachments_attachable` (create/update for ResearchPlan and Wellplate, both owned and cross-user, plus detach) — the only pending spec in this file, and the one that surfaced the broken-access-control gap above.
- `spec/api/chemotion/attachment_api_spec.rb` — filled `sample_analyses`, `thumbnail`, `thumbnails`, `files`, the no-image case of the annotated-download endpoint, and `infer`; added coverage for `save_spectrum`'s and the annotation endpoint's new authorization checks. Dropped the `svgs` stub — the endpoint has no frontend caller and is structurally unreachable (the shared before-hook never grants it `can_dwnld`, so it unconditionally 401s).

## Test plan

- [x] `bundle exec rspec spec/api/chemotion/user_label_api_spec.rb spec/api/chemotion/wellplate_api_spec.rb spec/api/chemotion/search_api_spec.rb spec/api/chemotion/user_api_spec.rb spec/api/chemotion/attachable_api_spec.rb spec/api/chemotion/attachment_api_spec.rb spec/models/attachment_spec.rb` — all green (one pre-existing, unrelated network-stub flake in `attachment_spec.rb`, confirmed present on `main` too)
- [x] `bundle exec rubocop` on every changed file — clean

### DIFF
--- a/app/api/chemotion/attachable_api.rb
+++ b/app/api/chemotion/attachable_api.rb
@@ -4,6 +4,17 @@
 
 module Chemotion
   class AttachableAPI < Grape::API
+    # Every attachable_type the frontend actually sends to update_attachments_attachable
+    # (AttachmentFetcher.js#updateAttachables). Anything else is rejected below rather than
+    # silently skipping authorization, as previously happened for every type but ResearchPlan.
+    ATTACHABLE_CLASSES = {
+      'ResearchPlan' => ResearchPlan,
+      'Wellplate' => Wellplate,
+      'DeviceDescription' => DeviceDescription,
+      'SequenceBasedMacromoleculeSample' => SequenceBasedMacromoleculeSample,
+      'SequenceBasedMacromolecule' => SequenceBasedMacromolecule,
+    }.freeze
+
     resource :attachable do
       params do
         optional :files, type: [File], desc: 'files', default: []
@@ -13,13 +24,9 @@ module Chemotion
         optional :del_files, type: [Integer], desc: 'del file id', default: []
       end
       after_validation do
-        case params[:attachable_type]
-        when 'ResearchPlan'
-          error!('401 Unauthorized', 401) unless ElementPolicy.new(
-            current_user,
-            ResearchPlan.find_by(id: params[:attachable_id]),
-          ).update?
-        end
+        klass = ATTACHABLE_CLASSES[params[:attachable_type]]
+        record = klass&.find_by(id: params[:attachable_id])
+        error!('401 Unauthorized', 401) unless ElementPolicy.new(current_user, record).update?
       end
 
       desc 'Update attachable records'
--- a/app/api/chemotion/attachment_api.rb
+++ b/app/api/chemotion/attachment_api.rb
@@ -250,6 +250,8 @@ module Chemotion
         params do
           require :updated_svg_string, type: String
         end
+        error!('401 Unauthorized', 401) unless writable?(@attachment)
+
         updater = Usecases::Attachments::Annotation::AnnotationUpdater.new
         updater.update_annotation(
           params['updated_svg_string'],
@@ -599,6 +601,8 @@ module Chemotion
         optional :lcms_integrals_str, type: String
       end
       post 'save_spectrum' do
+        error!('401 Unauthorized', 401) unless writable?(@attachment)
+
         lcms_data = params[:lcms_mz_page_data]
         if lcms_data.respond_to?(:read)
           params[:lcms_mz_page_data] = lcms_data.read
@@ -677,6 +681,8 @@ module Chemotion
         optional :layout, type: String
       end
       post 'infer' do
+        error!('401 Unauthorized', 401) unless writable?(@attachment)
+
         predict = @attachment.infer_spectrum(params)
         params[:predict] = predict.to_json
         jcamp_att = @attachment.generate_spectrum(
--- a/app/api/chemotion/user_api.rb
+++ b/app/api/chemotion/user_api.rb
@@ -69,7 +69,9 @@ module Chemotion
       namespace :omniauth_providers do
         desc 'get omniauth providers'
         get do
-          { providers: Devise.omniauth_configs.keys, current_user: current_user }
+          # Only the linked-provider ids are used by the frontend (OmniauthCredential.js); serializing
+          # the full current_user leaked encrypted_otp_secret/otp_secret/otp_backup_codes.
+          { providers: Devise.omniauth_configs.keys, current_user: { providers: current_user.providers } }
         end
       end
 
--- a/app/api/chemotion/user_label_api.rb
+++ b/app/api/chemotion/user_label_api.rb
@@ -32,7 +32,8 @@ module Chemotion
           }
           label = nil
           if params[:id].present?
-            label = UserLabel.find(params[:id])
+            label = UserLabel.find_by(id: params[:id], user_id: current_user.id)
+            error!('404 Not Found', 404) if label.nil?
             label.update!(attr)
           else
             label = UserLabel.create!(attr)
--- a/app/jobs/download_analyses_job.rb
+++ b/app/jobs/download_analyses_job.rb
@@ -82,7 +82,7 @@ class DownloadAnalysesJob < ApplicationJob
         end
       end
 
-      if rt == false
+      if @rt == false
         zip_file.rewind
         File.write(@file_path, zip_file.read)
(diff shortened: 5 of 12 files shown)


PiTrem:
@headri — I've split this branch so the spec work and the two clean authorization fixes aren't held up by the attachment/attachable rework. Sorry for touching your branch directly; I did it as a plain revert commit on top (`26b301a`) rather than a force-push, so none of your history or authorship is lost.

**What moved:** the `attachable_api.rb` authorization rework and the three `writable?` gates in `attachment_api.rb` are preserved verbatim on **`attachments-authorization`**, branched from this PR's previous head `4802b75`. Once this lands, merge `main` back into it and open the follow-up PR — the reverted hunks come back as the diff to rework.

**Why they need rework** — four things, the first being the significant one:

1. The `update_attachments_attachable` fix doesn't close the IDOR it targets. `after_validation` authorizes the element named by `attachable_id`, but the detach query filters on `id` + `attachable_type` only — so the object authorized and the object mutated are different. Pass your *own* `attachable_id` with someone else's ids in `del_files` and it detaches them. Since `del_files` is unbounded and attachment ids are sequential, one request clears every attachment of that type instance-wide, and it isn't recoverable through the UI (once `attachable_id` is nil, the GET guard denies the owner too). Worth noting the allowlist *widens* this — before, only `ResearchPlan` reached the detach path with any check at all. The fix is to scope the query to the authorized record, and the regression spec wants the own-id/foreign-files shape; the current one sends the victim's `attachable_id`, so it 401s at validation and never reaches the detach branch.

2. `SequenceBasedMacromolecule` through `ElementPolicy` hits `record_shared_with_minimum_detail_level?`, which builds `sequencebasedmacromolecule_detail_level` — that column doesn't exist on `collection_shares` (only the `...sample_` one does). Shared SBMM raises instead of deciding. This PR is the first code path passing that class to `ElementPolicy`, so it hasn't been hit before.

3. `ElementPolicy#update?` requires detail level 10 on shared collections, so a collection shared with `edit_elements` but e.g. `wellplate_detail_level = 5` now 401s on attach for four element types that previously had no check. That may be the posture you want, but it's a behavior change worth making explicit and spec'ing.

4. `writable?` is `AttachmentPolicy`, which resolves the element via `record.container` — nil for attachments linked directly via `attachable`. So for ResearchPlan/Wellplate/DeviceDescription/SBMM attachments it returns false for anyone but `created_for`, and collaborators lose annotation on shared elements. `AttachmentHelpers#write_access?` resolves via `Attachment#root_element`, handles every element type, and is already mixed into that API.

**What I kept here:** most of `attachable_api_spec.rb` actually passes against the current API, so only the three contexts asserting the deferred behavior came out — the ResearchPlan and owned-Wellplate cases stayed. Same in `attachment_api_spec.rb`: just the three "belongs to another user → 401" contexts, happy paths untouched.

Two notes on the rest, both verified rather than assumed:

- The `Attachment#reload` fix is load-bearing for what's left — `attachable_api_spec.rb:78` uses `attachment.reload.attachable_id`. I audited the return value repo-wide before keeping it: nothing depended on the old `set_key` value, `set_key` is never overridden anywhere, and the one production call site was side-effect-only. I also collapsed the `export_collections.rb` workaround, since your fix makes its comment false.
- While you're in there: `set_key` has been a no-op since it was introduced and is never overridden, so deleting the `reload`/`set_key` pair outright would give you `self` from `super` for free and make the bug unrepeatable. Didn't do it here — it's new scope on your PR.

A couple of smaller things I deliberately left for you rather than adding scope: `UserLabels.js` only disables Edit for `access_level === 2`, but `my_labels` returns level-1 labels from any owner — so with the `save_label` fix landing, those now 404 into a `console.log` with the modal still open. And `save_label` rebuilds `attr` from params unconditionally, so a partial payload wipes `description`/`access_level`; the new spec sends exactly such a payload and asserts only the count.

PiTrem (approved):
Approving the split branch: the two authorization/exposure fixes that remain (omniauth 2FA-secret leak, save_label IDOR) were each traced end-to-end, and the attachable/attachment authorization rework has been deferred to the attachments-authorization branch. CI green at 26b301a, coverage 73.2%.

PR #3553, opened 2026-09-03
fix(reactions): highlight Amount (not Mass) as the recalculated field when Eq is edited

### LINKED ISSUE #3538: Mass field highlighted instead of amount after editing Eq
- [x] I am an ELNer
- [ ] I am a developer
- [x] I am a system/ELN admin

# Expected Behavior

When the **Eq** value of a starting material or reactant is edited, the **Amount** field should be highlighted as the recalculated field, since equivalents are based on the molar amount.

# Current Behavior

When the **Eq** value is edited, the **Mass** field is highlighted instead, although the calculation is performed based on the molar amount. It looks as if the user defined the amount based on the mass.

# Steps to Reproduce

1. Create or open a reaction.
2. Add a reference starting material and another starting material or reactant.
3. Enter the quantities for the reference material.
4. Edit the **Eq** value of the other material.
5. Observe that the **Mass** field is highlighted rather than the **Amount** field.

# Context

- ELN version: 3.1.2
- OS: macOS
- Browser: Safari

TasnimMehzabin:
## Problem

When the **Eq** value of a starting material or reactant is edited in a reaction,
the **Mass** field is highlighted as the recalculated field. But the calculation is
molar-amount based (equivalents are derived from the molar amount), so the **Amount**
field should be highlighted instead. The current behavior makes it look as if the user
defined the amount by mass.

Reported on ELN 3.1.2.

## Root cause

The highlighted ("active") field in the reaction table is driven entirely by the
sample's `amount_unit`:

- Mass is highlighted when `amount_unit === 'g'` (`Material.js:1164`)
- Amount is highlighted when `amount_unit === 'mol'` (`Material.js:740`)

The Eq-change handler computed a new molar amount but stored it via
`setAmountAndNormalizeToGram(...)`, which forces `amount_unit = 'g'` (`Sample.js:1124-1127`).
So the amount was correctly computed from moles, but the unit flag was flipped to grams,
lighting up **Mass**.

## Fix

In the two equivalent-change paths in `ReactionDetailsScheme.js`
(`handleEquivalentBasedAmountUpdate` and `updatedSamplesForEquivalentChange`), regular
materials now use `setAmount({ value: newAmountMol, unit: 'mol' })` so `amount_unit`
stays `'mol'` and the **Amount** field is highlighted.

This mirrors the pattern the team already adopted for direct amount edits
(`ReactionDetailsScheme.js:746-750`), where `setAmountAndNormalizeToGram` was replaced
with `setAmount(amount)` specifically to preserve the edited unit for highlighting.

No displayed numbers change: `amount_g` and `amount_mol` are derived getters, and for a
material with a valid molecular weight the two storage forms round-trip to identical
values. Only which field is highlighted changes.

## Scope / safety

- **Gas** samples (amount derived from the reaction vessel) and **mixtures without a
  usable reference MW** (amount tracked as total mass) keep the original gram-normalized
  path, so their calculations are unchanged.
- SBMM samples were already correct (separate mol-driven path).
- `amount_unit === 'mol'` is already a supported, persisted state (it is what a direct
  mol-field edit produces), so serialization and the backend are unaffected.

## Testing

- Added regression tests: a regular material recalculates via `setAmount({ unit: 'mol' })`
  (Amount highlighted); a gas material still uses the gram-normalized path.
- All 36 tests in `ReactionDetailsScheme.spec.js` pass.

## Steps to verify

1. Open a reaction.
2. Add a reference starting material and another starting material/reactant.
3. Enter quantities for the reference material.
4. Edit the **Eq** value of the other material.
5. The **Amount** field is highlighted (previously **Mass** was).

Closes https://github.com/ComPlat/chemotion_ELN/issues/3538

### DIFF
--- a/app/javascript/src/apps/mydb/elements/details/reactions/schemeTab/ReactionDetailsScheme.js
+++ b/app/javascript/src/apps/mydb/elements/details/reactions/schemeTab/ReactionDetailsScheme.js
@@ -1853,9 +1853,19 @@ export default class ReactionDetailsScheme extends React.Component {
       // For mixture samples with a valid reference component, calculate mass from mol amount
       const newAmountG = newAmountMol * sample.reference_component.relative_molecular_weight;
       sample.setAmount({ value: newAmountG, unit: 'g' });
-    } else {
-      // For regular samples or mixtures without reference MW, fall back to standard method
+    } else if (sample.isGas() || sample.isMixture()) {
+      // Gas amounts are derived from the reaction vessel, and mixtures without a usable
+      // reference MW track their amount as total mass; keep the gram-normalized behavior
+      // for these so their existing calculations are unchanged.
       sample.setAmountAndNormalizeToGram({
+        value: newAmountMol,
+        unit: 'mol',
+      });
+    } else {
+      // For regular samples, set the mol amount directly. Equivalents are derived from the
+      // molar amount, so keep 'mol' as amount_unit here (rather than normalizing to grams)
+      // so the Amount field is highlighted as the recalculated field instead of Mass.
+      sample.setAmount({
         value: newAmountMol,
         unit: 'mol',
       });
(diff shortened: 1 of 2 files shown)


PiTrem (approved):

PiTrem (approved):
Re-approving at b1829fa662: no blocking findings; the Eq-edit highlight fix, the render-path unit preservation and the SBMM id/type matching are covered by specs, CI green.

PR #3561, opened 2026-09-11
Fix/spectra editor header

alizaib1217:

- [ ] added code is linted


PiTrem:
Suggested change for `app/assets/stylesheets/legacy/spectra.scss`, found while testing this PR with ComPlat/react-spectra-editor#337: the CV chart comes out far shorter than the space available.

**Why:** the column rule gives *every* direct child `div` of the chart/panel columns `flex: 1 1 auto`. The CV chart column holds two of them: the chart wrapper, and the voltammetry table stacked under it. So the table grows too and takes a share of the column's height. The chart sizes itself to its own wrapper, so it keeps drawing at the reduced height. This rule's specificity is higher than the package's JSS classes, so it has to be fixed on this side. RSE#337 also sets `flex: 0 0 auto` on the table, but that alone does not win.

Only the first wrapper needs to grow:

```diff
       // LayerContent/Panel render an unstyled wrapper <div> here before the
       // d3-mounted / InfoPanel content; without this it's a block box with
       // auto height, which breaks the percentage-height chain below it.
-      > div {
+      // Only that first wrapper grows: the CV chart column also stacks the
+      // voltammetry table under the chart, and growing it too hands it a share
+      // of the chart's height - which the chart, sizing itself to its own
+      // wrapper, then keeps. The table takes its natural height instead.
+      > div:first-child {
         flex: 1 1 auto;
         min-height: 0;
       }
```

Also note that RSE#337 moves every layout back to a **9/3** chart/panel split. The `xs-9`/`xs-3` selectors that were already here apply again. The `xs-8`/`xs-4` ones this PR adds are then unused, but harmless.

PR #3563, opened 2026-09-14
feat: inline image and file attachments in Quill rich-text fields

haditariq:

- [ ] added code is linted


haditariq (commented):
## Code Review — `feat/quill-inline-images-to-attachments`

**Verdict: Changes needed before merge** — one confirmed logic bug that silently breaks import remapping for all three new element types.

---

### 🔴 Must fix

**`lib/import/import_collections.rb:853` — identifier_map key mismatch (import remap always no-ops)**

The export ZIP stores attachments as `attachments/<identifier>.<ext>` (export_collections.rb:83):
\`\`\`ruby
attachment_path = dir_path.join("#{attachment.identifier}#{File.extname(attachment.filename)}")
\`\`\`

On import, `att.filename` becomes `file_name = entry_name.sub('attachments/', '')` → `"abc-uuid.png"` (with extension). The identifier map is then built as:
\`\`\`ruby
map[att.filename] = att.identifier   # key: "abc-uuid.png"
\`\`\`

But the Quill delta stores `attachment_identifier: "abc-uuid"` — a bare UUID with no extension. The lookup `identifier_map[old_id]` always misses, so import remapping for Reaction, Screen, and Wellplate richtext silently does nothing.

**Fix (one line):**
\`\`\`ruby
# before
map[att.filename] = att.identifier if att.filename.present? && att.identifier.present?
# after
map[File.basename(att.filename, '.*')] = att.identifier if att.filename.present? && att.identifier.present?
\`\`\`

---

### 🟡 Should fix

**`ReactionDetails.js` — badge only scans `observation`, not `description`**

Reaction has two richtext fields. The save-strip in `handleSubmit` correctly strips both, but `inlineAttachmentIdentifiers` passed to `<AttachmentTab>` only wraps `reaction.observation`. Files dropped into `reaction.description` won't show the inline badge in the Attachments tab.

\`\`\`js
// fix
inlineAttachmentIdentifiers={collectInlineAttachmentIdentifiers([
  { type: 'richtext', value: reaction.observation },
  { type: 'richtext', value: reaction.description },
])}
\`\`\`

**`INLINE_BLOT_KEYS` defined in 4 places**

The constant `%w[attachment-image attachment-file]` is duplicated across `reaction.rb`, `screen.rb`, `wellplate.rb`, and `import_collections.rb`. A new blot type requires editing all four. Extracting to a shared concern (e.g. `QuillInlineAttachmentRemappable`) would make this maintainable.

---

### 🟢 Nice to improve

**`copy.rb:26` — duck-type change has no explanatory comment**

`instance_of?(::ResearchPlan)` → `respond_to?(:update_body_attachments)` is the right call, but without a comment it reads as accidental broadening. A one-liner noting this is an intentional opt-in protocol would help future readers.

**`import_collections.rb:850` — timing dependency is undocumented**

`update_richtext_attachment_identifiers` runs inside `extract` (line 104), before `import_attachments` finalises the DB state (line 144). This ordering is load-bearing and worth a brief comment.

---

### ✅ Confirmed clean

- `@attachments +=` in the three new `fetch_*_richtext_attachments` methods is correct — `fetch_many` only serialises to `@data`, it does not touch `@attachments`. Pattern matches the existing ResearchPlan path.
- `WellplateProperties.onAttachmentsChange: () => {}` default is safe — no conditional guard around this prop in the component.
- Duck-type broadening in `copy.rb` is sound — all three new models implement `update_body_attachments`.
- Branch is well-scoped: all 15 commits are directly on-topic.

PR #3568, opened 2026-09-16
fix(reaction): update temperature display to include unit for consistency

### LINKED ISSUE #3451: Display °C in the structural equation
- I am an ELNer

°C is no longer displayed in the structural equation, either in the ELN or in the Repo. 
It only happens when you create a new reaction. 

Image

TasnimMehzabin:
## Problem
Temperature ranges (e.g. `21 ~ 25`) lost their unit (`°C`) in the reaction scheme. It showed most often on newly created reactions.

## Cause
The browser-side scheme builder only appended the unit when the temperature was a single number. A range failed the pattern test, so the unit was dropped, and the app persists the browser-composed SVG as-is.

## Fix
- Widen the numeric check to also match ranges.
- Move the logic into a single `Reaction#temperature_display_with_unit` getter (mirroring the backend method), replacing the two duplicated copies.

Free text (e.g. `reflux`) is still left untouched.

## Testing
- `25` → `25 °C`
- `21 ~ 25` → `21 ~ 25 °C`
- `reflux` → `reflux` (unchanged)

**Note**: previously saved reactions with a range still show the old unit-less SVG until edited or re-saved; only newly generated schemes are corrected.

Closes https://github.com/ComPlat/chemotion_ELN/issues/3451


PiTrem (approved):
Re-reviewed at 20e170b309: all previous findings addressed, CI green. Thanks!

PR #3569, opened 2026-09-16
fix(profile): add vessel to the element tab layout for existing users

### LINKED ISSUE #3467: Vessel element is missing from Element Bar, making created Vessel Templates inaccessible
# Expected Behavior

After creating a Vessel Template via **+ CREATE → Create Vessel Template**, the created template should remain accessible from the UI.

For example, users should be able to:

- open the Vessel element from the Element Bar,
- view existing Vessel Templates,
- reopen and edit a previously created Vessel Template,
- create and manage Vessel instances from the template.

If the Vessel element is hidden by default, it should at least be available in the Element Bar layout configuration so that users can enable it.

# Current Behavior

A Vessel Template can be created successfully via **+ CREATE → Create Vessel Template**.

However, after saving and closing the Vessel Template, there appears to be no way to access it again.

In particular:

- **Vessel is not shown in the Element Bar**
- **Vessel is also not available in the Element Bar layout configuration**
- therefore, there is no discoverable way to display existing Vessel Templates or Vessel instances after closing the creation dialog

# Failure Information 

## Steps to Reproduce

1. Select **+ CREATE → Create Vessel Template**.
2. Enter the required information and click **Create**.
3. Create Vessel Instances via **Vessel Instances → Bulk Create**.
4. Close the Vessel Template tab.
5. Check the Element Bar for a **Vessel** element.
6. Open **Element Tab Layout**.
7. Check whether **Vessel** is available in either the visible or hidden elements.
8. Observe that **Vessel** is not available in either location.
9. There is therefore no apparent way to reopen the Vessel Template or access the Vessel Instances created in step 3.

Image

## Context

 - OS: Windows 11 Pro 25H2
 - Chrome 151.0.7922.138
 - https://demo.chemotion.ibcs.kit.edu/
   - Chemotion ELN v3.1.2

TasnimMehzabin:
## Problem

Vessel is absent from the Element Tab Layout settings. It appears in neither
the Visible nor the Hidden list, so users cannot enable the Vessel element from
the layout editor.

## Root cause

The element tab layout is driven entirely by the per-user profile hash at
`profile.data['layout']`. The frontend only renders element types that exist as
keys in that hash. `vessel` was never added to existing profiles:

- `config/profile_default.yml` (the config the profile API reads to backfill
  missing keys) was missing `:vessel`, so `GET /profiles` never backfilled it.
- The hardcoded default in `app/models/profile.rb` includes `vessel`, but it
  only runs `before_create` and bails when a layout already exists, so it never
  touches existing profiles.
- Unlike `sequence_based_macromolecule_sample`, no data migration was ever
  written to backfill `vessel` into existing profiles.

## Changes

- Add `db/migrate/..._init_vessel_profile_definition.rb`, mirroring the sbmm
  migration, to backfill `vessel = -1300` into every existing profile that
  lacks it (reversible).
- Add `:vessel` (and `:sequence_based_macromolecule_sample`, for consistency
  with the tracked `.example`) to the `:layout:` block in
  `config/profile_default.yml` across all environments.

Vessel is seeded at `-1300`, so it lands in the Hidden list by default. Users
opt in by dragging it to Visible.

## Testing

- [ ] Run `RAILS_ENV=development bin/rails db:migrate`
- [ ] Reload profile, open Tab Layout, confirm Vessel appears under Hidden
- [ ] Drag Vessel to Visible and confirm it persists after reload

Closes https://github.com/ComPlat/chemotion_ELN/issues/3467


TasnimMehzabin:
Fixed with https://github.com/ComPlat/chemotion_ELN/pull/3572

PR #3572, opened 2026-09-16
fix(profile): keep built-in ELN elements in the tab layout when genericElement is enabled

TasnimMehzabin:
## Summary

When the `genericElement` matrix is enabled, all element tabs and every per-element option in the **＋ Create** menu (Create Sample, Create Reaction, etc.) disappear. Only the fixed items (Create Wellplate from Samples, Create Screen from Wellplates, Create Vessel Template) remain, so the user cannot create a sample/reaction/vessel at all.

## Root cause

`GET /api/v1/profiles` rebuilds the tab layout on every request. Inside the `if current_user.matrix_check_by_name('genericElement')` block it filters the layout against an allow-list:

```ruby
available_elements = Labimotion::ElementKlass.where(is_active: true).pluck(:name)
...
new_layout = new_layout.select { |e| available_elements.include?(e) }
```

`available_elements` only ever contained **active generic element names**, never the built-in ELN elements (`sample`, `reaction`, `vessel`, ...). When no generic `ElementKlass` is active, the allow-list is `[]`, so the filter strips every element and the browser receives an empty layout. The tab bar and the Create menu both derive from that layout, so both go blank.

The failure only surfaces once `genericElement` is turned on, because the whole block is skipped while the matrix is off.

## Fix

- Build the allow-list from `::API::ELEMENTS` (the canonical built-in element list) **plus** the active generic element names, so built-in elements are always preserved.
- Extract two helpers that were previously duplicated between the GET and PUT handlers:
  - `available_element_names` — the allow-list described above.
  - `reindex_layout` — the positive/negative sort-and-reindex logic.
- Source the PUT handler's default layout from `config/profile_default.yml` (via a `default_layout` helper) instead of a hardcoded literal, so the default values live in one place.
- Fix the `available_ements` typo.
- Write the result back to the string `'layout'` key (the previous symbol key `data[:layout]` collided with the existing `'layout'` string key on JSON serialization).

## Tests

Added a regression spec for `GET /api/v1/profiles` with `genericElement` enabled and no active `ElementKlass`, asserting the built-in elements remain in the returned layout. Full `profile_api_spec` passes (15 examples, 0 failures).

## Notes

- No changes to `profile_default.yml`; the layout data in the DB was never corrupted — the endpoint was dropping the elements in transit.
- `Profile#data_default_layout` still keeps its own hardcoded copy of the default layout. Left as-is here; a follow-up could point it at the same config for a single source of truth.


PiTrem (approved):
All blockers from the review are resolved: the allow-list now comes from profile_default.yml plus active generic elements, the config is the single source of truth for the layout, and no path can persist a blank one. All 7 checks green at 68c233af01.

PR #3573, opened 2026-09-16
fix(profile): stop mutating the shared default-layout config so concurrent first-time sign-ins don't crash

TasnimMehzabin:
## Problem

Users on a live instance hit intermittent 500s on sign-in:

    RuntimeError (can't add a new key into hash during iteration)
    app/api/chemotion/profile_api.rb:35:in `block (3 levels) in <class:ProfileAPI>'

It could not be reproduced by signing in as a single user, because it is a
concurrency bug on process-wide shared state.

## Root cause

`Rails.configuration.profile_default.layout` is one hash, loaded once at boot
and shared by every request. In `GET /api/v1/profiles`, for a profile whose
`data` has no `layout` key, the code:

1. leaked a **reference** to that shared hash into the per-request data, then
2. **mutated** it.

Under concurrent sign-ins, one thread adds keys to the shared hash while another
copies it (`.dup` iterates it), which raises `can't add a new key into hash
during iteration`.

Profiles with empty `data` are exactly the ones that trigger it: non-`Person`
users (Admin, Group) get a profile with no `layout` by design, since
`Profile#set_default` returns early for non-Persons. The reproduction ("sign in
as Admin, create users, assign to group, sign in as an assigned user") lines up:
the Admin's empty-data profile seeds the mutation, and the Persons signing in
concurrently race on it.

## Fix

Deep-dup the config when reading it in the endpoint, so each request works on its
own copy. Nothing shared is ever mutated, so the race is gone. As a bonus it also
stops the slow permanent pollution of the global config that every first-time GET
was causing.

    layout = Rails.configuration.profile_default&.layout&.deep_dup

The endpoint still fills the default layout into the per-request `data` for the
response, so behaviour is unchanged for every user type.

## Tests

Added `GET /api/v1/profiles` coverage in
`spec/api/chemotion/profile_api_spec.rb`:

- returns the profile successfully for a profile with no stored layout
- deep-freezes the shared config and asserts the request still succeeds; the old
  code raises `FrozenError` here, so a regression fails the test loudly

Verified the regression case fails without the fix (`FrozenError`) and passes
with it. Full file: 16 examples, 0 failures.

## Not changed (intentional)

Non-`Person` users (Admin, Group) still get no layout, which is by design. New
`Person` users, single or bulk (both go through `POST /api/v1/admin/users` ->
`User.create!`), still get their layout via `set_default` on first profile
access. Backfilling legacy Person profiles missing a layout is out of scope for
this crash fix.


PiTrem (approved):
Correct root cause and minimal fix: the endpoint no longer leaks a reference to the process-wide profile_default.layout hash into per-request data before mutating it. Spec coverage freezes the shared config so a regression fails loudly. CI green.

